PR #3, opened 2015-09-19
Fix Facebook login error

janicduplessis:
There was a javascript error that caused FacebookUtils.authenticate to crash and made login with Facebook impossible since v1.6.0.

Fixes https://developers.facebook.com/bugs/882121675197813/


andrewimm:
Thanks for catching this

PR #7, opened 2015-09-22
Avoid recursion on encoding circular pointers

andrewimm:
objects should only be encoded with flat pointers, not inflated objects. This adds a test that fails without this change.

### DIFF
--- a/src/__tests__/ParseObject-test.js
+++ b/src/__tests__/ParseObject-test.js
@@ -523,6 +523,47 @@ describe('ParseObject', () => {
     expect(o2.get('age')).toBe(22);
   });
 
+  it('does not stack-overflow when encoding recursive pointers', () => {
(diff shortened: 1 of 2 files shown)


src/__tests__/ParseObject-test.js
```diff
@@ -523,6 +523,47 @@ describe('ParseObject', () => {
...
+    expect(o.toJSON()).toEqual({
+      objectId: 'recurParent',
+      child: {
+        __type: 'Pointer',
+        className: 'Item',
+        objectId: 'recurChild'
+      }
+    });
+
+    var child = ParseObject.fromJSON({
```
hallucinogen: This assumes that single-object instance is true?

andrewimm: Single-instance is on by default for all tests, unless explicitly disabled.

PR #24, opened 2015-09-27
Fix outdated comments

danrobinson:
The function documentation on some `Parse.Object` methods refers to Backbone-specific behavior that was removed from the JS SDK in version 1.6.0.  This removes those references.


andrewimm:
Thanks, much appreciated!

PR #29, opened 2015-09-29
Perform user rewrite on Parse.Query

### LINKED ISSUE #10: matchesQuery API on a Relation gives error code 102
We have added a "joinCommunity" relation to the User object and use the following query.

new Parse.Query("User").matchesQuery("joinCommunity", new Parse.Query("Community").equalTo("objectId", "acMv7RpF1l")).find()

The above gives the following error in SDK v1.6.2:
{code: 102, message: "bad type for $inQuery"}

The SAME query works in v1.5.0.

andrewimm:
Fixes #10

### DIFF
--- a/src/__tests__/ParseQuery-test.js
+++ b/src/__tests__/ParseQuery-test.js
@@ -1289,4 +1289,13 @@ describe('ParseQuery', () => {
       done();
     });
   }));
+
+  it('rewrites User queries when the rewrite is enabled', () => {
+    CoreManager.set('PERFORM_USER_REWRITE', true);
+    var q = new ParseQuery('User');
+    expect(q.className).toBe('_User');
+    CoreManager.set('PERFORM_USER_REWRITE', false);
+    q = new ParseQuery('User');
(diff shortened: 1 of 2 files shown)


oriSomething:
it's compatible with old versions, but might be a good idea to use some deprecation API as in Ember or at least adding a `console.warn` for this issue

andrewimm:
@oriSomething This is simply correcting an oversight when we moved to the new codebase. We don't have plans to deprecate this behavior anytime soon -- it's just not thoroughly documented.

src/__tests__/ParseQuery-test.js
```diff
@@ -1289,4 +1289,13 @@ describe('ParseQuery', () => {
       done();
     });
   }));
+
+  it('rewrites User queries when the rewrite is enabled', () => {
```
wangmengyan95: Nit: we actually test two cases not only rewrite is enabled, so maybe change a name?
But really small nit, depends on you.

PR #30, opened 2015-09-30
Optimistically set ACLs from JSON blobs

### LINKED ISSUE #23: beforeSave makes all fields undefined
When I add a beforeSave function to my Parse hosted CloudCode, all fields in the class I'm saving get changed to undefined.  It doesn't matter what I put in the beforeSave code.  A simple function like:

    Parse.Cloud.beforeSave("chart", function(request, response) {
        console.log("Running beforesave")
         response.success()
    });

changes all fields to undefined.  If I switch to JSSDK 1.5.0, everything works fine.  So it appears to be a bug in 1.6.x and higher.

andrewimm:
Fixes #23

### DIFF
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -601,6 +601,12 @@ export default class ParseObject {
         newOps[k] = opFromJSON(changes[k]);
       } else if (k === 'objectId' || k === 'id') {
         this.id = changes[k];
+      } else if (
+        k === 'ACL' &&
+        typeof changes[k] === 'object' &&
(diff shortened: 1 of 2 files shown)


wangmengyan95:
Code LGTM, but why this cause other fields to become undefined?

andrewimm:
Validation is performed on every set(), and the default validation is to ensure that values in the ACL field are `instanceof ParseACL`. Cloud code performs an inner set() on beforeSave, so that the fields that were updated appear dirty within cloud code, but it was receiving a JSON-form ACL and failing that set.

PR #33, opened 2015-09-30
Fix Bug in add method

thalesfp:
Error description:

*/node_modules/parse/lib/node/ParseRelation.js:101
      return parent;
             ^

ReferenceError: parent is not defined


andrewimm:
Appreciate the PR, but this was actually already fixed: https://github.com/ParsePlatform/Parse-SDK-JS/blob/master/src/ParseRelation.js#L85

Issue #12, opened 2015-09-23
ParseUser._linkWith call to provider.authenticate need to reject promise on error


taivo:
At approx. at line 100 in ParseUser.js, these lines
```
this._linkWith(provider, opts).then(() => {
            promise.resolve(this);
          });
```
recursively call `_linkWith` and only handle the smooth sailing case. However, it's still possible for this error to happen `{ code=208,  message="Another user is already linked to this facebook id."}`. A `promise.reject` call is needed to propagate the error message.

andrewimm:
Should be fixed in #34, it'll go out in v1.6.5

PR #34, opened 2015-09-30
Expose error in _linkWith recursive call

wangmengyan95:
When `_linkWith` is called recursively, we can not get the error thrown by the not first `_linkWith` call, as mentioned in #12 . This PR fixes this and add a test.

### DIFF
--- a/src/__tests__/ParseUser-test.js
+++ b/src/__tests__/ParseUser-test.js
@@ -32,6 +32,7 @@ var ParseObject = require('../ParseObject');
 var ParsePromise = require('../ParsePromise');
 var ParseUser = require('../ParseUser');
 var Storage = require('../Storage');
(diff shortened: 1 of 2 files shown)


src/__tests__/ParseUser-test.js
```diff
@@ -360,4 +361,49 @@ describe('ParseUser', () => {
...
+      restoreAuthentication(authData) {},
+
+      getAuthType() {
+        return 'test';
+      },
+
+      deauthenticate() {}
+    };
+    
+    ParseUser.logInWith(provider, {}).then(() => {
```
andrewimm: you can write `.then(null, (error) => {})` instead if you don't intend to do anything with the first argument

andrewimm:
Style nit, but otherwise LGTM

PR #37, opened 2015-10-01
Parse.Query.fromJSON() method

kyptov:
I have a lot of records in my class and I wanted to fetch some of them on the server side and continue to fetching on client side. Like an infinite scroll. So I need to query records both from server and client sides. So I wanted to send query conditions from server side to client to avoid duplication. 

Parse.Query has method toJSON, I created method fromJSON.


andrejkosyak:
please pull this one

codebreach:
Can we pull this? I can add a test cases if thats what is blocking?

flovilmart:
@kyptov thanks for the PR (and sorry for the late response). Do you mind adding tests to cover that feature?

kyptov:
created new one https://github.com/parse-community/Parse-SDK-JS/pull/448

Issue #13, opened 2015-09-23
Cloud code - calling signUp on anonymous user doesn't make user non-anonymous


alexblack:
I'm using this cloud function to upgrade a user from anonymous to real, and then request a password reset, on app https://www.parse.com/apps/invoice-maker-dev

After doing this, I then login with that user on an AndroidDevice, then call ParseAnonymousUtils.isLinked and expected the result to be false (real) but I get true (anonymous).

Looking at the ParseUser in the parse console I see that the user is still anonymous.  If I follow a similar process from Android code, the user doesn't remain anonymous (the desired result).

See user bF2wHKAFDM in the screenshot, it does have the username set from the code below, but still is an anonymous user.

What do I need to do to make this user not anonymous?

Thanks!  Alex

```javascript
Parse.Cloud.define('requestPasswordResetInBackgroundAnonymous', function(request, response) {
  var googleAccountQuery = new Parse.Query(GoogleAccount);
  var email = request.params.email;

  if (!email || !email.length) 
    response.error('Email parameter required');

  Parse.Cloud.useMasterKey();

  googleAccountQuery.equalTo('name', email);
  googleAccountQuery.ascending('updatedAt'); // in case there are multiple
  googleAccountQuery.first({    
    success: function(googleAccount) {
      if (googleAccount) {
        var account = googleAccount.get("account");

        var userQuery = new Parse.Query(Parse.User);
        userQuery.equalTo('account', account);
        userQuery.ascending('updatedAt');
        userQuery.first({
          success: function(user) {
            if (user) {
              user.setUsername(email);
              user.setPassword('Eq54mY914Tt4:2U7WIf28K+o4483xi3Y');
              user.signUp(null, {
                success: function() {
                  Parse.User.requestPasswordReset(email, {
                    success: function() {
                      response.success();
                    },
                    error: function(error) {
                      response.error('requestPasswordReset failed: ' + error.code + ', ' + error.message);
                    }
                  });
                },
                error: function(user, error) {
                  response.error('signup failed: ' + error.code + ', ' + error.message);
                }
              });
            } else {
              response.error('No ParseUser found for account: ' + account.id);              
            }
          },
          error: function(error) {
            response.error('Error searching for ParseUser by account' + account.id + ': ' + error.code + ', ' + error.message);
          }
        });
      } else {
        response.error('No GoogleAccount found with email: ' + email);
      }
    },
    error: function(error) {
[...]

PR #47, opened 2015-10-09
Strip anonymous when we set username

wangmengyan95:
Strip anonymous when we set username, even we do not support anonymous user in JS SDK, we may encounter anonymous users created by android/iOS in cloud code, so we need to do this.
Fix issue #13

### DIFF
--- a/src/ParseUser.js
+++ b/src/ParseUser.js
@@ -277,6 +277,13 @@ export default class ParseUser extends ParseObject {
    * @return {Boolean}
    */
   setUsername(username: string) {
+    // Strip anonymity, even we do not support anonymous user in js SDK, we may
+    // encounter anonymous user created by android/iOS in cloud code.
+    var authData = this.get('authData');
+    if (authData && authData.hasOwnProperty('anonymous')) {
+      // We need to set anonymous to null instead of deleting it in order to remove it from Parse.
+      authData.anonymous = null;
+    }
     this.set('username', username);
   }
 
--- a/src/__tests__/ParseUser-test.js
+++ b/src/__tests__/ParseUser-test.js
@@ -401,4 +401,20 @@ describe('ParseUser', () => {
       done();
     });
   }));
+
+  it('strip anonymity when we set username', () => {
+    var user = new ParseUser();
+    var authData = {
+      anonymous : {
+        id : 'anonymousId'
+      }
+    }
+    user.set('authData', authData);
+    expect(user.get('authData').anonymous.id).toBe('anonymousId');
+
+    // Set username should strip anonymous authData
+    user.setUsername('test');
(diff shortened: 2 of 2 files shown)


stanleyw:
According to https://www.parse.com/docs/rest/guide#users-unlinking, the client needs to explicitly send `{ "anonymous": null}` to the server to remove the anonymous authData.  Not sending the key will cause the server to not modify the anonymous authData.

src/ParseUser.js
```diff
@@ -277,6 +277,13 @@ export default class ParseUser extends ParseObject {
    * @return {Boolean}
    */
   setUsername(username: string) {
+    // Strip anonymity, even we do not support anonymous user in js SDK, we may
+    // encoutner anonymous user created by android/ios in cloud code
```
stanleyw: encounter

src/ParseUser.js
```diff
@@ -277,6 +277,13 @@ export default class ParseUser extends ParseObject {
...
+    var authData = this.get('authData');
+    if (typeof authData !== 'undefined' &&  authData.hasOwnProperty('anonymous')) {
+      // We need to set anonymous to null instead of delete it in order to remve it from Parse
```
stanleyw: "instead of deleting it"
"to remove it"

src/ParseUser.js
```diff
@@ -277,6 +277,13 @@ export default class ParseUser extends ParseObject {
    * @return {Boolean}
    */
   setUsername(username: string) {
+    // Strip anonymity, even we do not support anonymous user in js SDK, we may
+    // encoutner anonymous user created by android/ios in cloud code
+    var authData = this.get('authData');
+    if (typeof authData !== 'undefined' &&  authData.hasOwnProperty('anonymous')) {
```
stanleyw: Extra space after &&

stanleyw: Looking at the existing style elsewhere in the SDK, this seems more consistent:

`if (authData && authData.hasOwnProperty('anonymous')) {`

stanleyw:
Sorry, noticed a few more nits.

src/ParseUser.js
```diff
@@ -277,6 +277,13 @@ export default class ParseUser extends ParseObject {
    * @return {Boolean}
    */
   setUsername(username: string) {
+    // Strip anonymity, even we do not support anonymous user in js SDK, we may
+    // encounter anonymous user created by android/iOS in cloud code.
+    var authData = this.get('authData');
+    if (authData && authData.hasOwnProperty('anonymous')) {
+      // We need to set anonymous to null instead of deleteing it in order to remove it from Parse.
```
stanleyw: "deleting"

stanleyw:
LGTM

Issue #73, opened 2015-10-29
Blockers migrating from SDK 1.5.x -> 1.6.x


TylerBrock:
I understand that the JS SDK is moving away from backbone style objects but there are a couple of pieces of that functionality that we'd like to see make their way into the 1.6 SDK at some point.

- Our application depends on the ability to silently update attributes of Parse.Objects

  Our existing parse objects (called models in our application's vernacular) all override `validate()` to provide client and server side validation for objects. We validate them before they are sent to Parse in the browser and after Parse receives them via a beforeSave hook. While it is not always the case, we often depend on being able to have objects in a non-valid state up until the point they are saved by constructing and updating them with `{silent: true}`.

   ```js
var thing = new Models.Thing({"awesome": true}), {silent: true});
thing.set("awesome", true, {silent: true});
   ```

  The reason we do this is because we want to build up a model with information we receive over time. For example a user wants to create a group in our system and we know the `organizationId` that it belongs to up front but don't have the other details such as group `name` and `location`.

  You can imagine having a <ModelForm> component in React that renders a form with data that you currently have about an object (that can't be saved as-is) but that can have the missing pieces filled in by the user to make it ultimately validate cleanly.

  The API doesn't have to be the same but it would be great if you could construct objects and manipulate their attributes freely **UNTIL** you try and save it, at which point everything must be valid. This surfaces as thousands of "Uncaught Error: Can't create an invalid Parse Object" in our application.

- Our application depends on defaults

  This one is easier to work around but as a result of the way we use validation (described in the first bullet point) we depend on some values having defaults.

[Here](https://gist.github.com/TylerBrock/67916527b5bb3b3ae7f9) is an example of a HustleModel and HustleModel's implementation where you can see defaults and constraints in use. I don't mind that this code is publicly visible.

How can we help add this functionality or work around is absence in versions 1.6+ of the SDK? I'd like to help in any way possible so that we can stay current with bugfixes, etc (the retrying of API requests in the new RESTController especially) and use more ES6.

andrewimm:
Defaults sounds like an easy enough thing to add; we support legacy style initialize() methods, so handling defaults should not be too different.

I definitely understand the need for explicitly avoiding validation; if anything, let's just enable it and call it something else. "Silent," to me, refers to Backbone's "change" event; simply passing `{ ignoreValidation: true }` as the options property.

[...]

PR #74, opened 2015-10-29
Add ability to ignore validation on Parse.Object constructor/set()

TylerBrock:
This is an attempt to resolve the first blocker in #73.

Not sure if the Flow typing is correct but I'm sure you'll let me know if it is not.

### DIFF
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -85,6 +85,7 @@ var singleInstance = (!CoreManager.get('IS_NODE'));
  * @constructor
  * @param {String} className The class name for the object
  * @param {Object} attributes The initial set of data to store in the object.
+ * @param {Object} options The options for this object instance.
  */
 export default class ParseObject {
   /**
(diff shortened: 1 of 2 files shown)


TylerBrock:
The coverage tool doesn't seem to understand the constructor arguments being split into multiple lines. Cest la vies, eh?

src/ParseObject.js
```diff
@@ -627,7 +631,7 @@ export default class ParseObject {
 
     // Validate changes
     var validation = this.validate(newValues);
```
andrewimm: Thought: it might make more sense to wrap this entire `var validation = ...; if (...) {}` block in an
`if (!options.ignoreValidation)` statement. That way, we don't needlessly run the validation logic.

TylerBrock: Yeah, for sure. I'll update the request.

I'm also not thrilled with the API. I think a deferValidation: true on the constructor that just defers it for all sets + construction might be better.

TylerBrock:
I'm also not thrilled with the API. I think a deferValidation: true on the constructor that just defers it for all sets + construction might be better.

TylerBrock:
Bump, lets get this rocking and rolling. I'd like to update our codebase without fear that this isn't going into the SDK. If there are issues or concerns lets discuss.

andrewimm:
I like it. I *think* every entry point is covered (there are a lot of constructors due to supporting the old `.extends()`. Tests seem good to me.

PR #99, opened 2015-11-25
Using proper undefined & null guard

herby:
`== null` / `!= null` is idiomatic and fast way to check for both `null` and `undefined`.
This is the guard that should be used prior to accessing a property of an object.
Otherwise, this blows when the object is `null`.

If accepted, pls release asap. Thanx.

### DIFF
--- a/src/ParsePromise.js
+++ b/src/ParsePromise.js
@@ -156,7 +156,7 @@ export default class ParsePromise {
       fn.call();
     };
     if (isPromisesAPlusCompliant) {
-      if (typeof process !== 'undefined' &&
+      if (process != null &&
           typeof process.nextTick === 'function') {
         runLater = function(fn) {
           process.nextTick(fn);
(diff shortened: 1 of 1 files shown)


herby:
FUCK your CLA, just pls add the fix. Thanks.

Facebook Community Bot wrote:

andrewimm:
Good catch. The reason your PR failed the tests is because you also modified an unrelated `typeof process` check. As you can imagine, we don't accept PRs that don't pass all tests.
I've landed a fix that only adjusts the one necessary check, and includes a number of new unit tests as well. It'll probably be released sometime in the next week as 1.6.10
In the future, if you choose to submit PRs, please abide by our process and our code of conduct, both outlined in https://github.com/ParsePlatform/Parse-SDK-JS/blob/master/CONTRIBUTING.md

herby:
I never contribute to anything with code of conduct, sorry.

Dňa 1. decembra 2015 19:44:09 CET používateľ Andrew Imm <notifications@github.com> napísal:

PR #110, opened 2015-12-04
Fix documentation parsing error

vreemt:
the documentation incorrectly tries to use the `success` function when rendering the `objectClass`
this makes the display of the parameter more consistent


andrewimm:
Good catch, appreciate the PR!

vreemt:
Is there any way I can test if it actually looks ok?
I figured making it match what's in other doc blocks (e.g. the ones for
Relations and Objects) would work, but no idea about testing it... feeling
like a bit of a noob (;
Thanks

2015-12-04 17:03 GMT+00:00 Andrew Imm <notifications@github.com>:

PR #116, opened 2015-12-10
Updating doc comment for updated 'when' format

braker1nine:
This is my bad. This should have been in #115


braker1nine:
Hmm, this should probably be reverted since the changes in #115 were reverted. Now the docs incorrectly characterize how the `when` callback works. I don't want some other dev banging their head expecting it to work one way when it actually works another

PR #123, opened 2015-12-17
Handle ACL on save

wangmengyan95:
It seems we do not handle ACL on save. Normally for a save/update, the server just return `createAt, objectId` and `updateAt`, so it is fine. However, if the developers enable beforeSave for this class, the server will return the full object. In this situation, if we do not handle ACL, it will make `object.getACL()` return `null`.
I follow the way we handle ACL in `_finishFetch` to handle ACL. Not sure it is the best way to do that.


wangmengyan95:
@hallucinogen not sure andrew has time to review, so could you also take a look?
cc @peterdotjs .

andrewimm:
lgtm, ship it

Issue #139, opened 2015-12-27
Parse.Object.clone() it's not working


yutin1987:
https://github.com/ParsePlatform/Parse-SDK-JS/blob/master/src/ParseObject.js#L750

```
clone(): any {
    var clone = new this.constructor();
    console.log(1, this.attributes);
    console.log(2, clone.toJSON());
    if (clone.set) {
      clone.set(this.attributes);
    }
    console.log(3, clone.toJSON());
    if (!clone.className) {
      clone.className = this.className;
    }
    console.log(4, clone.toJSON());
    return clone;
}
```

console.log
```
1 Object {after: 0, before: 0, createdAt: Sat Dec 26 2015 21:53:14 GMT+0800 (CST), createdBy: ParseUser, enable: false…}
2 Object {}
3 Object {after: 0, before: 0, createdBy: Object, enable: false, monthlyLimit: 0…}
4 Object {}
```

PR #140, opened 2015-12-27
bugfix object clone() method

yutin1987:
should precede setting className
#139


andrewimm:
Good catch. In cases where the parent constructor is not setting the class name, setting attributes will end up placing them in a throwaway container.

PR #152, opened 2016-01-05
Using serverUrl constant for batch requests.

peterdotjs:


peterdotjs:
Missing `package.json` dependency - creating another pr.

PR #153, opened 2016-01-05
Using serverUrl constant for batch requests.

peterdotjs:


andrewimm:
Importing `url` seems like overkill to me.

peterdotjs:
@andrewimm I could take strip out the parsing pieces for the path rather than use the module if you concerned about bloat.

andrewimm:
```js
// Remove the protocol
var url = serverUrl.replace(/https?:\/\//, '');
var path = url.substr(url.indexOf('/'));
```
should work, right?

peterdotjs:
That'll work for the current `serverUrl` config value but I'm also thinking about any corner cases that users might change the value to.

andrewimm:
And what would those cases be? This works for urls with and without protocol. In a valid URL, slashes can only exist between the protocol and the domain, or within the path, so once the protocol is removed, the first slash is the beginning of the path. Beyond that, setting an invalid value to this will break the entire SDK anyways, so it's not unreasonable to us to say that only valid URLs are supported.

andrewimm:
Actually, we need the "does it end with a slash?" check to make it really work. Otherwise a value like "api.parse.com" fails. Otherwise, I think it's relatively bulletproof.

peterdotjs:
We'll need to handle the host case as well: `'http://localhost.com:8080/p/a/t/h`

andrewimm:
@peterdotjs the code works for that:
var url = 'https://localhost.com:8080/p/a/t/h'.replace(/https?:\/\//, ''); // url = localhost.com:8080/p/a/t/h
var path = url.substr(url.indexOf('/')); // path = '/p/a/t/h'

Feel free to unit-test this thing into oblivion

andrewimm:
My only concern is adding a 650 LOC library to do something that can be done in 1-2 lines.

peterdotjs:
@andrewimm agreed that there are a lot of things in the lib we won't use. I'm sure there will be other cases that come up (`//foo/bar `) but this should work for most cases.

andrewimm:
For now, treat it as a private API (it's not like anyone could have their own Parse address anyways, this is basically for our internal tests for now). If we ever publicize it, we can establish expectations and shore up the code, but immediately, I like treating it like a black box.

andrewimm:
Also, once you land this change, I'd like to cut a release for 1.6.14, which will probably be the last one before 1.7

peterdotjs:
Ready for merging.

andrewimm:
LGTM, merging.

Issue #136, opened 2015-12-21
Release 1.7


andrewimm:
Planned changelog:
- [x] Store server data and pending ops on-object when single instance mode is disabled. This will prevent memory leaks. (structural change only, does not affect APIs or behavior)
- [x] Default to setting A+ Compliance to `true`. See #57 for more details. (landed in 96e211d4fad49929d3db522254d3540c14ee3a77)
- [x] Enable native-style construction of `Parse.Promise`. See #57 for more details. (landed in 5e5c444d0bd02d5955c8b43b6257ca11b74eeb14)
- [x] Re-commit #115, changing the behavior of `Parse.Promise.when()`. The response provided to the asynchronous callback will match the style of its arguments. (landed in 7da2856b18ca1672766612700139757300ecbf62)
- [x] Add `revert()` instance method to ParseObject (landed in 7c16dee0f90a6d85bb409851de86107a04007c06)

When these changes have all been committed to master, and end-to-end tested, we'll release 1.7.0

TylerBrock:
This would be great. Thanks Andrew

grantland:
I :heart: #115

mohnazemi:
Changelogs were not updated for the last several releases on the Downloads page. Is there a reason ?

andrewimm:
@mohnazemi the source of truth for changelogs will always be here on this repo: https://github.com/ParsePlatform/Parse-SDK-JS/releases

I'll update the downloads page to include the last few releases

srameshr:
@andrewimm Caching built into the JS SDK would be a very useful addition. App's built for the webview in IOS and Android platform would have a huge boost. In the docs you clearly say don't use REST API when you have native sdk for your platform.

andrewimm:
@srameshr we are working out a number of strategies for caching, but it will not be an official feature until at least 1.8. It will most likely end up being a plug-in that can be added to the SDK, rather than a change to the SDK itself.

srameshr:
Any particular date on which, this would be available?

andrewimm:
@srameshr As a rule, we don't commit to dates for releases. Even less so for features where we're still unsure what to implement.

As it is, it's trivial to implement your own cache in your own application:
```js
updateCache(object) {
  let entry = object.className + ':' object.id;
}
```

andrewimm:
@srameshr As a rule, we don't commit to dates for releases. Even less so for features where we're still unsure what to implement.

[...]

PR #154, opened 2016-01-05
Allow ES6-style Promise construction

andrewimm:
Allows Parse.Promises to be constructed in the ES6 "executor" style, without intervening with old construction.

```js
// Valid:
return new Parse.Promise((resolve, reject) => {
  myAsyncFunction((result) => {
    resolve(result);
  });
});

// Also still valid:
let p = new Parse.Promise();
myAsyncFunction((result) => {
  p.resolve(result);
});
return p;
```

cc @grantland

### DIFF
--- a/src/ParsePromise.js
+++ b/src/ParsePromise.js
@@ -26,11 +26,15 @@ var isPromisesAPlusCompliant = true;
  * @constructor
  */
 export default class ParsePromise {
-  constructor() {
+  constructor(closure) {
     this._resolved = false;
     this._rejected = false;
     this._resolvedCallbacks = [];
     this._rejectedCallbacks = [];
+
+    if (typeof closure === 'function') {
+      closure(this.resolve.bind(this), this.reject.bind(this));
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


andrewimm:
Part of the launch checklist for #136

andrewimm:
Also added the following static constructors from the ES6 spec:
```js
Parse.Promise.resolve(value);
Parse.Promise.resolve(thenable); // Chains the new promise to the end of the old one, essentially a noop
Parse.Promise.reject(error);
```

src/ParsePromise.js
```diff
@@ -26,11 +26,15 @@ var isPromisesAPlusCompliant = true;
  * @constructor
  */
 export default class ParsePromise {
-  constructor() {
+  constructor(closure) {
```
peterdotjs: can we use `executor` as param name for consistency?

andrewimm: :+1: absolutely. I was calling it "closure" before I read the spec and determined it was deemed an "executor"

peterdotjs:
LGTM.

Issue #111, opened 2015-12-07
Memory leak when using ParseQuery in a loop


ghugues:
The title of this issue might be misleading as I don't know exactly what is happening.

I originally tried to write a background job that loops over all the rows of a table. Now I'm running this on my computer until I find a fix (which is why I use the node environement).

I've tried to use both `ParseQuery.each()` and my own version of this method (using `find()` and a chain of Promises), and in both cases the memory usage of node increases until it reaches the 1.4G limit and node crashes. I think that the objects that are queried are never released either because of a memory leak or because of a chain of retains that I have yet to understand.
I've tried to run node with `--expose-gc` and call `gc()` at different points, but it doesn't work.

Here is a sample code that will reproduce the issue. It queries all the objects of a table and then starts over, indefinitely (in case there are not enough objects in the table to make the script reach the 1.4G limit).
if `useParseQueryEach` is set to `true`, it will use the ` ParseQuery.each()` otherwise it will use my own implementation (`queryEachWithBatchSize`).

If this is not related to the Parse SDK but rather to the way I implemented this, please let me know, I will repost it on StackOverflow.

```js
var Parse = require('parse/node').Parse;
var TestObject = Parse.Object.extend("TestObject");

var useParseQueryEach = true;
var queriesBatchSize = 1000;
var statusReportInterval = 1000;

Parse._initialize("APPLICATION_ID", "JAVASCRIPT_KEY", "MASTER_KEY");
Parse.Cloud.useMasterKey();

return queryAllTestObjects(1).then(null, function(error) {
	console.error("Script failed with error " + error.code + " : " + error.message);
});


function queryAllTestObjects(loopCount)
{
	console.log("Will process all test objects (loop count = " + loopCount + ")...");

	var processedTestObjectsCount = 0;
	var allTestObjectsQuery = new Parse.Query(TestObject);

	var callbackFunction = function(testObject) {
		if (++processedTestObjectsCount % statusReportInterval === 0) {
			console.log(processedTestObjectsCount + " test objects processed...");
		}
	};
	var completionHandler = function() {
		console.log(processedTestObjectsCount + " test objects processed.");
		return queryAllTestObjects(loopCount + 1);
	};

	if (useParseQueryEach) {
		return allTestObjectsQuery.each(callbackFunction, {batchSize:queriesBatchSize}).then(completionHandler);
	} else {
[...]

PR #164, opened 2016-01-11
Split ObjectState into utilities and storage implementations

andrewimm:
ObjectState used to both store object data *and* the perform the functionality to mutate it.
Now, to facilitate different behavior between single-instance mode and unique-instance mode, ObjectState simply contains the mutation behavior, independent of how it's stored.
This PR introduces `SingleInstanceState` and `UniqueInstanceState`, storage layers that are directly accessed by ParseObject, and use ObjectState as a utility.

This change is chiefly to prevent the memory leaks reported in #111 

Essentially, there are 4 pieces:
 - Isolation of ObjectState into utility functions that have their data sources passed in
 - Implementation of SingleInstanceState and UniqueInstanceState
 - Reimplementation of tests for both of these storage systems
 - Branching of ParseObject logic to interact with one or the other

I'll be adding more tests, but things look good for now.


andrewimm:
Final code piece for #136

andrewimm:
This change also brings ObjectState and the two State implementations to complete code coverage in unit tests.

andrewimm:
Just tried allocating 10MM objects with UniqueInstanceState while profiling node, watched it continually GC and recover memory :D

roddylindsay:
woohoo

TylerBrock:
w00t

andrewimm:
cc @hallucinogen and @wangmengyan95 for review as well

src/ParseObject.js
```diff
@@ -131,9 +132,10 @@ export default class ParseObject {
   /** Prototype getters / setters **/
 
   get attributes(): AttributeMap {
-    return Object.freeze(
-      ObjectState.estimateAttributes(this.className, this._getStateIdentifier())
-    );
+    let attributes = singleInstance ?
```
hallucinogen: I think it would be better if we abstract this into `InstanceStateController` that can be configured by `CoreManager.js`

andrewimm: @hallucinogen I thought about that too, but they take completely different arguments -- one uses a set of strings as identifiers, the other uses an object. We could pass the Object and introspect the identifiers in most, but not all of the cases

src/ParseObject.js
```diff
@@ -1046,7 +1093,11 @@ export default class ParseObject {
   /** Static methods **/
 
   static _clearAllState() {
-    ObjectState._clearAllState();
+    if (singleInstance) {
+      SingleInstanceState._clearAllState();
```
hallucinogen: Unless I miss something, this is the only place that doesn't accept `this.className, this._getStateIdentifier()`. And even this is abstract-able. What prevent us from passing Object (e.g. `this`) to `SingleInstanceState`'s methods?

src/ParseObject.js
```diff
@@ -316,11 +342,16 @@ export default class ParseObject {
...
-      this.id = serverId;
-      delete this._localId;
-      if (oldState) {
-        ObjectState.initializeState(this.className, this._getStateIdentifier(), oldState);
+      if (singleInstance) {
+        var oldState = SingleInstanceState.removeState(this.className, this._getStateIdentifier());
+        this.id = serverId;
+        delete this._localId;
+        if (oldState) {
+          SingleInstanceState.initializeState(this.className, this._getStateIdentifier(), oldState);
```
hallucinogen: Even here, you can pass `this` to `UniqueInstanceState` and just do nothing in `removeState` and `initializeState`.

andrewimm:
So, we could probably abstract everything away and make this a controller, but one of the core reasons that ObjectStore exists, and why single-instance mode is addressable by a className + ID mapping, is to eventually publicize APIs that provide functional access over Parse Objects (a la Parse+React) without requiring any new code. Whether we consider the idea of SDK modularization or not, I like to think that it would be possible to build a reasonable lightweight application with just a Store, ParseOp, and RESTController.

I also recognize the value of shared interfaces, so here's what I think we should do. Rather than make the first argument a `ParseObject` (pulling in unnecessary dependencies), it should be an object that implements the interface `{ className: string, id: string }`. Since a Parse Object implements this interface, we can always pass it as the first argument.

andrewimm:
Continuing... another possible option is to replace the way we use `_getStateIdentifier`. Right now it returns the object ID; what if we used it to return the object I mention above instead?

In single instance mode, it returns an object of the shape `{ className: string, id: string }`. In unique instance mode, it returns the ParseObject itself.

This is my favorite approach thusfar

grantland:
This is going to be confusing, since the JS SDK is one level deeper than the other SDKs:

* JS `ParseObject`
* JS `ObjectState` == Android/iOS/.NET `ParseObject`
* JS `UniqueInstanceState`/`SingleInstanceState` == Android/iOS/.NET `ParseObject.State`

Does ParseReact publicize `ObjectState` directly? If not, it'd be great if we can rename things to make it less confusing moving between codebases.

andrewimm:
`ObjectState` is not documented, so we can rename things. I'd support renaming it, since it's really a set of state modifiers, not the state itself anymore.

hallucinogen:
> So, we could probably abstract everything away and make this a controller, but one of the core reasons that ObjectStore exists, and why single-instance mode is addressable by a className + ID mapping, is to eventually publicize APIs that provide functional access over Parse Objects (a la Parse+React) without requiring any new code.

It's still possible to eventually publicize APIs that provide functional access over `ParseObject`s (a.k.a publicize our current `ObjectState`) while still abstracting the State store. IMO, it's even more powerful if we abstract it as controller (and the API will allow people to build `localStorage` based State store! #randomidea #dontmindthis).

I agree with passing `_getStateIdentifier`.

grantland:
Maybe `ReactObject` instead and `ObjectState` and `Single`(`Instance`)`ObjectState`?

andrewimm:
Object State is now controlled by ObjectStateController, and swappable like any other controller. This should be good for expandability.

Also renamed `ObjectState` to `ObjectStateMutations`, since it's now a set of standardized mutation functions, and does not affect any state.

src/SingleInstanceState.js
```diff
@@ -0,0 +1,130 @@
...
+    return null;
+  }
+  delete objectState[obj.className][obj.id];
+  return state;
+}
+
+export function getServerData(obj: ObjectIdentifier): AttributeMap {
+  let state = getState(obj);
+  if (state) {
+    return state.serverData;
```
peterdotjs: Do we want to be consistent in the default return value for all these functions? 
`return state.serverData || {};` 
like we did in `getState`?

peterdotjs:
Really like the extensibility now with `ObjectStateController`.

src/SingleInstanceState.js
andrewimm: Can you explain further? Each function's "default state" has a pretty specific meaning that is leveraged in the code. In general, I don't think that's the bit that should be up for review now; those are the things that are remaining the same before & after this diff.

andrewimm: Oh, for your specific example, it's an invariant that `state.serverData` exists if `state` exists, check out the typedef at https://github.com/ParsePlatform/Parse-SDK-JS/pull/164/files#diff-5c0499aa286ed5bb8b6f6ad90b87830cR26

peterdotjs: ah i missed that - looks like it's all covered then.

src/CoreManager.js
```diff
@@ -43,6 +43,24 @@ type ObjectController = {
...
+  getState: (obj: ParseObject) => ?State;
+  initializeState: (obj: ParseObject, initial?: State) => State;
+  removeState: (obj: ParseObject) => ?State;
+  getServerData: (obj: ParseObject) => AttributeMap;
+  setServerData: (obj: ParseObject, attributes: AttributeMap) => void;
+  getPendingOps: (obj: ParseObject) => Array<OpsMap>;
+  setPendingOp: (obj: ParseObject, attr: string, op: ?Op) => void;
+  pushPendingState: (obj: ParseObject) => void;
+  popPendingState: (obj: ParseObject) => OpsMap;
+  mergeFirstPendingState: (obj: ParseObject) => void;
```
hallucinogen: Hmm, you sure this is correct? `mergeFirstPendingState` of `ObjectStateController` accepts `ParseObject`.....

src/ParseObject.js
```diff
@@ -357,12 +378,13 @@ export default class ParseObject {
...
 
-    ObjectState.commitServerChanges(this.className, this._getStateIdentifier(), changes);
+    stateController.commitServerChanges(this._getStateIdentifier(), changes);
   }
 
   _handleSaveError() {
     var pending = this._getPendingOps();
-    ObjectState.mergeFirstPendingState(this.className, this._getStateIdentifier());
+    let stateController = CoreManager.getObjectStateController();
+    stateController.mergeFirstPendingState(this._getStateIdentifier());
```
hallucinogen: ... meanwhile in practice, it accepts `stateIdentifier`....

src/ParseObject.js
```diff
@@ -172,22 +177,26 @@ export default class ParseObject {
...
-    if (typeof this.id === 'string') {
-      if (singleInstance) {
-        return this.id;
+    if (singleInstance) {
+      let id = this.id;
+      if (!id) {
+        id = this._getId();
       }
-      return this.id + '_' + String(this._objCount);
+      return {
```
hallucinogen: ... which can be `{ id: string, className: string}`, and not always `ParseObject`.

hallucinogen:
Loving it! One question though

andrewimm:
I'll let @hallucinogen give the green light on this one.

src/ParseObject.js
```diff
@@ -34,10 +33,12 @@ import {
 import ParsePromise from './ParsePromise';
 import ParseQuery from './ParseQuery';
 import ParseRelation from './ParseRelation';
+import * as SingleInstanceState from './SingleInstanceState';
```
hallucinogen: I hope you won't hate me for this. But this should be `SingleInstanceStateController`. I feel strongly about this since `SingleInstanceState` sounds like model, while `SingleInstanceStateController` clearly... a controller.

src/ParseObject.js
```diff
@@ -34,10 +33,12 @@ import {
...
 import unique from './unique';
+import * as UniqueInstanceState from './UniqueInstanceState';
```
hallucinogen: Ditto

src/ParseObject.js
andrewimm: nah, makes sense, though with every extra word I worry about things getting too Java-y. I'll go find-replace.

AbstractInstanceStateControllerFactory, here we come!

andrewimm:
Boomdone. Really missing phab macros right about now

src/CoreManager.js
```diff
@@ -207,6 +218,107 @@ module.exports = {
     return config['ObjectController'];
   },
 
+  setObjectStateController(controller: ObjectStateController) {
+    if (typeof controller.getState !== 'function') {
+      console.log(controller);
```
hallucinogen: productioncleanup

hallucinogen:
LGTM. One last nits, but accepting anyway.

andrewimm:
tyvm

PR #179, opened 2016-01-21
Make sure the className is correct for results on unknown relations

andrewimm:
If the target className of a relation is not known ahead of time, and the server does not pass the className back on the raw JSON of Parse Object results, the query results end up with the wrong className.
The className to be used comes back as a top-level key on the query result payload. This diff makes sure that value is used, if available.


peterdotjs:
lgtm.

PR #186, opened 2016-01-26
Added webworker support

shanebdavis:
To make Parse work in a webworker, it needs access to the localStorage object on the main/browser thread. This small patch to Parse allows clients to provide a custom asyncLocalStorage implementation.

How to use:

define self.asyncLocalStorage, in the worker thread, BEFORE requiring parse.js
asyncLocalStorage should support: getItem, setItem, removeItem and clear
asyncLocalStorage should take exactly the same inputs as localStorage
asyncLocalStorage should return promises (ES6 or Parse compatible) for all four functions which resolve to the same results localStorage would return.


andrewimm:
Appreciate the PR. Beyond style discussions, there are a number of concerns that need to be addressed around runtime detection of an implementation-specific global variable, which seems unsafe. Additions to the APIs here are semi-permanent, and ideally we try to aim for solutions that work out of the box. It seems like what you're doing here – adding a "proxy" storage implementation, that in turn points to an object implementing localstorage APIs – could more easily be implemented by providing a custom Storage Controller for your WebWorker environment. This is why Controllers are hot-swappable, and it is the pattern we encourage. It doesn't have to happen at compile time, so it doesn't need to be a part of the block at the bottom of `Storage.js`. At initialization time, when you call `Parse.initialize`, you can also set up whatever custom Controllers you need.
```js
Parse.CoreManager.setStorageController(myController)
```

Since this really is a proxy controller which could be replaced with individual controllers, I don't think we'll merge it in its current state. I hope you understand

andrewimm:
Also, Parse Promises are interoperable with ES6 promises, so there shouldn't be a need to convert between them.

shanebdavis:
Thank you for your reply! Setting a custom storage controller sounds
perfect, and I agree, a much better solution. I didn't find any
documentation on that. Is there any?

Thank you!
-Shane

andrewimm:
Good point, this is definitely the third or fourth time a scenario involving custom storage has come up here. Let me throw together something in the docs!

shanebdavis:
Thanks! And I tried it, it worked perfect.

-Shane

shanebdavis:
FYI: Parse Promises do not appear to be interoperable with ES6 promises on
Safari or Firefox. Oddly, they do appear to interop on Chrome. In order to
get my webworker working on all platforms, I had to add back in
my es6PromiseToParsePromise function for my custom storage controller.

The error I get is "TypeError:
controller.find(...).then(...)._thenRunCallbacks is not a function".

-Shane

andrewimm:
Ah, correct. You can use Parse Promises within ES6 promises, but you can't yet use ES6 promises on the internals of Parse methods (at least, not until we remove support for the legacy Backbone-style callbacks)

shanebdavis:
And now Parse is going away :(. That is a major disappointment. I wonder
why management didn't sell it so someone else could keep it going... Thanks
for your help.

PR #192, opened 2016-01-28
Option to specify https for Parse.File url()

jsonmaur:
This is a simple feature, and it would go a long way to have it implemented. Since files on `http://files.parsetfss.com` can also be accessed at `https://files.parsetfss.com`, I was surprised this isn't available yet. Been running into problems using `Parse.File` with images on a site using https, as it throws a Mixed Content warning. There may be a more consistent way to handle the api, but this works fine in my opinion.

### DIFF
--- a/src/ParseFile.js
+++ b/src/ParseFile.js
@@ -137,9 +137,12 @@ export default class ParseFile {
    * @param {Object} options An object to specify url options
    * @return {String}
    */
-  url(options?: { secure?: boolean }): ?string {
+  url(options?: { forceSecure?: boolean }): ?string {
     options = options || {};
-    if (options.secure) {
+    if (!this._url) {
+      return;
+    }
+    if (options.forceSecure) {
       return this._url.replace(/^http:\/\//i, 'https://');
     } else {
       return this._url;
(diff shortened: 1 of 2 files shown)


src/ParseFile.js
```diff
@@ -134,10 +134,16 @@ export default class ParseFile {
    * Gets the url of the file. It is only available after you save the file or
    * after you get the file from a Parse.Object.
    * @method url
+   * @param {Object} options An object to specify url options
    * @return {String}
    */
-  url(): ?string {
-    return this._url;
+  url(options?: { secure?: boolean }): ?string {
```
andrewimm: How about changing `secure` to `forceSecure`? If the server ever decided to default to https, the API would still be correct.

src/ParseFile.js
```diff
@@ -134,10 +134,16 @@ export default class ParseFile {
...
+    options = options || {};
+    if (options.secure) {
```
andrewimm: `if (this._url && options.forceSecure)`

If the file doesn't have a URL set, this would have previously returned undefined; now, it'll crash.

andrewimm:
Couple comments, but otherwise looks good. Appreciate that you added tests to cover this functionality.

jsonmaur:
Yeah,  `forceSecure` is better. How's that look?

andrewimm:
Awesome, merging this. Look for it in 1.7.1, which I'll probably release on Monday.

PR #197, opened 2016-02-02
Add the ability to pass an installationId in the backbone-style options on signup() and login().

wuotr:
I made it possible to pass along the _installationId_ (string) as property on the backbone-style options object. This is especially useful on the _login()_ and _signup()_ API when managing installations upfront.

If no _installationId_ is passed, the API acts as normal.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules
 test_output
 *~
 .DS_Store
+.idea/
--- a/src/CoreManager.js
+++ b/src/CoreManager.js
@@ -21,6 +21,7 @@ import type { PushData } from './Push';
 type RequestOptions = {
   useMasterKey?: boolean;
   sessionToken?: string;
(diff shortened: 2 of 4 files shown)


andrewimm:
I need to think about the risk involved here. We've discussed this as the next step for server-side code, but it might need a little more discussion.

In the meantime, please don't include changes to `package.json` in your PRs unless they are the subject of the PR. Incrementing the version number should be handled by the team releasing this library.

wuotr:
@andrewimm Allright, thank you for taking my PR in consideration ;-). And sorry for the version number indeed... I kinda updated it without thinking. (I have removed that commit now.) Cheers!

andrewimm:
This all looks good to me. I'm going to manually merge this into the `1.8.0-release` branch, rather than do it here from GH. Since this represents a change in API, I'd like to wait to release it on a minor version increment.

PR #206, opened 2016-02-12
Clear old data on fetch

andrewimm:
Handles `Object.fetch()` and Query results


drew-gross:
Seems to make sense. Integration tests sounds like a good idea, feel free to merge this before or after making those.

andrewimm:
Integration tests run against the latest "master," so I'll merge now.

PR #209, opened 2016-02-17
New API for creating a new instance of an object

andrewimm:
`ParseObject.prototype.newInstance()` creates a new, memory-unique instance of a Parse Object, as though it had been fetched again from the server. This is only really relevant in Unique Instance mode.

### DIFF
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -775,7 +775,7 @@ export default class ParseObject {
   }
 
   /**
-   * Creates a new model with identical attributes to this one.
+   * Creates a new model with identical attributes to this one, similar to Backbone.Model's clone()
    * @method clone
    * @return {Parse.Object}
    */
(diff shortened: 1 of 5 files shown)


andrewimm:
@TylerBrock would love your input here

TylerBrock:
Looks good, would duplicateState work for objects with read only attributes? I assume setting attrs on serverData directly bypasses that.

andrewimm:
All validation is handled within the ParseObject instance itself, but let me throw together some test cases for ParseUser and ParseSession

TylerBrock:
Thanks Andrew. Are you thinking this slated for 1.7.2 or 1.8.0?

andrewimm:
1.8.0, since it represents a new API. I'm currently adding one last change, and then probably pushing out an RC. I like the RC approach, I think I'll adopt it for all future minor version increments.

PR #233, opened 2016-03-21
Fix getting WebSocket implemetation in react-native

farwayer:


wangmengyan95:
Thanks for the fix :+1:

xcarpentier:
+1

Issue #232, opened 2016-03-20
Exception in ParseLiveQuery in react-native


farwayer:
`getSessionToken()` in `ParseLiveQuery` calls **sync** `currentUser()` and throw exception in react-native (with **async** storage)

andrewimm:
@peterdotjs @wangmengyan95 is it possible for you to make the method in question, and whatever's calling it async? All current user fetching internally should be using the async methods, to support async storage systems.

peterdotjs:
@andrewimm I'll make the change to use `currentUserAsync`.

andrewimm:
Okay, my concern is that the caller functions might be synchronous, and APIs would need to change to adjust for this.

peterdotjs:
Currently we're using `currentUser` to init the `DefaultLiveQueryClient` [here](https://github.com/ParsePlatform/Parse-SDK-JS/blob/master/src/ParseLiveQuery.js#L116).  But we are using the sessionToken again [here](https://github.com/ParsePlatform/Parse-SDK-JS/blob/master/src/ParseLiveQuery.js#L143). Assuming we want to query for the token on each subscribe the interface for LiveQueryClient.subscribe will need to change. @wangmengyan95

farwayer:
Thanks for fast response guys. I found one more bug with WebSockets implementation in react-native. Look like live queries are not very popular functionality in RN :)
PR #233

PR #234, opened 2016-03-21
Updating Parse.LiveQuery to use currentUserAsync.

peterdotjs:
Addressing #232.

### DIFF
--- a/src/LiveQuerySubscription.js
+++ b/src/LiveQuerySubscription.js
@@ -102,8 +102,11 @@ export default class Subscription extends events.EventEmitter {
    * @method unsubscribe
    */
   unsubscribe() {
-    var liveQueryClient = CoreManager.getLiveQueryController().getDefaultLiveQueryClient();
-    liveQueryClient.unsubscribe(this);
-    this.emit('close');
+    let _this = this;
+    CoreManager.getLiveQueryController().getDefaultLiveQueryClient().then((liveQueryClient) => {
+      liveQueryClient.unsubscribe(_this);
+      _this.emit('close');
+      this.resolve();
+    });
   }
 }
--- a/src/ParseLiveQuery.js
+++ b/src/ParseLiveQuery.js
@@ -123,11 +123,9 @@ let DefaultLiveQueryController = {
       // Register a default onError callback to make sure we do not crash on error
       defaultLiveQueryClient.on('error', (error) => {
         LiveQuery.emit('error', error);
-      });
-      defaultLiveQueryClient.on('open', () => {
+      }).on('open', () => {
         LiveQuery.emit('open');
-      });
-      defaultLiveQueryClient.on('close', () => {
+      }).on('close', () => {
         LiveQuery.emit('close');
       });
 
--- a/src/ParseLiveQuery.js
+++ b/src/ParseLiveQuery.js
@@ -145,7 +145,7 @@ let DefaultLiveQueryController = {
     });
   },
   subscribe(query: any): EventEmitter {
-    let subscriptionWrap = new events.EventEmitter();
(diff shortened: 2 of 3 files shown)


src/ParseLiveQuery.js
```diff
@@ -69,16 +70,19 @@ LiveQuery.on('error', () => {
 export default LiveQuery;
 
 let getSessionToken = () => {
-  let currentUser = CoreManager.getUserController().currentUser();
-  let sessionToken;
-  if (currentUser) {
-    sessionToken = currentUser.getSessionToken();
-  }
-  return sessionToken;
+  let promiseUser = CoreManager.getUserController().currentUserAsync();
```
wangmengyan95: It is wired to create promiseXXX for each inner promises, why not just chain them?

src/ParseLiveQuery.js
```diff
@@ -86,65 +90,76 @@ let DefaultLiveQueryController = {
...
+
+      return ParsePromise.as(defaultLiveQueryClient);
     });
-    return defaultLiveQueryClient;
   },
   open() {
-    let liveQueryClient = getLiveQueryClient();
-    liveQueryClient.open();
+    getLiveQueryClient().then((liveQueryClient) => {
+      return ParsePromise.as(liveQueryClient.open());
```
wangmengyan95: Discuss offline, we do not need to wrap it around a new promise.

joshblour:
I tried this out in my react-native app. `query.subscribe()` now returns a promise and the following example fails:
```js
		let subscription = query.subscribe()
		subscription.on('open', () => {
 			console.log('subscription opened');
		});
```

peterdotjs:
I'm working on changing this to keep the interface the same - when landed .subscribe will return a subscription.

src/LiveQuerySubscription.js
```diff
@@ -102,8 +102,11 @@ export default class Subscription extends EventEmitter {
    * @method unsubscribe
    */
   unsubscribe() {
-    var liveQueryClient = CoreManager.getLiveQueryController().getDefaultLiveQueryClient();
-    liveQueryClient.unsubscribe(this);
-    this.emit('close');
+    let _this = this;
```
wangmengyan95: Do we need this? I think the arrow function auto bind this for you?

src/ParseLiveQuery.js
```diff
@@ -86,65 +89,103 @@ let DefaultLiveQueryController = {
...
+      // Get currentUser sessionToken if possible
+      defaultLiveQueryClient = new LiveQueryClient({
+        applicationId,
+        serverURL: liveQueryServerURL,
+        javascriptKey,
+        masterKey,
+        sessionToken,
+      });
+      // Register a default onError callback to make sure we do not crash on error
+      defaultLiveQueryClient.on('error', (error) => {
```
wangmengyan95: I do not know we can chain the `on` method, let's clean this.

src/ParseLiveQuery.js
wangmengyan95: Any reason you want to keep this?

src/LiveQuerySubscription.js
peterdotjs: arrow does auto bind but in this case we'll need to distinguish between the promise and the subscription context.

src/ParseLiveQuery.js
peterdotjs: what is the issue here? I thought it's more readable than chaining directly from the above.

src/ParseLiveQuery.js
peterdotjs: sure ill update this

PR #243, opened 2016-03-27
Adds Hooks API on parse/node

flovilmart:
- Adds hooks API to parse/node in perspective to remove the hooks API from experimental on parse-server
- Will remove Parse.Hooks from parse-server once merged
- Will keep E2E tests in parse-server to make sure the API is compliant
- Tested on parse.com and create/update/remove works perfectly without any changes


andrewimm:
If we move to 1.9.0, I'd be happy to merge this in. I just want to take another look at it before we do so.

flovilmart:
That would be nice to remove it from parse-server. I'll update it when you release

PR #257, opened 2016-04-13
Corrected ParseLiveQuery for react native

reyalpsirc:
The EventEmitter class of React Native does not allow nested events and so, the `query.subscribe()` was not working.


andrewimm:
Sorry for the delay. I appreciate you catching this. We'll see if we can cut a new release soon.

PR #274, opened 2016-05-08
Automatically export the right the required SDK for a given platform

### LINKED ISSUE #269: Problems using the Parse-JS-SDK universally
Hey @andrewimm I've got what I think is a unique situation I've gotten our codebase into but I was wondering if you had any wisdom with regards to loading the sdk universally.

In our codebase we share our Parse model definitions everywhere:

 - our front end code, which must `require('parse')`
 - our back-end code, which must `require('parse/node')`
 - our mobile code (react-native), which must `require('parse/react-native')`

As a hack solution, all model definitions and all code that uses the JS sdk take advantage of a library I wrote called `parse-shim` which we require all over our codebase so that in the front end it loads `parse` and on the backend it loads `parse-node`, etc... The code boils down to a trivial module:

```js
if (onReactNative()) {
  module.exports = require('parse/react-native');
} else if (typeof window === 'undefined') {
  module.exports = require('parse/node');
} else {
  module.exports = require('parse');
}
```

This lets us create models like this one below, with custom instance methods, etc and use them everywhere we use parse:

```js
const Parse = require('parse-shim');

export default class Obj extends Parse.Object { ... }
```

However, when using `parse-server` and mounting that within our existing Express application it's causing some interesting problems that are obvious in retrospect. Namely `Parse._encode()` checks to see if the value passed to it is a Parse.Object (as defined inside of this library) which of course differs from the module we export from `parse-shim`.

It seems as though parse-server's version of Parse.Object is different from ours (because it doesn't consider the parse-shim and parse-server module prototypes to be the same via `instanceof`). The result is that `Parse._encode()` just encodes objects or arrays of objects as a strings or arrays of strings that are equivalent to calling `Parse.Object.toString()` instead of serializing them with `_toFullJSON()`. Bummer.

```js
const ParseFromShim = require('parse-shim');
const ParseLegit = require('parse');

class Obj extends ParseFromShim.Object { ... };
let obj = new Obj();

obj instanceOf ParseLegit.Object // returns -> false ... obviously this won't work!
```

I realized this was going on because inside `project/node_modules/parse-server/node_modules/parse` we have this code block in encode.js:encode()

```js
if (value instanceof _ParseObject2['default']) { // <- fudge, not same PO as parse-shim's
  // code to encode the object as required
}
```

Do you think it would be possible to simply have the parse-shim functionality be part of the Parse-JS-SDK proper so that parse-server's required module and our codebase's required module are considered equivalent? Any other thoughts on a good solution here?

I was also thinking we could check if something is a Parse.Object another way instead of comparing prototypes with `instanceof` but that may just be a solution for me and not something that would work for everyone in the long term.

TylerBrock:
This would resolve #269 and allow us to remove the `parse-shim` module from our codebase.

I assume we are not alone in using the JS SDK on all three platforms (Browser, Node, React-Native) and it has become harder to make it work seamlessly on all three as packages like parse-server and parse-cloud-express modify the exported parse module (overwriting Parse.Cloud members for example). So we'd like to make it official instead of using a shim.

Additionally, it's also more Parse-like to 'just work' as a JS sdk instead of having another set of instructions: "require x when you are on platform y, but z when you are on platform whatever"

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,7 @@
 // Export the correct version of the SDK for each platform that requires 'parse'
 if (typeof window === 'undefined') {
   module.exports = require('./lib/node/Parse.js');
-} else if (navigator && navigator.product === 'ReactNative') {
+} else if (typeof navigator === 'object' && navigator.product === 'ReactNative') {
   module.exports = require('./lib/react-native/Parse.js');
 } else {
   module.exports = require('./lib/browser/Parse.js');
--- a/index.js
+++ b/index.js
@@ -1,1 +1,8 @@
-module.exports = require('./lib/browser/Parse.js');
+// Export the correct version of the SDK for each platform that requires 'parse'
+if (typeof window === 'undefined') {
+  module.exports = require('./lib/node/Parse.js');
+} else if (navigator && navigator.product === 'ReactNative') {
+  module.exports = require('./lib/react-native/Parse.js');
+} else {
+  module.exports = require('./lib/browser/Parse.js');
+}


andrewimm:
I appreciate the reasons for wanting this, and I want to pursue a workaround with you. That said, this just doesn't work with the build pipelines people use, and requires packaging 3 versions of the same code in every SDK download.

Prior to 1.6, we shipped node and web in the same SDK, and as features diverged, it became a real mess to maintain. Working with the React Native team before their initial release, we found that adding yet a third platform into this mix was an even bigger hassle. We wrote some god-awful code that ran `eval()` just to properly target the differences between node, browserify, webpack, the RN packager... And that is why when 1.6 came around, we intentionally split the node and RN versions into separate packages. Runtime checks aren't reliable; explicitly-built packages are.

The first issue is that you can't count on every developer to be running dead-code elimination. If they're not, they will be shipping three versions of nearly the same code for no reason, and that's not an experience we want to give. The second issue is that it's nearly impossible to even reach that step with the build pipelines people typically use. Systems tend to run dead code elimination _after_ building the dependency tree, and dependency fetching isn't privy to runtime state. In the code attached to this PR, webpack or the RN packager would follow all three code trees, as well as the modules within them. You'd see webpack complaining about RN's AsyncStorage dependencies, and you'd probably be pulling in a dysfunctional `xmlhttprequest` module for RN. The only way to support this scenario is to "outsmart" the dependency fetchers and attempt to stay one-step ahead of them as they change. Here are some examples from v1.5.0:

```js
// Rename require so that browserify / RN don't follow it
var req = typeof(require) === 'function' ? require : null;
// Load references to other dependencies
if (typeof(XMLHttpRequest) !== 'undefined') {
  Parse.XMLHttpRequest = XMLHttpRequest;
} else if (typeof(require) === 'function' &&
    typeof(require.ensure) === 'undefined') {
  Parse.XMLHttpRequest = req('xmlhttprequest').XMLHttpRequest;
}
```

```js
if (typeof require === 'function') {
  var AsyncStorage;
  try {
    // RN packager uses regex to find require, so it catches this,
    // Webpack and browserify don't.
    // On node, it'll outright fail, hence the try/catch statement
    AsyncStorage = eval("require('AsyncStorage')");
    // ...
```

These had to be run in conjunction with all sorts of per-packager overrides to ignore specific modules. It was a mess, and created more issues than it solved.

It seems to me like you have two problems: you want to use the same require statement in all code, so you can use the same files on three platforms, and you want to have your setup work well with `parse-server` and the cloud experience it provides. I think a shim that lives outside of the SDK is your best approach on the first, but we should have a discussion about the second. As far as this PR goes, I tried very hard to make things "just work," but I've been very far down this road and it doesn't pan out.

(FWIW, the use cases you're describing are the reasons why I'm building that lower-level SDK I mentioned to you previously. It's gotten sidetracked since this really isn't my job anymore, but it solves a lot of cross-platform issues.)

TylerBrock:
Ok, I hear you, thanks for responding. This was mostly to test the waters.

I can work around the `parse-server` problem by just doing the same thing it does to the module exported by `parse` to the module exported by our shim but it's not clean.

Maybe the problem is that this tight coupling of cloud code + parse-server is what needs to change. It would be far better to run parse-server externally (isolated) and register the triggers/cloud code as webhooks. I can bring that discussion up there and help implement it.

Issue #271, opened 2016-04-30
SSL and LiveQuery?


cheesykyle:
I've secured my app with SSL and everything is working as expected, aside from LiveQueries. I'm accessing my parse-server from an SSL webpage as well and receive the following error from the web console:

`https://example.com/ was loaded over HTTPS, but attempted to connect to the insecure WebSocket endpoint 'ws://example.com/parse'. This request has been blocked; this endpoint must be available over WSS`

So I assume this is a simple matter of LiveQueries not creating a secure WebSocket connection and providing the correct protocol. Is this something that is being worked on, or how can I ensure I have access to the live query protocol behind ssl.

Thanks!

cheesykyle:
I've fixed this client-side for the meantime by replacing with the following at line 3653 in parse-latest.js
```javascript
if (!liveQueryServerURL) {
        var tempServerURL = _CoreManager2['default'].get('SERVER_URL');
        var protocol = 'ws://';
        if (tempServerURL.indexOf('https') === 0) {
          protocol = 'wss://'
        }
        var host = tempServerURL.replace(/^https?:\/\//, '');
        liveQueryServerURL = protocol + host;
        _CoreManager2['default'].set('LIVEQUERY_SERVER_URL', liveQueryServerURL);
      }
```

andrewimm:
Want to submit a PR for this?

cheesykyle:
PR Submitted!

PR #277, opened 2016-05-12
Fix LiveQuery SSL

cheesykyle:
This allows LiveQueryServer to work over Parse Sever with SSL

### DIFF
--- a/src/ParseLiveQuery.js
+++ b/src/ParseLiveQuery.js
@@ -104,13 +104,13 @@ let DefaultLiveQueryController = {
       
       // If we can not find Parse.liveQueryServerURL, we try to extract it from Parse.serverURL
       if (!liveQueryServerURL) {
-        var tempServerURL = CoreManager.get('SERVER_URL');
-        var protocol = 'ws://';
+        const tempServerURL = CoreManager.get('SERVER_URL');
+        let protocol = 'ws://';
         // If Parse is being served over SSL/HTTPS, ensure LiveQuery Server uses 'wss://' prefix
         if (tempServerURL.indexOf('https') === 0) {
           protocol = 'wss://'
         }
-        var host = tempServerURL.replace(/^https?:\/\//, '');
+        const host = tempServerURL.replace(/^https?:\/\//, '');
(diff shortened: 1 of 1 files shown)


cheesykyle:
This is the fix described in #271

src/ParseLiveQuery.js
```diff
@@ -101,11 +101,17 @@ let DefaultLiveQueryController = {
       if (liveQueryServerURL && liveQueryServerURL.indexOf('ws') !== 0) {
         throw new Error('You need to set a proper Parse LiveQuery server url before using LiveQueryClient');
       }
-
+      
       // If we can not find Parse.liveQueryServerURL, we try to extract it from Parse.serverURL
       if (!liveQueryServerURL) {
-        let host = CoreManager.get('SERVER_URL').replace(/^https?:\/\//, '');
-        liveQueryServerURL = 'ws://' + host;
+        var tempServerURL = CoreManager.get('SERVER_URL');
```
andrewimm: `const`

src/ParseLiveQuery.js
```diff
@@ -101,11 +101,17 @@ let DefaultLiveQueryController = {
...
+        var protocol = 'ws://';
```
andrewimm: `let`

src/ParseLiveQuery.js
```diff
@@ -101,11 +101,17 @@ let DefaultLiveQueryController = {
...
+        // If Parse is being served over SSL/HTTPS, ensure LiveQuery Server uses 'wss://' prefix
+        if (tempServerURL.indexOf('https') === 0) {
+          protocol = 'wss://'
+        }
+        var host = tempServerURL.replace(/^https?:\/\//, '');
```
andrewimm: `const`

andrewimm:
Just a few style concerns: we're trying to use `let`/`const` on all new code.

cheesykyle:
'var' has updated with 'let' and 'const'

andrewimm:
Awesome. I'll see if I can cut a 1.8.5 branch to include this.

PR #285, opened 2016-05-25
Update documentation for Each queries.

adamserafini:
Hi, it appears Each queries can also accept useMasterKey and sessionToken option but it wasn't documented, this PR is just a minor docs fix.


andrewimm:
Thanks, looks good!

PR #289, opened 2016-05-29
Adds ability to interpolate delete operations

flovilmart:
On parse.com, when running a beforeSave hook, if a key is `unset` the change is not propagated to the client. 

On parse-server, we forward this change by setting the `key: {__op: Delete}` that is interpolated by the JS SDK as an UnsetOp.

This proposed PR will set key = undefined in those cases fixing https://github.com/ParsePlatform/parse-server/issues/1840

https://github.com/ParsePlatform/parse-server/pull/1946

### DIFF
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -375,6 +375,9 @@ export default class ParseObject {
         changes[attr] = new ParseACL(response[attr]);
       } else if (attr !== 'objectId') {
         changes[attr] = decode(response[attr]);
+        if (changes[attr] instanceof UnsetOp) {
+          changes[attr] = undefined;
+        }
       }
     }
     if (changes.createdAt && !changes.updatedAt) {
--- a/src/__tests__/ParseObject-test.js
+++ b/src/__tests__/ParseObject-test.js
@@ -852,6 +852,25 @@ describe('ParseObject', () => {
       done();
     });
   }));
+  
+  it('interpolates delete operations', asyncHelper((done) => {
+    CoreManager.getRESTController()._setXHR(
+      mockXHR([{
+        status: 200,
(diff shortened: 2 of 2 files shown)


TylerBrock:
Greatness

flovilmart:
We could have used null (it works right of the bat) but it's inconsistent with the undefined retuned when key is not set. Also, may cause problems with the iOS SDK as it interpolates to NSNull * and we usually don't love those.

andrewimm:
Shouldn't break things. For compatibility reasons, I might want to bump the version to 1.9.0 for this.

andrewimm:
Also I can't express how happy I am that we're resolving this ridiculous piece of tech debt that has plagued Cloud Code for years 💯

flovilmart:
That should feel good to say it out loud :+1:

flovilmart:
You can bump to 1.9.0, but it's unlikely that will break backwards.

andrewimm:
Agree it shouldn't break backwards, but I think it'd be easier to say "this functionality is fixed in SDK versions 1.9.0 and up"

flovilmart:
Yup!

TylerBrock:
Is there any timing on 1.9.0?

andrewimm:
Sorry @TylerBrock I forgot we were blocking this. I can ship it tomorrow.

TylerBrock:
no rush, i was just curious

andrewimm:
I was already intending to ship it this evening, actually, but then I got caught up in investigating the best CDN solution.

andrewimm:
Just released 1.9.0-rc1 as a beta release.

andrewimm:
@TylerBrock turns out I totally f-d up that publish (late night, little sleep). I'm trying to unpublish rc1, but it should all be there under `parse@1.9.0-rc2`, aka `parse@rc`

PR #294, opened 2016-06-16
Update ParseFile.js

benjaminmcdonald:
Fixed bug when base64 string is too long

RangeError: Maximum call stack size exceeded
    at RegExp.exec (native)
    at new e (http://localhost:2052/scripts/parse-1.6.14.min.js:13:25178)

### DIFF
--- a/src/ParseFile.js
+++ b/src/ParseFile.js
@@ -24,7 +24,7 @@ export type FileSource = {
 };
 
 var dataUriRegexp =
-  /^data:([a-zA-Z]*\/[a-zA-Z+.-]*);(charset=[a-zA-Z0-9\-\/\s]*,)?base64,(\S+)/;
+  /^data:([a-zA-Z]*\/[a-zA-Z+.-]*);(charset=[a-zA-Z0-9\-\/\s]*,)?base64,/;
 
 function b64Digit(number: number): string {
   if (number < 26) {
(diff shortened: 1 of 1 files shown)


facebook-github-bot-1:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

facebook-github-bot-0:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

andrewimm:
Thanks for the PR.

I'm still reviewing to make sure this won't create any backwards incompatibility. My immediate concern is the use of `window.atob()` to check whether the string is valid. The Parse SDK ecosystem contains more than just the browser, and neither `window` nor `atob` are globally available in Node.js or React Native. The Travis CI tests only passed because Jest uses `jsdom` to polyfill a browser-like `window` global.
Secondly, the slice from 0 to 100 seems pretty arbitrary. If we're no long using `dataUriRegexp` to validate the latter part of the string, we don't need to keep it as is -- just slice from 0 up until `commaIndex`, and then use a more compact regexp to try and extract the mimetype.

Additionally, do you mind doing two things?
1. I know there's no official "style guide," but please do your best to match the existing style of the code around your changes. This means spaces, not tabs, and leaving whitespace around ternary operators.
2. Unit tests. Since this adds new codepaths, we need more coverage. I'm happy to do this if it's too much trouble, but I worry about shipping any changes that aren't well-covered.

benjaminmcdonald:
Well if atob can't be used because it's not supported across platforms and a regular expression will result in a "Maximum call stack size exceeded" with long strings then we should drop the check to validate the base64 string.

And a regex expression only be used to find the data type

andrewimm:
Agree, it's probably not too important to validate the base64 string on the client. If the server fails, it'll result in a failure on the client eventually.

benjaminmcdonald:
Update the patch. Tabs converted to spaces, removed base64 string validation

andrewimm:
Thanks for this!

PR #313, opened 2016-07-15
closed #311: Parse.File sends incorrect Content-Type to API

### LINKED ISSUE #311: Parse.File sends incorrect Content-Type to API
Hi,

first the issue:
a) I want to upload a PDF file using Parse.File:
```
// assume you have a File-Object in variable "file" with a PDF file
var file_obj  = new Parse.File('test.pdf', file, 'application/pdf');
```

b) Everything is fine while saving, S3 stores the file
c) But then I get "Failed to load PDF document" from S3.

Reason: I get this error message because S3 sends me `Content-Type: text/plain` for my PDF file

- - -

So back to step b) &mdash; I can see the following happening:
Parse sends a `POST` request to the API to save the file with the following headers:
```
X-Parse-Application-ID: ...
X-Parse-JavaScript-Key: ...
Content-Type: text/plain
```

Q: Why is it sending `"text/plain"`, although I've told Parse.File that I want to upload something with mime type `"application/pdf"`?
A: Someone wanted to avoid pre-flight requests and hard-coded the `Content-Type: text/plain` in the REST Controller (see [this](https://github.com/ParsePlatform/Parse-SDK-JS/blob/9a521e948a966188361af9d9bf544e3da47f01cc/src/RESTController.js#L128)).

- - -

**Solution 1**

After deleting this line: `headers['Content-Type'] = 'text/plain'; // Avoid pre-flight`
there is no problem with the Content-Type and I get my PDF file with the correct Content-Type.

Is there any reason to hard-code the Content-Type there? The only downside is that browsers will send `OPTIONS` requests to make sure that they can do CORS.

**Solution 2**

If there are other disadvantages I didn't consider in Solution 1, is it safe to use something like this? i.e.

```
// replacing line #128: https://github.com/ParsePlatform/Parse-SDK-JS/blob/9a521e948a966188361af9d9bf544e3da47f01cc/src/RESTController.js#L128
if (!(data instanceof File)) {
    headers['Content-Type'] = 'text/plain'; // Avoid pre-flight
}
```

**Solution 3**

All those suggestions above would ignore the mime type that gets passed as a third argument for Parse.File e.g. if `file` is a File-Object with `file.type == 'application/pdf'`, then `new Parse.File('filename.png', file, 'image/png')` would still send Content-Type `"application/pdf"`. It would be interesting to hear if there are some real world use cases that need to set another mime type.  

If there are use cases for this, `saveFile` (see [this](https://github.com/ParsePlatform/Parse-SDK-JS/blob/9a521e948a966188361af9d9bf544e3da47f01cc/src/ParseFile.js#L235)) should be changed and combined with Solution 2. This would result in the following change in `saveFile`:

    var headers = {
      'X-Parse-Application-ID': CoreManager.get('APPLICATION_ID'),
      'X-Parse-JavaScript-Key': CoreManager.get('JAVASCRIPT_KEY'),
      'Content-Type': source.type || (source.file? source.file.type : null)
    };

This would have the desired behaviour:
- `Parse.File('test.pdf', file, 'image/png')` will send `Content-Type: image/png` (Third argument overrides `file` mime type)
- `Parse.File('test.pdf', file)` will send e.g. `Content-Type: application/pdf` if `file` is a PDF file (or `file.type` so for a png file it would send `"image/png"`)

- - -

I would like to push **Solution 3** into the repository, do you agree?  
Please comment on this issue or - if you experienced similar problems - say if Solution 3 helped you.

samuelantonioli:
See issue #311 for details and discussion.
Maybe test coverage should be improved for this functionality.

### DIFF
--- a/src/ParseFile.js
+++ b/src/ParseFile.js
@@ -245,7 +245,8 @@ var DefaultController = {
     // To directly upload a File, we use a REST-style AJAX request
     var headers = {
       'X-Parse-Application-ID': CoreManager.get('APPLICATION_ID'),
-      'X-Parse-JavaScript-Key': CoreManager.get('JAVASCRIPT_KEY')
+      'X-Parse-JavaScript-Key': CoreManager.get('JAVASCRIPT_KEY'),
+      'Content-Type': source.type || (source.file? source.file.type : null)
     };
     var url = CoreManager.get('SERVER_URL');
     if (url[url.length - 1] !== '/') {
--- a/src/RESTController.js
+++ b/src/RESTController.js
@@ -125,7 +125,9 @@ const RESTController = {
       };
 
       headers = headers || {};
-      headers['Content-Type'] = 'text/plain'; // Avoid pre-flight
+      if (typeof(headers['Content-Type']) !== 'string') {
+        headers['Content-Type'] = 'text/plain'; // Avoid pre-flight
+      }
       if (CoreManager.get('IS_NODE')) {
         headers['User-Agent'] = 'Parse/' + CoreManager.get('VERSION') +
           ' (NodeJS ' + process.versions.node + ')';


andrewimm:
I'll find some time to cover this with tests

PR #350, opened 2016-09-22
Add isFetched method to Parse.Object

dpoetzsch:
Hi,

often, I want to assert that in a specific method a parse object has been fetched. Also, sometimes I only want to fetch an object if it hasn't been fetched before. Others seem to have [similar requests](https://parse.com/questions/determine-if-parseobject-has-been-fetched).

So far I've been using the proposed work-around: `!!parseObject.createdAt`. This works fine, however, makes the code harder to read and if the behavior of `createdAt` changes at some point code might get broken.

To fix this for all of us, I created this PR. It adds a `isFetched` property to ParseObject which currently is calculated exactly like the work-around (I'm not an expert on parse object internals).

I'm happy for feedback.

### DIFF
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -168,6 +168,10 @@ export default class ParseObject {
     return this._getServerData().updatedAt;
   }
 
+  get isFetched(): boolean {
+    return !!this.createdAt;
+  }
+
   /** Private methods **/
 
   /**
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -168,6 +168,11 @@ export default class ParseObject {
     return this._getServerData().updatedAt;
   }
 
+  /**
+   * Returns true if the object was fetched previously.
+   * @property isFetched
+   * @type boolean
+   */
   get isFetched(): boolean {
     return !!this.createdAt;
   }
--- a/src/__tests__/ParseObject-test.js
+++ b/src/__tests__/ParseObject-test.js
@@ -2101,4 +2101,29 @@ describe('ParseObject extensions', () => {
     var i = new InitObject()
     expect(i.get('field')).toBe(12);
   });
+
+  it('returns correct isFetched state', asyncHelper((done) => {
+    CoreManager.getRESTController()._setXHR(
+      mockXHR([{
+        status: 200,
+        response: {
+          objectId: 'P5'
+        }
+      }, {
+        status: 200,
+        response: {
+          createdAt: '2013-12-14T04:51:19Z'
+        }
+      }])
+    );
+
+    return new ParseObject('Test').save().then((obj) => {
+      var p = ParseObject.createWithoutData(obj.id);
+      expect(p.isFetched).toBe(false);
+      return p.fetch();
+    }).then((o) => {
+      expect(o.isFetched).toBe(true);
+      done();
+    });
+  }));
 });


andrewimm:
Unfortunately, checking the `createdAt` field isn't necessarily an indicator of whether an object has been fetched. If you do a `select` clause in your query, it will only retrieve some of the fields, and may either exclude the `createdAt` field on an object with all other data provided, or may include it on a partially-fetched object. In general, I'm wary of adding the idea of a "fetched" object, since the name isn't really truthful. If I query an object, and another client changes it, my copy is out of date and whether it is fetched or not is probably going to create incorrect results.

I realize this is an issue, but there is no one-size-fits-all solution for it. In your personal code, I'd recommend two strategies:

1. If you need to verify that a specific field has been fetched before using an object, check the value of that field. `.has()` is actually designed for this.
2. If you need to guarantee that you have the latest copy of an object, always re-retrieve it from the server before using it. It may have changed.

Hope this makes sense! Happy to discuss further, if your use cases are somewhat different from what I've described.

dpoetzsch:
Thanks for the answer. I unterstand the problem that arises when dealing with `select` clauses.

I checked my usages of the `createdAt` hack (which still works fine for the mongo db adapter as far as I get it) and yes it would be possible to use the `.has()` method for those, although it makes the code much more unreadable because I'd expect all fields to be set. I think, for me I'd rather define a `fetched()` method in the subclasses of ParseObject to make the code more readable.

Thanks for the clarification why a general approach is unfeasible, this PR can be closed then.

andrewimm:
I would like to pursue this further, if you can think of more reliable ways to implement it. I think it's definitely valuable to know if an object is empty because it came through a shallow pointer, or whether it's been fully fetched and has no fields.
I'm just wary of adding new APIs that can do nearly the same thing as existing ones, and want to make sure that anything we add isn't misleading to users. If we provide an `isFetched()` method, I'd want to make sure it's really telling the truth. It might be possible to do so by inspecting the the underlying Store entry for the object's data, but I'd have to dig into the code to see if an unfetched pointer behaves any differently.

dpoetzsch:
Interesting, you made one point of this method very clear that I wasn't able to formulate clearly so far: `isFetched()` should be `true` after I call `fetch()` on an object that has all fields set to `null` in the database. This was the original point of using `createdAt`, assuming that even if nothing else has been set, `createdAt` will be set.

To expand on that idea and differentiate it from the `has()` method, maybe an `isFetched(fieldName)` makes sense: It returns `true`, if that field was part of a prior fetch (i.e. either the complete object was fetched or the field was part of a select clause). In particular, `isFetched(fieldName)` also returns true if that field has a `null` or `undefined` value in the database.

Spinning this thought a bit further, `isFetched()` without parameters could mean that an object has been full fetched, i.e. it is true for all fields of the table. In many cases I want to assert just that, or at least for almost all fields. However, I am not sure if the `ParseObject` would be able to know that..

More pragmatically, a method that checks if any field was fetched might be easier to implement and in many practical cases suffice (on mongo db where I fetch all or nothing I just want to know if I called fetch at any point on that object). Basically this corresponds to the negation of a (imho pretty understandable) `isPointer()` method.

flovilmart:
@dpoetzsch do you want to pursue that feature further?

dpoetzsch:
Actually I would like to at some point but at the moment I'm too busy to pursue this. Maybe better to close this PR and open a new one once I manage to look into that.

flovilmart:
@dpoetzsch closing this one as suggested, feel free.

PR #358, opened 2016-10-12
Add support to secured endpoints throught Authorization header

sebasjm:
This is a change I made to allow communication with a server that require authorization.


andrewimm:
Thanks for the PR.

I have to requests: first, `credentials` isn't clear what it refers to. I'd want to use something along the lines of "bearer token" or "server auth token."

It will probably be decided by the second question I have: what is the best way to include http basicauth as well? We want to be able to specify both an auth type and an auth token.

Taking all of that into account, I'd ask that this be extended into two fields: `serverAuthType` and `serverAuthToken` (in CoreManager, `SERVER_AUTH_TYPE` and `SERVER_AUTH_TOKEN`).

Then, change the logic in the REST controller to

```
if (CoreManager.get('SERVER_AUTH_TYPE') && CoreManager.get('SERVER_AUTH_TOKEN') {
  headers['Authorization'] = CoreManager.get('SERVER_AUTH_TYPE') + ' ' + CoreManager.get('SERVER_AUTH_TOKEN');
}
```

PR #368, opened 2016-10-27
Allow validation of objects without Object Prototype

haydenbleasel:
https://github.com/ParsePlatform/parse-server/issues/2949

### DIFF
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -880,7 +880,7 @@ export default class ParseObject {
    * @see Parse.Object#set
    */
   validate(attrs: AttributeMap): ParseError | boolean {
-    if (attrs.hasOwnProperty('ACL') && !(attrs.ACL instanceof ParseACL)) {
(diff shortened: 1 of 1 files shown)


andrewimm:
I appreciate the PR. 
If something in your code is overwriting your prototype, this isn't going to be the last time you hit problems in your client environment. You probably have a library that is modifying the core JS objects, and that's likely to cause other unexpected issues again.

I will merge when I can get Travis working again.

callmephilip:
Any update on this? I am seeing this issue as well when using Parse with graphql input objects in resolvers

haydenbleasel:
@callmephilip Since the cause of the error is the prototype being stripped out, you could try creating an new object and copying the properties:

```
const newObject = Object.assign({}, oldObject);
```

Alternatively, I've heard a quick, hacky workaround is:

```
const newObject = JSON.parse(JSON.stringify(oldObject))
```

PR #373, opened 2016-11-05
Add support for providing file upload progress.

aaron-blondeau-dose:
It would be very convenient if developers could get progress when uploading a file with Parse.File.  The use case would look something like this:

```
var parseFile = new Parse.File(file.name, file);
parseFile.save({
    progress: (value) => {
        console.log("TODO show progress to user : "+value);
    }
}).then();
```

See also https://www.parse.com/questions/file-upload-progress-javascript-api


doender:
Will this be merged?

tommy-b-10:
Just wondering when this will be merged?

jcguarinpenaranda:
Please merge this!!

flovilmart:
@aaron-blondeau-dose Thanks for the PR, could you fix the conflicts so we can get that one in for the next release?

flovilmart:
@aaron-blondeau-dose ping?

flovilmart:
@aaron-blondeau-dose I'm doing some cleanup here, it seems that it diverged a lot since your PR, are you still interested at getting this one in?

dplewis:
@flovilmart I can help out with this

dplewis:
@flovilmart Quick update on this.

This will only work in the browser. I haven't tried it on IE8 / IE9.

The xmlhttprequest node package seems to be missing the progress event. I would do a PR to add support for it but it seems like it isn't well maintained.

https://github.com/driverdan/node-XMLHttpRequest

What do you think?

src/ParseFile.js
```diff
@@ -157,9 +158,14 @@ class ParseFile {
   /**
    * Saves the file to the Parse cloud.
    * @param {Object} options
+   *  * Valid options are:<ul>
+   *   <li>useMasterKey: In Cloud Code and Node only, causes the Master Key to
+   *     be used for this request.
+   *   <li>progress: Callback for upload progress
```
acinader: should add a note here that progress is only supported in the browser

src/__tests__/ParseFile-test.js
```diff
@@ -6,17 +6,20 @@
  * LICENSE file in the root directory of this source tree. An additional grant
  * of patent rights can be found in the PATENTS file in the same directory.
  */
-
+/* global File */
```
acinader: what's this?

src/__tests__/ParseFile-test.js
```diff
@@ -183,6 +186,22 @@ describe('ParseFile', () => {
     expect(a.equals(b)).toBe(false);
     expect(b.equals(a)).toBe(false);
   });
+
+  it('reports progress during save when source is a File', () => {
+    const file = new ParseFile('progress.txt', new File(["Parse"], "progress.txt"));
```
dplewis: @acinader /* global File */ is for the `new File(...)` so that it can find it.

kartheeknagasuri:
what's the correct usage of progress call back? I am using like this and not getting any progress call back

var parseFile = new Parse.File(file.name, file);
parseFile.save({
    progress: (value) => {
        console.log("TODO show progress to user : "+value);
    }
}).then();

dplewis:
@kartheeknagasuri That is correct. This feature is for the browser only.

kartheeknagasuri:
@dplewis I am trying through the browser, but not getting the call back, not sure what's the problem

ladigitale:
The progress calback seems to be ran just one time with value 1.
I'm trying the upload of files (size about 1Mo) with a slow network in order to follow the evolution

aliasad106:
Doesn't work, Any progress on this ?

mtrezza:
If there is an issue with this feature, please open a new issue (if it doesn't exist already) to report a bug.

PR #374, opened 2016-11-05
Fix async tests in ParseFile-test.js

aaron-blondeau-dose:
Not sure how widespread of a problem this is, but I recently submitted a PR for ParseFile and noticed that some of the tests for it were not handling async properly and weren't actually executing assertions.  This PR addresses ParseFile's tests only.


flovilmart:
Thanks for catching that one!

PR #386, opened 2016-12-19
ParseFile & useMasterKey: pass options through to RESTController from saveBase64

rsoffer:
Without this, we are getting unauthorized messages when saving base64 files.


Botz:
+1

flovilmart:
LGTM! Thanks!

PR #388, opened 2016-12-22
merge in definately typed TS definitions to parse project

chrismatheson:
Taking the work allready done [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/parse/index.d.ts) and merge it into the project as per sugestions of [typescript project](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html)

This will mean users of typescript & parse will have one less step to do (manually pull in types) and
hopefully changes in the definitions themselves will stay closer to the actual thing.

tests for this have been added as a pretest npm script (if it compiles with no warnings/errors then "tests" are passed)

### DIFF
--- a/integration/types/index.ts
+++ b/integration/types/index.ts
@@ -1,0 +1,415 @@
+/// <reference path="../../index.d.ts" />
+
+function test_events() {
+
+    var object = new Parse.Events();
+    object.on("alert", (eventName: string) => alert("Triggered " + eventName));
+
+    object.trigger("alert", "an event");
+
+    var onChange = () => console.log('whatever');
+    var context: any;
+
+    object.off("change", onChange);
+    object.off("change");
+    object.off(null, onChange);
+    object.off(null, null, context);
+    object.off();
+}
+
+class GameScore extends Parse.Object {
+
+    constructor(options?: any) {
+
+        super("GameScore", options);
+    }
+}
+
+class Game extends Parse.Object {
+
+    constructor(options?: any) {
+
+        super("GameScore", options);
+    }
+}
+
+function test_object() {
+
+    var game = new Game();
+
+// Create a new instance of that class.
+    var gameScore = new GameScore();
+
+    gameScore.set("score", 1337);
+    gameScore.set("playerName", "Sean Plott");
+    gameScore.set("cheatMode", false);
+
+
+    var score = gameScore.get("score");
+    var playerName = gameScore.get("playerName");
+    var cheatMode = gameScore.get("cheatMode");
+
+    gameScore.increment("score");
+    gameScore.addUnique("skills", "flying");
+    gameScore.addUnique("skills", "kungfu");
+
+    game.set("gameScore", gameScore);
+}
+
+function test_query() {
+
+    var gameScore = new GameScore();
+
+    var query = new Parse.Query(GameScore);
+    query.equalTo("playerName", "Dan Stemkoski");
+    query.notEqualTo("playerName", "Michael Yabuti");
+    query.greaterThan("playerAge", 18);
+    query.limit(10);
+    query.skip(10);
+
+    // Sorts the results in ascending order by the score field
+    query.ascending("score");
+
+    // Sorts the results in descending order by the score field
+    query.descending("score");
+
+    // Restricts to wins < 50
+    query.lessThan("wins", 50);
+
+    // Restricts to wins <= 50
+    query.lessThanOrEqualTo("wins", 50);
+
+    // Restricts to wins > 50
+    query.greaterThan("wins", 50);
+
+    // Restricts to wins >= 50
+    query.greaterThanOrEqualTo("wins", 50);
+
+    // Finds scores from any of Jonathan, Dario, or Shawn
+    query.containedIn("playerName",
+                      ["Jonathan Walsh", "Dario Wunsch", "Shawn Simon"]);
+
+    // Finds scores from anyone who is neither Jonathan, Dario, nor Shawn
+    query.notContainedIn("playerName",
+                         ["Jonathan Walsh", "Dario Wunsch", "Shawn Simon"]);
+
+    // Finds objects that have the score set
+    query.exists("score");
+
+    // Finds objects that don't have the score set
+    query.doesNotExist("score");
+    query.matchesKeyInQuery("hometown", "city", query);
+    query.doesNotMatchKeyInQuery("hometown", "city", query);
+    query.select("score", "playerName");
+
+    // Find objects where the array in arrayKey contains 2.
+    query.equalTo("arrayKey", 2);
+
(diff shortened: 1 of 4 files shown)


agonbina:
@chrismatheson any updates on this? Looks like the types are a bit outdated already. For example, they're still referencing types for jquery and underscore which afaik are no longer used in the SDK? If you need any help to finish this maybe I can jump in and spend some time on it 👍

chrismatheson:
@agonbina i had exactly the same thought (that the typedefs were out of date). The i started to think about updating those shared typedefs i went down a hole of "well this should probably be a PR to the project itself as those maintainers would have a better understanding of the correctness" so i figure the smallest change was to pull in the current typedefs to this project 1st, then PR here to bring them up-to-date

however i could see the benefit of doing it all at once, so if thats considered a pre-req of merging this change ill happily attack that & welcome all help :)

flovilmart:
Committing those files is cumberstone and the type definitions should probably distributed separately. I'm no expert in this, but it seems that the burden is on the developer to run those definitions for every PR, and it doesn't work well. We can think of a CI step that would, upon releases rebuild those. What do you think?

flovilmart:
Can you get us setup to publish to the `@types` npm org? This is something manageable for us on CI for each releases!

chrismatheson:
@flovilmart admitidtly I've been quite lax on this one, but I'm still keen for it to be part of this project itself, a couple of reasons ....

As part of the project there is no need to install two things which is nice for those of us using typescript. The definitions add no weight to production code (left out by both browserify/webpack style build & HTML script style builds)

But also, separating the types from the project allows too easily for drift between the "real" world and the type defs, if the types are to be useful then anyone changing the api of the parse my. Should update the typedefs alongside that change (so keep the two thing together :) ) 

As for the "burden", yes this would be on the developer submitting any PR that changes API of the parse lib (happy side note that anything touching the typedefs is probably a major or minor semver change and this make it more visible)

flovilmart:
@chrismatheson What I suggest is that we automate this process but don't put the burden on the developer (which don't make sense to me) nor commit those files which are purely generative. 

What I recommend moving forward is to include it when releasing the package, the generation can occur at `prepublish` at the same time we build the libraries. Then it's just a matter of configuring the package.json to include the definitions.

What do you think?

package.json
```diff
@@ -24,11 +24,16 @@
     "PATENTS",
     "README.md"
   ],
+  "types": "./index.d.ts",
   "browser": {
     "react-native": false
   },
   "dependencies": {
+    "@types/jquery": "^2.0.34",
```
flovilmart: those are dev dependencies.

integration/types/tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+        ],
+        "types": [],
+        "noEmit": true,
+        "forceConsistentCasingInFileNames": true
+    },
+    "files": [
+        "../../index.d.ts",
+        "parse-tests.ts"
+    ]
+}
```
flovilmart: add line break

package.json
```diff
@@ -53,11 +58,13 @@
     "gulp-uglify": "^1.4.0",
     "jasmine-reporters": "~1.0.0",
     "jest-cli": "^15.1.1",
+    "typescript": "^2.1.4",
```
flovilmart: this one is duplicated here.

package.json
```diff
@@ -53,11 +58,13 @@
...
     "vinyl-source-stream": "^1.1.0"
   },
   "scripts": {
     "build": "./build_releases.sh",
     "release": "./build_releases.sh && npm publish",
+    "pretest": "tsc integration/types/index.ts --noEmit",
```
flovilmart: add a step to build `index.d.ts` why would you want tsc to be run on all machines and a requirement for all developers? I don't mind adding it to the CI. What's the benefit of running it as a pre-test?

integration/types/index.ts
```diff
@@ -0,0 +1,415 @@
...
+    // Find objects where the array in arrayKey contains all of the elements 2, 3, and 4.
+    query.containsAll("arrayKey", [2, 3, 4]);
+
+    query.startsWith("name", "Big Daddy's");
+    query.equalTo("score", gameScore);
+    query.exists("score");
+    query.include("score");
+    query.include(["score.team"]);
+
+    var testQuery = Parse.Query.or(query, query);
```
flovilmart: unused variable

integration/types/index.ts
```diff
@@ -0,0 +1,415 @@
...
+}
+
+class TestCollection extends Parse.Collection<Object> {
+
+    constructor(models?: Parse.Object[]) {
+
```
flovilmart: remove blank line

integration/types/index.ts
```diff
@@ -0,0 +1,415 @@
...
+    var bytes = [ 0xBE, 0xEF, 0xCA, 0xFE ];
+    var file = new Parse.File("myfile.txt", bytes);
+
+    var file = new Parse.File("myfile.zzz", {}, "image/png");
+
+    var src = file.url();
+
+    file.save().then(
+    () => {
```
flovilmart: put on previous line

integration/types/index.ts
```diff
@@ -0,0 +1,415 @@
...
+    },
+    (error) => {
+        // The file either could n ot be read, or could not be saved to Parse.
+    });
+
+    Parse.Cloud.httpRequest({ url: file.url() }).then((response: Parse.Promise<Parse.Cloud.HttpResponse>) => {
+        // result
+    });
+
+    // TODO: Check
```
flovilmart: that would be nice to not have TODO's on a brand new clean PR

flovilmart:
@chrismatheson any update on that? As I mentioned, I'm more enclined to handle this at CI time instead of with the repository, as we don't commit build artifacts. This goes also for parse-server for example as the compiled code is not committed for example.

chrismatheson:
@flovilmart Hi! yeah i agree that the best solution would be to generate the definitions at CI / publish time. However I'm not sure thats possible ? Maybe I'm missing something, but this codebase is using flow types? and i haven't seen any way to create a typescript definition from the flowtype system?

flovilmart:
How did you generate the definitions in the 1st place?

chrismatheson:
@flovilmart i didn't, it was hand made by others and was part of the definitely typed project. the idea of merging into this project was to get the description of the types closer to the code (like you mentioned) and actually to have the updating of those same typedefs go hand in hand with updating the code. :) plus its a much nicer dev experience for those who are using types.

flovilmart:
Yeah but if you don't provide a way to keep them up to date, that's quite useless and again, I believe I've been clear enough on how I'd like those to be distributed. Also, the packages are bundled as dependencies, we need them as devDependencies. 

Closing as not fitting the requirements, feel free to open a new PR that addresses the concernes expressed above.

PR #405, opened 2017-01-26
ParseLiveQuery controller should wrap all subscription events

dstarke:
When you call `subscribe()` on a query (which goes through the ParseLiveQuery controller), the resulting subscription does not currently receive `close` or `error` events for the underlying connection. This was occurring because the LiveQueryController is wrapping the underlying subscriptions, and was not forwarding the `close` or `error` events.

This pull request correctly wraps these two event types so that subscriptions receive all the event notifications they are supposed to, and introduces a new test case to check that all subscription events are forwarded correctly.


flovilmart:
LGTM! Thanks for the PR, and sorry for the late review!

PR #406, opened 2017-01-27
Add sensible string representation of ParseError objects

claudijo:
I noticed that our error logs are filled with thousands of "[Object object]" due to the ParseError object not having a sensible string representation. Errors ripple down to the Express.js error log handler, which calls `console.error(err.stack || err.toString())`. In our case, these are typically `209 Invalid session token` errors, as we just migrated to using revocable sessions. The proposed fix in this pull request is to add a `toSting()` method for ParseErrors.

### DIFF
--- a/src/ParseError.js
+++ b/src/ParseError.js
@@ -19,6 +19,10 @@ export default class ParseError {
     this.code = code;
     this.message = message;
   }
+  
+  toString() {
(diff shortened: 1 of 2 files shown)


flovilmart:
Thanks for the PR and sorry for the late response! We'Ll get that in for 1.10!

PR #409, opened 2017-02-11
Select queries should not erase cached data from unselected fields

### LINKED ISSUE #371: after select query object overide
Hi Parse Team,

I am facing problem with parse open source js sdk 1.9.2 .when i apply find query without select.i got full object from parse. after that i make query on same class with select and specified key server return selected object with specified key. but my previous data/object override.
can you any tell me this is open issue with parse open source js sdk.

dstarke:
When a query is issued with a select, we know that the object returned will only be part of the truth from the server. The current implementation, on retrieving the results of a query with a select, erases all the information in the shared object except the fields included in the select, which are updated. That's not the best possible estimate of the state of truth on the server.

Ideally, for a query with a select, the cached object should only be updated or erased for the fields involved in the query.

This PR introduces test cases for retrieving previously retrieved objects with a more restrictive select clause in accordance with the above goal, and proposes a code change that fixes the issue and allows these new test cases to pass.

Fixes #371.

### DIFF
--- a/src/ParseQuery.js
+++ b/src/ParseQuery.js
@@ -45,6 +45,70 @@ function quote(s: string) {
 }
 
 /**
+ * Handles pre-populating the result data of a query with select fields,
+ * making sure that the data object contains keys for all objects that have
+ * been requested with a select, so that our cached state updates correctly.
+ */
+function handleSelectResult(data: any, select: Array<string>){
+  var serverDataMask = {};
+
+  select.forEach((field) => {
+    let hasSubObjectSelect = field.indexOf(".") !== -1;
+    if (!hasSubObjectSelect && !data.hasOwnProperty(field)){
+      // this field was selected, but is missing from the retrieved data
+      data[field] = undefined
+    } else if(hasSubObjectSelect){
+      // this field references a sub-object,
+      // so we need to walk down the path components 
+      let pathComponents = field.split(".");
+      var obj = data;
+      var serverMask = serverDataMask;
+
+      pathComponents.forEach((component, index, arr) => {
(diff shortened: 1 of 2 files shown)


jeveloper:
Just curious, is anyone going to merge any of those pending pull requests ?

flovilmart:
@hramos can you get me access here?

flovilmart:
@dstarke what a beautifully tested PR. I'll jump into it later today

flovilmart:
@dstarke sorry, I've been under the water a bit lately. Just a small recommendation on the code formatting for that PR (spaces mostly). Can you address those?

flovilmart:
Thanks for the fixes! This reminds me I need to setup a linter on that repo as well.

flovilmart:
@dstarke it seems that this PR has introduced an issue, can you have a look shortly please?

dstarke:
The tests for ParseQuery do specifically test for that condition, and they pass (I just checked this as well), so I think the problem may be somewhere else. The tests for ParseQuery, though, directly supply the objects returned by queries, since they come from another unit.

I'm trying to figure out how to reproduce it.

flovilmart:
I didn’t managed to repro either, but I believe @dplewis had a hunch

dstarke:
I did find some problems with how sub-objects were handled. Added tests and fixes for those problems in the PR I just created. I still can't quite reproduce the issue as described in #475, but this should fix at least one of the issues that is discussed in: parse-community/parse-server#3999.

flovilmart:
@dstarke thanks for the quick response!

PR #422, opened 2017-04-05
update path of react-native module

markdowney:
As of react-native 0.43, the main entry point of the package has been renamed (see: https://github.com/facebook/react-native/commit/cf1bc8d6443432ba24efa17cd15ba73072fad193).

This allows the library to work with RN-0.43 but will break on prior versions.

### DIFF
--- a/src/StorageController.react-native.js
+++ b/src/StorageController.react-native.js
@@ -11,7 +11,7 @@
 
 import ParsePromise from './ParsePromise';
 // RN packager nonsense
-import { AsyncStorage } from 'react-native/Libraries/react-native/react-native.js';
+import { AsyncStorage } from 'react-native/Libraries/react-native/react-native-implementation.js';
 
 var StorageController = {
   async: 1,


thiagoghisi:
@markdowney I would love to see this change in the next parse version. Any idea what is going on? It looks like the broken build was removed from travis-ci.

markdowney:
Nope, no idea. But meanwhile, you can use the module from here: 
```
  "dependencies": {
    ...
    "parse": "git+https://github.com/bakery/Parse-SDK-JS.git#release-1.10",
    ... 
  }
```

flovilmart:
@markdowney can we get a proper fix for both versions using require?

markdowney:
Not sure how you'd fixed this. I tried catching the error on require to fallback on the old version, as in:

```
let rnModule;
try {
    rnModule = require('react-native/Libraries/react-native/react-native-implementation.js');
} catch (err) {
    rnModule = require('react-native/Libraries/react-native/react-native.js');
}
```

but this won't work for react-native due to https://github.com/facebook/react-native/issues/5079

flovilmart:
Not sure why they want to do it that way, but it's quite annoying as it's a transitive dependency, not a hard one. Does the try/catch solution work for you?

flovilmart:
React-native is not a dependency of the JS SDK, I believe that's why it returns an empty module

markdowney:
But adding it as a peerDependency should do the trick, shouldn't it?

flovilmart:
Perhaps :) worth having a look, problem is that it will probably force the version to the one in the peers dependencies.

flovilmart:
Closing as already addressed will land in 1.10.

PR #437, opened 2017-05-26
New query condition to match all strings that starts with some other given strings

eduardbosch:
Hi to all! 🎉 

This is the same contribution as the link below, but for Parse-SDK-JS
## https://github.com/parse-community/Parse-SDK-Android/pull/673

😃 

PD: @flovilmart  is ti ok to revert the _dist & lib_ folders commit or should I delete this commits?


flovilmart:
waiting on parse-community/parse-server#3864 to land before we merge.

eduardbosch:
Great! 🎉

PR #438, opened 2017-05-31
Added WithinPolygon to Query

dplewis:
I added support for querying within polygon on the latest server update

https://github.com/parse-community/parse-server/pull/3866

Let me know if I should add more tests.

Sometimes errors when I run test I get a lot of these errors
```
Error: timeout of 5000ms exceeded. Ensure the done() callback is being called in this test.
```
I don't think my objects are getting delete before and after each it() test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ lib
 logs
 node_modules
 test_output
+integration/test_logs
 *~
 .DS_Store
 .idea/
--- a/integration/package.json
+++ b/integration/package.json
@@ -3,7 +3,7 @@
   "dependencies": {
     "express": "^4.13.4",
     "mocha": "^2.4.5",
-    "parse-server": "2.2.17"
+    "parse-server": "^2.4.2"
   },
   "scripts": {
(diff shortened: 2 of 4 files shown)


flovilmart:
@dplewis yes the objects are not cleared as the test objects are just created once for the spec. Could you update the tests so they are using the 

```
      let sacramento = new TestPoint();
      sacramento.set('location', new Parse.GeoPoint(38.52, -121.50));
      sacramento.set('name', 'Sacramento');

      let honolulu = new TestPoint();
      honolulu.set('location', new Parse.GeoPoint(21.35, -157.93));
      honolulu.set('name', 'Honolulu');

      let sf = new TestPoint();
      sf.set('location', new Parse.GeoPoint(37.75, -122.68));
      sf.set('name', 'San Francisco');

```

This way, your tests will just reuse the fixtures already used.

dplewis:
sorry for the late reply tests are working

flovilmart:
LGTM! Thanks!

PR #446, opened 2017-06-17
Properly logs out user between tests

flovilmart:


flovilmart:
Merging as just affecting integration tests with parse-server

PR #448, opened 2017-06-25
Parse.Query.fromJSON method

kyptov:
Made mistake with previous PR. This is new one.
Method restores query from json gained from Parse.Query.toJSON()

### DIFF
--- a/src/ParseQuery.js
+++ b/src/ParseQuery.js
@@ -269,7 +269,7 @@ export default class ParseQuery {
    * query.equalTo(key: value);
    * query.limit(100);
    * ... (others queries)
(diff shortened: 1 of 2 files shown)


kyptov:
@flovilmart no problem, do you mean this?
* rename `fromJSON` -> `withJSON`,
* add new static method `fromJSON` which creates new Parse.Query and calling `withJSON` method
* some test to cover this

flovilmart:
@kyptov, yes that’s what I had in mind. Does that makes sense for you?

kyptov:
@flovilmart made some changes

flovilmart:
@kyptov looking great! Thanks!

PR #455, opened 2017-07-06
Add Type Polygon to Schema and PolygonContains to query

dplewis:
New Polygon Type and new Query for finding all polygons that contains a point

https://github.com/parse-community/parse-server/pull/3944

Waiting for 2.6.0 release


dplewis:
@flovilmart can you look at this?

flovilmart:
This looks great man!

PR #457, opened 2017-07-06
Corrects Guide Link in README.md

montymxb:
Just a small fix of `https` to `http` for the JS guide link, which is only available over `http`.


montymxb:
A quick note the description for this repo...
```
Parse SDK for JavaScript https://parse.com
```
uses the outdated https://parse.com link. This should probably be `http://parseplatform.org` or something that's equally relevant.

PR #458, opened 2017-07-06
Addresses flaky test

flovilmart:


flovilmart:
@dplewis you may wanna have a look on this flaky test, odd that it runs locally properly :)

dplewis:
@flovilmart Thanks for addressing this. I've been avoiding using .catch() and .fail() and only done.fail. Is `REQUEST_ATTEMPT_LIMIT` in parse server?

flovilmart:
No it’s in the JS SDK, it retired on errors, the skipped test works locally but not on the CI, or at least inconsistently.

dplewis:
I see you have fixed a lot of flaky tests on JS SDK and parse server. Have you seen a pattern with inconsistent test? This would help with future tests.

Does `this.skip` come from mocha?

I try to follow this javascript guide for ES6 conventions. Maybe include this or something similar in CONTRIBUTION.md
https://github.com/airbnb/javascript

flovilmart:
Skip comes from mocha yes, it let you keep a test in your suite explicitly skipping it. I don’t really understand why it sometimes fails,

PR #459, opened 2017-07-08
Add methods `addAll`, `addAllUnique` and `removeAll`

arniu:


flovilmart:
Thanks for the feature, can you please not alter existing tests and instead write new tests for that feature?

flovilmart:
@arniu on the implementation side, those all all the same as add, addUnique, remove etc... Why not just change the accepted type to `mixed | Array<mixed>` as it is the same underlying implementation.

arniu:
@flovilmart This is mainly due to compatibility considerations:

1. Keep the API consistent with other SDK's;
2. Keep backward compatible.

In the old version, you can use `add` to add tuples to the array. For example, 

```javascript
var obj = new Parse.Object('TestObject')
obj.add('array', [1, 2])
obj.add('array', [1, 2])
```

The result is `[[1, 2], [1, 2]]`.

If you allow `add` batch to add objects, the result will be `[1, 2, 1, 2]`.

flovilmart:
My bad, I misread the code, as it looked very similar :)

PR #461, opened 2017-07-09
Support React Native < 0.43 once again

arniu:


flovilmart:
@arniu alright, is it working for you correctly?

PR #465, opened 2017-07-22
support creating Parse.File using uri on react-native

yomybaby:
## Before this PR
When we create Parse.File on react-native, we should transfer to base64 from file uri.
 It is **very inefficient** to transfer large quantities of binary data between JS and native code.
```
ImageStore.getBase64ForTag("rct-image-store://0", (base64)=>{
  var f = new Parse.File('my.jpg',{
    base64
  });
 f.save();
});
```

## After this PR
You can create Parse.File using uri directly (see blow code).  Under the hood, `uri` will be passed to `RCTNetworkingNative`.  It transfers only uri string from JS to native code. 😄 
```
var file = new Parse.File('my.jpg',{uri:"rct-image-store://0"});
file.save();
```


runia1:
What is the status of this PR?  This is exactly what I need but it looks dead :/

dplewis:
Sorry for the late reply. I think it would be better to make a feature to support all uri, I know it’s doable in the PHP SDK.

Also I don’t see how the uri would get pass under the hood to RCTNetworkingNative. Have you tried your fix in your project?

@acinader Thoughts

dplewis:
Closing via https://github.com/parse-community/Parse-SDK-JS/pull/749

yomybaby:
@dplewis Thanks for review and make new PR. In react-native, there is no need to download if uri for local assets(Photo Library, File system..). I'll check your PR and make new PR for react-native if needed.

dplewis:
No problem, I’m was trying to wrap my head around file.save() in your example. Like how would the uri / data get sent to the server. I don’t think the RestController would allow it.

PR #469, opened 2017-07-29
Several Changes for ParseQuery API

jcguarinpenaranda:
* Implemented method `copy():ParseQuery` that returns a copy of the current query
* Added parameter count:boolean to `find` method, that if defined allows you to get an object with attributes `{ count: number, results: ParseObject[] }` with the object count that matches the query and the results
* Implemented method `_removeCondition` that allows deleting query conditions after set
* Implemented method `clearMatch` that allows clearing the regex condition for a given key

### DIFF
--- a/src/ParseQuery.js
+++ b/src/ParseQuery.js
@@ -228,6 +228,24 @@ export default class ParseQuery {
     this._where[key][condition] = encode(value, false, true);
     return this;
   }
+  
+  /**
+   * Helper for removing conditions on the query
+   */
+  _removeCondition(key: string, condition: string): ParseQuery {
+    if (this._where[key] && this._where[key][condition]) {
+      delete this._where[key][condition]
+    }
+    return this;
+  }
+
+  /**
+   * Returns a copy of the current query
+   * @method copy
+   */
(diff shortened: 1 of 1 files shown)


flovilmart:
Thanks for the PR, however we will not move forward with all the changes as they introduce breaking changes:

> Implemented method copy():ParseQuery that returns a copy of the current query

Nice, this is an intereesting feature, but be careful the underlying objects in your implementation may end up being shared.

> Added parameter count:boolean to find method, that if defined allows you to get an object with attributes { count: number, results: ParseObject[] } with the object count that matches the query and the results

This does not work as advertised and isn't tested. 

> Implemented method _removeCondition that allows deleting query conditions after set

Why make it a private looking method? I woudl argue taht mutating and removing conditions is a bad patterns that we try not to encourage, the best is to consider queries with a short lifespan and rebuild them from your app props / state.

> Implemented method clearMatch that allows clearing the regex condition for a given key

Too specific, and very redundant with `_removeCondition`

Overall, the clone feature is interesting, and I would not encourage for the others to be baked in as they may either not work as expected or feature side effects that are unexpected.

PR #477, opened 2017-09-09
fix some interactions between select queries and sub objects

### LINKED ISSUE #475: Using Query.select() will break ParseObject.has() on undefined field
Ref: https://github.com/parse-community/parse-server/issues/3999
Caused: https://github.com/parse-community/Parse-SDK-JS/pull/409
```
  const obj = new Parse.Object('TestObject'); // Yay no foo
    obj.save().then(() => {
      const query = new Parse.Query('TestObject');
      query.select('foo');
      return query.get(obj.id);
    }).then((result) => {
      expect(result.has('foo')).toBe(false); // result.has('foo') returns true ?!?? :(
      done();
    }).catch(done.fail);
```

Up for grabs

dstarke:
Fixes some issues introduced by #409, and I think should fix #475.

### DIFF
--- a/src/ParseQuery.js
+++ b/src/ParseQuery.js
@@ -67,10 +67,12 @@ function handleSelectResult(data: any, select: Array<string>){
 
       pathComponents.forEach((component, index, arr) => {
         // add keys if the expected data is missing
-        if (!obj[component]) {
-          obj[component] = (index == arr.length-1) ? undefined : {};
+        if (obj && !obj.hasOwnProperty(component)){
+          obj[component] = undefined;
         }
-        obj = obj[component];
+        if(obj !== undefined){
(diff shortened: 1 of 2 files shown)


dstarke:
Ref: parse-community/parse-server#3999
This should address the issues described there.

flovilmart:
Thanks @dstarke

flovilmart:
@dstarke I just tested the integration with parse-server and it breaks 2 tests:

```
    ✗ select nested keys 2 level (issue #1567)
      - TypeError: Cannot read property 'bazoo' of undefined
    ✗ select nested keys 2 level without include (issue #3185)
      - TypeError: Cannot read property 'bazoo' of undefined
```

I'll have a look

PR #483, opened 2017-09-21
Update ParseFile to handle a data URI containing a number

AppOrchestra:
The regex to extract the components from a base64 encoded string doesn't handle mime types which contain a number, e.g. audio/m4a.

It was failing with the error:
TypeError: Cannot read property '1' of null
at ParseFile (/app/node_modules/parse/lib/node/ParseFile.js:121)

The regex has been updated to allow numbers, and a test added which failed before the change.

### DIFF
--- a/src/ParseFile.js
+++ b/src/ParseFile.js
@@ -25,7 +25,7 @@ export type FileSource = {
 };
 
 var dataUriRegexp =
-  /^data:([a-zA-Z]*\/[a-zA-Z+.-]*);(charset=[a-zA-Z0-9\-\/\s]*,)?base64,/;
+  /^data:([\w\/\+]+);(charset=[a-zA-Z0-9\-\/\s]*,)?base64,/;
 
 function b64Digit(number: number): string {
   if (number < 26) {
--- a/src/ParseFile.js
+++ b/src/ParseFile.js
@@ -25,7 +25,7 @@ export type FileSource = {
 };
 
 var dataUriRegexp =
-  /^data:([\w\/\+]+);(charset=[a-zA-Z0-9\-\/\s]*,)?base64,/;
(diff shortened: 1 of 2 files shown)


src/ParseFile.js
```diff
@@ -25,7 +25,7 @@ export type FileSource = {
 };
 
 var dataUriRegexp =
-  /^data:([a-zA-Z]*\/[a-zA-Z+.-]*);(charset=[a-zA-Z0-9\-\/\s]*,)?base64,/;
+  /^data:([\w\/\+]+);(charset=[a-zA-Z0-9\-\/\s]*,)?base64,/;
```
montymxb: @AppOrchestra Nice catch regarding the inability to match digits. Although the way you've set up this regex would also match things with an arbitrary # of slashes, like `too//many//slashes/here`. In addition `\w` (if I'm correct regarding the regex flavor) will probably match `_` as well, which is not specifically desired.

To avoid the included `_` character with `\w`, and to factor in digits, you could probably do something like:
```
[-a-zA-Z0-9]+\/[-a-zA-Z0-9+.]+
```
This basically uses the existing regex, but makes a couple changes. It adds a `-` to the first part of the regex, allowing for valid mime types with a `-` at the start, like `x-conference/x-cooltalk` (which is a thing, although rare). It also changes the quantifier `*` for a `+`, specifying that we expect at least one or more characters before and after the slash, something like `/mimetype` or `audio/` shouldn't be considered valid; the existing regex would allow this.

You could also just use `\d` instead of `0-9`, totally up to you, just wanted to give you a heads up on that.

AppOrchestra:
I had a proper look at a few pages:
https://tools.ietf.org/html/rfc6838 (4.2)
https://en.wikipedia.org/wiki/Data_URI_scheme
https://www.freeformatter.com/mime-types-list.html

(instead of copying what a popular data uri regex had...) and updated the regex to handle all valid mime-types, with some basic validation, and also fixed it to handle the charset param too.

PR #485, opened 2017-10-01
Support for Aggregate Queries

dplewis:
Support for `distinct()` and `aggregate()` queries.

Pending: https://github.com/parse-community/parse-server/pull/4207

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 ---
 language: node_js
 node_js:
-- '5'
+- '6.11.4'
 
 branches:
   only:
--- a/integration/package.json
+++ b/integration/package.json
@@ -3,7 +3,7 @@
   "dependencies": {
     "express": "^4.13.4",
     "mocha": "^2.4.5",
-    "parse-server": "^2.6.0"
+    "parse-server": "^2.7.0"
   },
   "scripts": {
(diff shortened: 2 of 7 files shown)


oallouch:
That's really great. All the power of MongoDB in Parse !
Do you have prepared a little documentation update for this ?

dplewis:
@oallouch Documentation will be coming soon!

dplewis:
@flovilmart Can you check this?

flovilmart:
code looks good, but would it be too much to ask for added Jest tests for this class?

flovilmart:
Thanks @dplewis for this! 🙏

dplewis:
@flovilmart Quick question. I was lazy and didn't require aggregate in the query controller https://github.com/dplewis/Parse-SDK-JS/blob/aggregate/src/CoreManager.js#L272

Is that ok? I can update all the tests if needed.

flovilmart:
That would be better to add those methods now so we don’t have to come back to it later.

flovilmart:
CoreManager-test is failing (it specifically validates the interface)

flovilmart:
Perfect! Waiting for CI but it all looks great!

dplewis:

flovilmart:
Perfect 👌🏻

PR #495, opened 2017-10-23
Fix #494: Reject a promise on error condition

### LINKED ISSUE #494: Promise is never resolved in case of server-side error when invoking Parse.Object.saveAll
We have a node process that works on multiple batches of Parse objects. Each batch is saved through Parse.Object.saveAll. However, as soon as a batch errors out, the node process stops immediately and none of the subsequent batches get processed. We found that an error handling component was missing. 

Raising a PR shortly. It adds just two lines of code.

nrchandan:


src/ParseObject.js
```diff
@@ -1847,6 +1847,8 @@ var DefaultController = {
             }, options);
           }).then((response, status, xhr) => {
             batchReturned.resolve(response, status);
+          }, (error) => {
+            batchReturned.reject(new _ParseError2.default(undefined, error.message));
```
flovilmart: this seems wrong, should probably be `new ParseError(error.message);`

nrchandan:
Thanks for pointing it out. Will correct and push shortly.

flovilmart:
Also consider adding tests as you are changing a public interface, and this may be unexpected for many users.

nrchandan:
Tried going through the tests in ParseObjectTest.js. Currently, there are no tests that cause failure of saveAll. I'm not sure how to make it fail.

nrchandan:
I think I can make it fail. Let me give it a try.

integration/test/ParseObjectTest.js
```diff
@@ -254,6 +254,14 @@ describe('Parse Object', () => {
     });
   });
 
+  it('cannot set an invalid date', (done) => {
+    let obj = new TestObject();
+    obj.set('when', new Date(Date.parse(null)));
+    Parse.Object.saveAll([obj]).fail((e) => {
```
flovilmart: Please add .then(done.fail)

integration/test/ParseObjectTest.js
```diff
@@ -254,6 +254,14 @@ describe('Parse Object', () => {
...
+      done();
```
flovilmart: Please test for the expected error message.

flovilmart:
Also, the coverage is not improved as you’ve added an integration test and not a jest unit test. You may wanna consider doing that as there’s no way for now to ensure the code path is properly covered.

nrchandan:
That makes sense. Let me try adding a unit test.

nrchandan:
Found two duplicate unit tests (both in the unit test file `ParseObject-test.js`):
`can save an array of objects` is repeated twice (no diff)
`can save the object` appears twice with a negligible difference (objectId: 'P1' vs 'P5')

Should I remove these two in the same PR, or would you recommend creating a new PR for this?

flovilmart:
I prefer we don’t remove unit tests, PR’s should be as atomic as possible.

nrchandan:
Thanks Florent. Let me undo that and push again.

nrchandan:
Added a new error code: 
```
/**
 * Error code indicating an invalid value of an object field is passed.
 * @property INVALID_VALUE
 * @static
 * @final
 */
ParseError.INVALID_VALUE = 603
```

Also, updated the tests to check for both error message and this new error code.

nrchandan:
Unit test has passed, but integration failed. Both the tests are very similar. @flovilmart Could you help me fix it, or would it make sense to remove it instead?

nrchandan:
Hi @flovilmart, I made a couple of changes. Could you take a look at this revised PR and let me know if this looks good for merging.

flovilmart:
This is looking good now! Let's go!

PR #497, opened 2017-10-23
Add unit test towards #494

nrchandan:
This shouldn't fail ideally.

### DIFF
--- a/integration/test/ParseObjectTest.js
+++ b/integration/test/ParseObjectTest.js
@@ -254,6 +254,14 @@ describe('Parse Object', () => {
     });
   });
 
+  it('cannot set an invalid date', (done) => {
+    let obj = new TestObject();
+    obj.set('when', new Date(Date.parse(null)));
+    Parse.Object.saveAll([obj]).fail((e) => {
+      done();
+    });
+  });
+
   it('cannot create invalid class names', (done) => {
     let item = new Parse.Object('Foo^Bar');
     item.save().fail((e) => {


integration/test/ParseObjectTest.js
```diff
@@ -254,6 +254,14 @@ describe('Parse Object', () => {
     });
   });
 
+  it('cannot set an invalid date', (done) => {
+    let obj = new TestObject();
+    obj.set('when', new Date(Date.parse(null)));
+    Parse.Object.saveAll([obj]).fail((e) => {
```
flovilmart: This actually succeeds , but failed the CI as the done() is never called.

nrchandan: I just pushed this along with the other PR. This test passes after PR #495. So, we know that the unit test covers this fail condition. However, coverage is still down. What should I do?

nrchandan: I should also mention that I copy pasted this test case from just above this unit test. That one invokes save instead of saveAll, and passes. It is the saveAll that was misbehaving.

flovilmart:
You should probably put the test and the fix in the same PR. Also your test is broken, as you should have a `done.fail` in the ‘success’ handler if that’s your intention to test.

nrchandan:
The test uses the same convention followed in the tests surrounding it. I
have merged this test with the original PR. I just wanted to see if the
test fails without my fix, and it doesn't. So, my fix works. Could you
please take a look at #495. I will close this one out.

nrchandan:
I just wanted to see if the test fails without my fix, and it does.

nrchandan:
Closing this one as it is incorporated in #495

PR #512, opened 2017-11-27
Fixes jsdoc generation, adds jsdoc support

flovilmart:
This PR fixes a few issues when generating docs with JS docs, See the generated docs here:

https://flovilmart.github.io/parse-js-docs-demo/index.html

If we're all ok with it, I'll push to the gh-pages branch and automate the release on tags.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,5 @@ integration/test_logs
 .DS_Store
 .idea/
 integration/test_logs
+out/
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 ---
 language: node_js
 node_js:
-- '5'
+- '6.11.4'
 
 branches:
   only:
--- a/jsdoc-conf.json
+++ b/jsdoc-conf.json
@@ -5,7 +5,6 @@
   },
   "source": {
     "include": ["./package.json", "./README.md"],
-    "exclude": [ /* array of paths to exclude */ ],
     "excludePattern": "(^|\\/|\\\\)_"
   },
   "templates": {
--- a/jsdoc-conf.json
+++ b/jsdoc-conf.json
@@ -4,9 +4,8 @@
       "plugins": ["transform-flow-strip-types"]
   },
   "source": {
-    "include": ["./package.json", "./README.md", "./src/Analytics.js", "./src/Cloud.js", "./src/Push.js", "./src/FacebookUtils.js", "./src/LiveQueryClient.js", "./src/LiveQuerySubscription.js" ],
+    "include": ["./package.json", "./README.md"],
     "exclude": [ /* array of paths to exclude */ ],
-    "includePattern": "src/Parse.+\\.js(doc|x)?$",
     "excludePattern": "(^|\\/|\\\\)_"
   },
   "templates": {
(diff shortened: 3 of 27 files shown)


dplewis:
Some of the pages have duplicate sections.

https://flovilmart.github.io/parse-js-docs-demo/Parse.LiveQuerySubscription.html

flovilmart:
@dplewis yeah I've noticed that as well, I'm not sure what's causing it, and why?

flovilmart:
I need to find out why the docs are duplicated...

flovilmart:
Other question/ more management wise. Do we keep the api references in each gh-pages or do we move them on the docs repo?

dplewis:
I tried it locally earlier and ran `npm run docs` and got a different result than https://flovilmart.github.io/parse-js-docs-demo/index.html

When I removed `includePattern` everything looks ok.

dplewis:
@flovilmart I prefer the gh-pages. Looking at the php sdk, gh-pages is far ahead of php docs.

flovilmart:
@dplewis it seems that removing those configs did the trick, check again the demo repo, all look good now, some private (_) methods are still missing some docs, but we can live with it.

dplewis:
@flovilmart jsdoc is weird. includes > includePattern. But if the includes doesn't match the includePattern it gets ignored.

flovilmart:
Yeah I figured the hard way, now the docs look good though!
I'll open a PR against gh-pages.;

flovilmart:
And it's live! http://parseplatform.org/Parse-SDK-JS/api/1.10.2/

dplewis:
lgtm!

flovilmart:
I need to automate that on release at one point... May not be trivial but we'll see.

flovilmart:
we could use that: https://docs.travis-ci.com/user/deployment/pages/
But because of the nature of gh-pages deployment / force push, this would override any existing content (unless cloned beforehand)

dplewis:
@montymxb advice on gh-pages deployment?

montymxb:
@dplewis flovilmart is correct, it would be an overriding change. In the case of the php sdk we force gh-pages updates on [deploying to the master only](https://github.com/parse-community/parse-php-sdk/blob/master/.travis.yml#L45...L55). It is a forced change which will overwrite existing changes in gh-pages. Depending on the setup this could be undesirable if you have people adding changes by hand from time to time, as their work could be undone.

In our case the entire set of docs is regenerated every time anyways. Automating it as part of a PR merging into master allows the docs to always be up to date as soon as changes are accepted. Just one less thing we have to do that we would be doing anyways, so it makes sense for us there. If it's a similar case here it might be equally useful to do something similar. If anyone is (or will be) adding changes between updates it would be a problem if we did that.

Another thing, I put something together over there that [ensures that the doc that would be generated for a given PR must be valid](https://github.com/parse-community/parse-php-sdk/blob/master/.travis.yml#L42). No more accidental missing docs for methods, vars, etc. If doc generation was automated here I would recommend something similar to help ensure a complete doc.

flovilmart:
Yep, that’s a good idea, but we loose library versioning :/

montymxb:
Hmm...if we want to preserve it we can checkout the existing gh-pages, archive the contents in a folder, and then proceed to make our brute-force update. If you did it right you could archive on the releases specifically, and end up with a nice set of folders containing the doc at that point in time. With the forefront docs being the latest set.

flovilmart:
That’s what I’m thinking about, i’ll Try to come up with a solution

PR #518, opened 2017-12-01
Support fields in Live Query subscriptions

TimNZ:
Use field keys specified via query.select() as fields value for subscribe.

Default fields of [className,objectId, createdAt, updatedAt, ACL] are always sent by server, which means the LQ subscribe message description for query.fields is inaccurate.

https://github.com/parse-community/parse-server/wiki/Parse-LiveQuery-Protocol-Specification#subscribe-message

I don't think ACL should be in the default fields but I don't know yet if it's used in the client somehow.  Looking at that is another task.

### DIFF
--- a/src/LiveQueryClient.js
+++ b/src/LiveQueryClient.js
@@ -186,14 +186,17 @@ export default class LiveQueryClient extends EventEmitter {
     if (!query) {
       return;
     }
-    let where = query.toJSON().where;
     let className = query.className;
+    let queryJSON = query.toJSON();
+    let where = queryJSON.where;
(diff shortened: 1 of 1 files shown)


flovilmart:
Perfect @TimNZ ! Just checked and you're 💯 parse-server supports the fields option from the get-go!

montymxb:
Thanks a ton @TimNZ !

TimNZ:
I'm not a coder wizz but always interested in following best/popular practice

Would you have written:
```
let fields = queryJSON.keys ? queryJSON.keys.split(',') : undefined;
```

as 
```
let fields = queryJSON.keys && queryJSON.keys.split(',')
```

PR #522, opened 2017-12-03
Docs releases

flovilmart:
- Adds docs for cloud code methods (original methods, still missing beforeFind)
- Adds automated deployment of docs to `gh-pages` with preservation of current branch.
- Adds tagged deployment of the SDK automatically to npm.
- Adds tagged deployment of docs in <tag> folder in docs.
- Improve performance of travis-ci by enabling npm caching
- Use minami template for docs rendering, nicer, probably more flexible.


.travis.yml
```diff
@@ -1,19 +1,49 @@
...
+      after_script: skip
+      script: npm run release_docs
+      deploy:
+        - provider: pages
+          skip_cleanup: true
+          github_token: $GITHUB_TOKEN # Set in travis-ci.org dashboard
+          local_dir: docs/
+          on:
+            all_branches: true
+        - provider: npm
```
montymxb: Me Gusta! Awesome that this would cut out a step we would have to do by hand 👍 .

flovilmart: Yes! Figured we should do as much as we could!

release_docs.sh
```diff
@@ -0,0 +1,28 @@
...
+git pull origin gh-pages
+cd ..
+
+DEST="master"
+
+if [ "${TRAVIS_TAG}" != "" ];
+then
+  DEST="${TRAVIS_TAG}"
+  # change the default page to the latest
+  echo "<meta http-equiv='refresh' content='0; url=/Parse-SDK-JS/api/${DEST}'>" > "docs/api/index.html"
```
montymxb: This is interesting... So per release we're just redirecting via meta to the latest set? That's kind of cool actually. One thing I'm curious about is if we're maintaining an index or a table of contents for existing doc versions?

flovilmart: So far we don’t, but with this strategy we’ll be able to :)

flovilmart:
Ci passes! Once merged this should put the master doc set

flovilmart:
Yes! https://travis-ci.org/parse-community/Parse-SDK-JS/jobs/310729345

flovilmart:
And YES! http://parseplatform.org/Parse-SDK-JS/api/master/

montymxb:
Now that's a sight to see 🍻 .

flovilmart:
Took a few weekends to get there, next is all iOS repos! We can later customize the theme/template if we have time, to make it more parse-like . There's [docstrap](https://github.com/docstrap/docstrap) that offers a nice template, more customizable, and with more fluid navigation.

Issue #476, opened 2017-09-06
Full Text Search


dblythy:
Any news on when query.fullTextSearch will be available?

Cheers

seriph:
@dblythy I've had this working since April and just updated it to be in synch with the latest version of the JS SDK. You're welcome to use the following until we get it merged in:
https://www.npmjs.com/package/parse-text
If you're using Mongo, I would recommend using the following as well. It's the latest version of the Parse Server, but will work properly for compound text indices (you have to create them yourself).
https://www.npmjs.com/package/parse-server-text

dplewis:
@dblythy There is a PR pending https://github.com/parse-community/Parse-SDK-JS/pull/470, Don't know why that one hasn't been merged yet.

russell-dot-js:
I just reviewed #470, could use a little cleanup prior to merge IMO

montymxb:
We're all good, full text support has been merged. Once the next release rolls out everyone will be able to utilize this. Still pending formal docs update to our [js docs](http://docs.parseplatform.org/js/guide/), which may precede the release of this feature.

PR #470, opened 2017-08-02
Add the new fullTextSearch method with the associated tests.

SebC99:
To be used with parse-server version > 2.5.0

PR #541, opened 2018-01-10
Support for Full Text Search Query

dplewis:
Supersed: https://github.com/parse-community/Parse-SDK-JS/pull/470
Closes: https://github.com/parse-community/parse-server/issues/4490, https://github.com/parse-community/Parse-SDK-JS/issues/476

To be used with Parse-Server 2.5.0+

* Fixed docs typos
* Lint Test files
* Remove unnecessary code from past tests
* Bump parse-server to 2.7.1

### DIFF
--- a/src/__tests__/ParseQuery-test.js
+++ b/src/__tests__/ParseQuery-test.js
@@ -1864,4 +1864,68 @@ describe('ParseQuery', () => {
     });
   });
 
+  it('full text search', () => {
+    const query = new ParseQuery('Item');
+    query.fullText('size', 'small');
+
+    expect(query.toJSON()).toEqual({
+      where: {
+        size: {
+          $text: {
+            $search: {
+              $term: "small"
+            }
+          }
+        }
+      }
+    });
+  });
+
+  it('full text search sort', () => {
+    const query = new ParseQuery('Item');
+    query.fullText('size', 'medium');
+    query.ascending('$score');
+    query.select('$score');
+
+    expect(query.toJSON()).toEqual({
+      where: {
+        size: {
+          $text: {
+            $search: {
+              $term: "medium",
+            }
+          }
+        }
+      },
+      keys : "$score",
+      order : "$score"
+    });
+  });
+
+  it('full text search key required', (done) => {
+    try {
+      const query = new ParseQuery('Item');
+      query.fullText();
+    } catch (e) {
+      done();
+    }
+  });
+
+  it('full text search value required', (done) => {
+    try {
+      const query = new ParseQuery('Item');
+      query.fullText('key');
+    } catch (e) {
+      done();
+    }
+  });
+
+  it('full text search value must be string', (done) => {
+    try {
+      const query = new ParseQuery('Item');
+      query.fullText('key', []);
+    } catch (e) {
+      done();
+    }
+  });
 });
--- a/src/__tests__/ParseQuery-test.js
+++ b/src/__tests__/ParseQuery-test.js
@@ -1903,29 +1903,20 @@ describe('ParseQuery', () => {
   });
 
   it('full text search key required', (done) => {
-    try {
-      const query = new ParseQuery('Item');
-      query.fullText();
-    } catch (e) {
-      done();
-    }
+    const query = new ParseQuery('Item');
+    expect(() => query.fullText()).toThrow('A key is required.');
+    done();
   });
 
   it('full text search value required', (done) => {
-    try {
       const query = new ParseQuery('Item');
-      query.fullText('key');
-    } catch (e) {
+      expect(() => query.fullText('key')).toThrow('A search term is required');
       done();
-    }
   });
 
   it('full text search value must be string', (done) => {
-    try {
-      const query = new ParseQuery('Item');
-      query.fullText('key', []);
-    } catch (e) {
-      done();
-    }
+    const query = new ParseQuery('Item');
+    expect(() => query.fullText('key', [])).toThrow('The value being searched for must be a string.');
+    done();
   });
 });
--- a/src/__tests__/ParseSchema-test.js
+++ b/src/__tests__/ParseSchema-test.js
@@ -13,15 +13,6 @@ var ParseSchema = require('../ParseSchema').default;
 var ParsePromise = require('../ParsePromise').default;
 var CoreManager = require('../CoreManager');
 
-function generateSaveMock(prefix) {
-  return function(name, payload) {
-    return ParsePromise.as({
-      name: name,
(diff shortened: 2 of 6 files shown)


dplewis:
I request a review from @flovilmart and @SebC99

SebC99:
Why don't you use the diacritics, language and casesensitive options?
As for the doc I know the rest call needs a key but mongo doesn't, so we shouldn't say in the docs that the key will be the search field as mongo will use the text index without looking at the key.
Apart from that I'm ok with it! Thanks!

dplewis:
That is for Parse-Server as a service. We can include them but we have to think about the current and future database adapters. Postgres requires a key for example.

src/__tests__/ParseSchema-test.js
```diff
@@ -13,15 +13,6 @@ var ParseSchema = require('../ParseSchema').default;
 var ParsePromise = require('../ParsePromise').default;
 var CoreManager = require('../CoreManager');
 
-function generateSaveMock(prefix) {
```
montymxb: Unused?

src/__tests__/ParseSchema-test.js
```diff
@@ -388,11 +379,9 @@ describe('SchemaController', () => {
   beforeEach(() => {
     CoreManager.setSchemaController(defaultController);
     var request = function(method, path, data, options) {
-      var name = path.substr(path.indexOf('/') + 1);
       return ParsePromise.as([]);
     };
     var ajax = function(method, path, data, headers) {
-      var name = path.substr(path.indexOf('/') + 1);
```
montymxb: Unused here as well?

src/__tests__/ParseQuery-test.js
```diff
@@ -1864,4 +1864,68 @@ describe('ParseQuery', () => {
...
+      keys : "$score",
+      order : "$score"
+    });
+  });
+
+  it('full text search key required', (done) => {
+    try {
+      const query = new ParseQuery('Item');
+      query.fullText();
+    } catch (e) {
```
montymxb: Rather than just try/catching I would use `expect(...).toThrow('my error...')`, that would be much more explicit in terms of that we are expecting and what we are expecting to receive as the error. You can see an example in the [parse-server tests](https://github.com/parse-community/parse-server/blob/8c2c76dd269f5aaf71439acb7a42a971a6560e49/spec/AdaptableController.spec.js#L27...L29).

src/__tests__/ParseQuery-test.js
```diff
@@ -1864,4 +1864,68 @@ describe('ParseQuery', () => {
...
+      done();
+    }
+  });
+
+  it('full text search value required', (done) => {
+    try {
+      const query = new ParseQuery('Item');
+      query.fullText('key');
+    } catch (e) {
```
montymxb: Same here.

src/__tests__/ParseQuery-test.js
```diff
@@ -1864,4 +1864,68 @@ describe('ParseQuery', () => {
...
+      done();
+    }
+  });
+
+  it('full text search value must be string', (done) => {
+    try {
+      const query = new ParseQuery('Item');
+      query.fullText('key', []);
+    } catch (e) {
```
montymxb: And here

src/__tests__/ParseSchema-test.js
dplewis: Yeah I copied the jest test from Files when I added schema support. Forgot I left those in 😅

dplewis:
@montymxb Thanks for the review, changes added.

dplewis:
@montymxb I'll start working on the docs after this gets in

montymxb:
That would be good, that and aggregate as mentioned in #543 (we do have that here right?).

dplewis:
Yes, quick question should we have documentation in docs folder, the gh-pages branch or both?

montymxb:
We'll want docs in the docs repo separately, but you should be good here.

dplewis:
@flovilmart Shouldn't http://parseplatform.org/Parse-SDK-JS/api/master/ be in the README so new users can see it. I can add some examples to be auto generated for aggregate like I did for this PR.

flovilmart:
This is linked into http://docs.parsplatform.org/JS/api

dplewis:
@flovilmart Ah, I see. I'm so used to going to the guide in the readme instead of the api reference.

flovilmart:
Yeah, the guide is for the usage, also we have the API references in each repositories

montymxb:
We should have a link to the API reference on the README 🔖

flovilmart:
yep those are:

http://parseplatform.org/Parse-SDK-JS/api -> lastest release

http://parseplatform.org/Parse-SDK-JS/api/v1.11.0/ -> tagged release

http://parseplatform.org/Parse-SDK-JS/api/master/ -> master branch

montymxb:
@flovilmart probably should clarify specifically from the README on github repo page, not already from within the parseplatform.org location. https://github.com/parse-community/Parse-SDK-JS/.

flovilmart:
Yes I agree ! let's do it, I just posted the links for reference.

srameshr:
How do I install the version where this feature has been merged?

srameshr:
Guys, can you release a version with this fix or update 1.11.0 to include this?

srameshr:
Any updates?

montymxb:
@srameshr since this is an enhancement we'll probably wait a bit to bundle this with other features (if any are pending) and any fixes we may want. There's no official release schedule here, but if you feel that what we currently have is sufficient for a release go ahead and open up a separate issue for that. We can focus on talking about a release there, rather than mixing topics on this PR thread.

srameshr:
@montymxb Yup, will do that.
But, I also think opening up an issue just to **remind** some other fix has to be merged would not make much sense.

montymxb:
If you open up another issue keep it focused on a request for a release and your reason why. We may still hold it back if we decide we want to add more than what we currently have. This isn't necessarily another fix that _needs_ to be merged, this just brings it to our attention so we can evaluate it formally.

srameshr:
Shouldn't what **needs** to be merged, be evaluated on current needs of the community / users instead of how **big** the fix is or **you deciding** what is more than what you currently have?

montymxb:
Absolutely, so please open up an issue for that topic so we can all deliberate on that specifically. Since this is a topic for a given PR that has been merged I would consider what you're talking about as a separate topic, i.e. drafting a release. Moving the convo there will allow us to have that focused on input for said topic specifically, and if it makes sense we'll go ahead and draft.

PR #544, opened 2018-01-16
Adds Purge & Polygon to Parse.Schema

dplewis:
This adds the purge method to ParseSchema. This allows the deletion of all objects without having to run through and query all objects for removal. This was added to ParseSchema as the related delete method requires the class to be first empty of objects.

Polygon type support has also been added to ParseSchema, both via a supported method to add a polygon type field and as a recognized field when asserting types.

### DIFF
--- a/integration/test/ParseSchemaTest.js
+++ b/integration/test/ParseSchemaTest.js
@@ -47,6 +47,7 @@ describe('Schema', () => {
       .addDate('dateField')
       .addFile('fileField')
       .addGeoPoint('geoPointField')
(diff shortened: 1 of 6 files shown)


dplewis:
I wish I could merge this :)

flovilmart:
@dplewis I’ll add you to the JS maintainers :)

dplewis:
@flovilmart 🙏

montymxb:
Awesome, another person to help with PRs here!

dplewis:
@montymxb Welcome back

PR #565, opened 2018-04-27
Allow specific keys to be reverted in unsaved objects

omairvaiyani:
Currently, the ```Parse.Object``` prototype method ```revert``` resets ALL unsaved changes. With this PR, specific fields can be provided to the function to allow selective reversion.

Currently:
```javascript
var person =  ParseObject.fromJSON({
      className: 'Person',
      objectId: 'xxxx',
      name: 'Omair'
 });
person.set('gender', 'male');
person.set('age', 25);
// 'age' is ignored, all fields will be reverted
person.revert('age');
// person -> { name: 'Omair' }
```

With PR (test cases in commit):
Example 1 (single field):
```javascript
var person =  ParseObject.fromJSON({
      className: 'Person',
      objectId: 'xxxx',
      name: 'Omair'
 });
person.set('gender', 'male');
person.set('age', 25);
person.revert('age');
// person -> { name: 'Omair', gender: 'male' }
```

Example 2 (multiple fields):
```javascript
var person =  ParseObject.fromJSON({
      className: 'Person',
      objectId: 'xxxx',
      name: 'Omair'
 });
person.set('gender', 'male');
person.set('age', 25);
person.set('occupation', 'Doctor');
person.revert(['gender', 'occupation']);
// person -> { name: 'Omair', age: '25' }
```

Example 3 (no fields, works as now):
```javascript
var person =  ParseObject.fromJSON({
      className: 'Person',
      objectId: 'xxxx',
      name: 'Omair'
 });
person.set('gender', 'male');
person.set('age', 25);
person.set('occupation', 'Doctor');
person.revert();
// person -> { name: 'Omair' }
```


flovilmart:
Can we make this backwards compatible please? Now if no keys are passed the SDK an exception will be thrown.

omairvaiyani:
@flovilmart I'm getting 584/584 passed on the branch which submitted this PR, is it failing after you attempt a merge with master?

src/ParseObject.js
```diff
@@ -935,10 +939,22 @@ class ParseObject {
...
   /**
-   * Clears any changes to this object made since the last call to save()
+   * Clears any (or specific) changes to this object made since the last call to save()
+   * @param {String|Array<String>|...String} [keys] - specify which fields to revert
    */
-  revert(): void {
-    this._clearPendingOps();
+  revert(...keys?: Array<string>): void {
+    let _keysToRevert;
+    if(keys.length) {
```
flovilmart: This will throw an exception if keys is undefined. Please address

src/ParseObject.js
```diff
@@ -935,10 +939,22 @@ class ParseObject {
   }
 
   /**
-   * Clears any changes to this object made since the last call to save()
+   * Clears any (or specific) changes to this object made since the last call to save()
+   * @param {String|Array<String>|...String} [keys] - specify which fields to revert
    */
-  revert(): void {
-    this._clearPendingOps();
+  revert(...keys?: Array<string>): void {
```
flovilmart: The type is wrong here, please fix

src/ParseObject.js
```diff
@@ -935,10 +939,22 @@ class ParseObject {
...
+    let _keysToRevert;
```
flovilmart: Please fix variable name, no need for the _

src/ParseObject.js
```diff
@@ -935,10 +939,22 @@ class ParseObject {
...
+    if(keys.length) {
+      _keysToRevert = [];
+      keys.forEach((key) => {
+        if (Array.isArray(key)) {
```
flovilmart: I’m not sure why it would support nested arrays.

src/ParseObject.js
```diff
@@ -935,10 +939,22 @@ class ParseObject {
   }
 
   /**
-   * Clears any changes to this object made since the last call to save()
+   * Clears any (or specific) changes to this object made since the last call to save()
+   * @param {String|Array<String>|...String} [keys] - specify which fields to revert
```
flovilmart: This is very confusing, it’s either an array or a list of arguments, but all at once a makes it hard to reason about and potentially buggy

src/__tests__/ParseObject-test.js
```diff
@@ -845,6 +845,49 @@ describe('ParseObject', () => {
...
+      gender: 'female'
+    });
+    o.set({ cool: true, gender: 'male' });
+    o.increment('count');
+    o.increment('age');
+    expect(o.get('cool')).toBe(true);
+    expect(o.get('count')).toBe(6);
+    expect(o.get('age')).toBe(19);
+    expect(o.get('gender')).toBe('male');
+    o.revert('age', 'count', ['gender']);
```
flovilmart: This is really non sensical to allow such things. It’s not because it’s ‘supported’ that it should be allowed.

src/ParseObject.js
flovilmart: As for now it doesn’t throw an exception but I’m genuinely curious as for why. Can you explain?

src/ParseObject.js
omairvaiyani: Me neither, but I'm following the pattern seen in ```Parse.Query#select|include```, e.g. https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseQuery.js#L1374

flovilmart: The big difference is the type definition of the keys property. It’s not the same in the 2 implementations

src/ParseObject.js
flovilmart: Not at all,  the `?` never sees the light of the Babel transformation as it’s stripped out of the source code first.

The reason is just because ...keys is simply the same as doing `keys = Array.from(arguments)`

omairvaiyani: Ah okay, got it, you're right - I can add an empty check regardless to avoid future confusion

flovilmart: It’s not that, it’s just that calling with not arguments or an empty array will have different meanings and results, as in the latter case the length will be 1 and it may have additional side effects

flovilmart:
LGTM’ thanks!

omairvaiyani:
@flovilmart should be good to go now

PR #570, opened 2018-05-17
Make javascript key optional

### LINKED ISSUE #569: ParseFile.saveFile failed if javascriptKey is not set
As state by [parse-server](https://github.com/parse-community/parse-server/)'s README, `javascriptKey` is no longer necessary but without setting it, using ParseFile will fail.

In this [line](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseFile.js#L246), `X-Parse-JavaScript-Key` will be `undefined`, and when the xhr is finally sent, NodeJS will throw a `ERR_HTTP_INVALID_HEADER_VALUE` [the line here](https://github.com/nodejs/node/blob/master/lib/_http_outgoing.js#L458)

The easy fix should look like this

```js
var jsKey = CoreManager.get('JAVASCRIPT_KEY');
if (jsKey) {
 headers[ 'X-Parse-JavaScript-Key'] = jsKey;
}
```

I'll be glad to send in a PR if the fix is as easy as above.

This is referencing to Parse SDK JS v1.11.1 and NodeJS v9.11.1

alvinthen:
Close #569

### DIFF
--- a/src/ParseFile.js
+++ b/src/ParseFile.js
@@ -243,9 +243,12 @@ var DefaultController = {
     // To directly upload a File, we use a REST-style AJAX request
     var headers = {
       'X-Parse-Application-ID': CoreManager.get('APPLICATION_ID'),
-      'X-Parse-JavaScript-Key': CoreManager.get('JAVASCRIPT_KEY'),
       'Content-Type': source.type || (source.file? source.file.type : null)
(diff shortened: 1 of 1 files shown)


flovilmart:
Thanks for the PR

PR #573, opened 2018-05-22
Add options to full text search

danielbm:
This PR adds the support for options on the full text search, namely: $caseSensitive, $language and $diacriticSensitive.


dplewis:
@flovilmart Can you take a look at travis?

flovilmart:
@dplewis seems that integration tests are failing. Not sure why or if we broke something else.

dplewis:
@flovilmart Thanks for fixing the tests

flovilmart:
NP!

src/ParseQuery.js
```diff
@@ -1005,9 +1005,13 @@ class ParseQuery {
...
   * @param {String} key The key that the string to match is stored in.
   * @param {String} value The string to search
+  * @param {Object} options (Optional) See https://docs.mongodb.com/manual/reference/operator/query/text/#text-query-operator-behavior
+  * @param {String} options.$language
+  * @param {String} options.$caseSensitive
+  * @param {String} options.$diacriticSensitive
   * @return {Parse.Query} Returns the query, so you can chain this call.
   */
- fullText(key: string, value: string): ParseQuery {
+ fullText(key: string, value: string, options: ?Object): ParseQuery {
```
dplewis: Can you add something similar to other functions that use options?

https://github.com/danielbm/Parse-SDK-JS/blob/4ea0c11bd375f0f1c8f942b7308e57ad9ba6ce30/src/ParseQuery.js#L495

danielbm: Done!

dplewis: I mean something like for every option.

```
var fullOptions = {};
    if (options.hasOwnProperty('language')) {
      fullOptions.$language = options.language;
    }
```

Can you also write a test case for $caseSensitive, $diaraticSensitivity, and options.unknown (this might throw an error)?

danielbm: Do you think this code would do the trick?

```javascript
    options = options || {};

    var fullOptions = {};
    for (var key in options) {
      switch (key) {
        case '$language':
          fullOptions.$language = options[key];
          break;
        case '$caseSensitive':
          fullOptions.$caseSensitive = options[key];
          break;
        case '$diacriticSensitive':
          fullOptions.$diacriticSensitive = options[key];
          break;
        default:
          throw new Error('Unknown option: ' + key);
          break;
      }
    }
```

dplewis: It should do the trick but seems verbose. Also I don't think we should have the `$` in the options getting passed in. $ is an internal thing. What do you think?

dplewis: @danielbm Can you write the test cases first?

danielbm: It seemed a bit verbose to me too, but doing the validation for the three options separately, plus the validation for unknown options was also. As per the ```$```, I totally agree. I'll remove it.

dplewis: Sweet!

danielbm: @dplewis I just pushed the code with the options validations, more test cases.

dplewis:
@flovilmart is there an easy way to view generated docs for a PR without pulling it down?

flovilmart:
Not really ;/ you’ll have to pull it

dplewis:
@flovilmart I think the CONTRIBUTION.md looks outdated. Doesn’t say how to run those unit jest test, integration test, generate docs, etc.

flovilmart:
@dplewis you're right ;) we need to do somethings about that

src/ParseQuery.js
```diff
@@ -1005,21 +1005,49 @@ class ParseQuery {
...
+        case 'diacriticSensitive':
+          fullOptions.$diacriticSensitive = options[key];
+          break;
+        default:
+          throw new Error('Unknown option: '+key);
+          break;
+      }
+    }
+
+    if (options && options.hasOwnProperty('$diacriticSensitive')) {
```
dplewis: Remove this as it makes all your test fail. Its basically saying `$diacriticSensitive` is required

Were you able to run test locally.

src/ParseQuery.js
```diff
@@ -993,7 +993,7 @@ class ParseQuery {
   *
   * In order to sort you must use select and ascending ($score is required)
   *  <pre>
-  *   query.fullText('term');
+  *   query.fullText('field', 'term', { $language: "spanish" });
```
dplewis: For documentation purposes leave an example of with and w/o options.

Docs are auto generated by using `npm run docs`. Use it to check your formatting.

`$language` is now `language`

src/ParseQuery.js
```diff
@@ -1005,21 +1005,49 @@ class ParseQuery {
   *
   * @param {String} key The key that the string to match is stored in.
   * @param {String} value The string to search
+  * @param {Object} options (Optional) See https://docs.mongodb.com/manual/reference/operator/query/text/#text-query-operator-behavior
+  * @param {String} options.language
+  * @param {String} options.caseSensitive
+  * @param {String} options.diacriticSensitive
```
dplewis: Diacritic and case are Boolean type. What happens if they aren't boolean 🤔?

Instead of the mongo docs, put a summary next to the options for readability

src/ParseQuery.js
```diff
@@ -1005,21 +1005,49 @@ class ParseQuery {
...
-   if (!value) {
-     throw new Error('A search term is required');
-   }
-   if (typeof value !== 'string') {
-     throw new Error('The value being searched for must be a string.');
-   }
-
-   return this._addCondition(key, '$text', { $search: { $term: value } });
- }
+  fullText(key: string, value: string, options: ?Object): ParseQuery {
```
dplewis: Looks like this is indented improperly

integration/test/ParseQueryTest.js
```diff
@@ -1537,4 +1537,110 @@ describe('Parse Query', () => {
...
+      objects.push(obj);
+    }
+    Parse.Object.saveAll(objects).then(() => {
+      const q = new Parse.Query(TestObject);
+      q.fullText('comment', 'preparar', { language: "portuguese" });
+      return q.find();
+    }).then((results) => {
+      assert.equal(results.length, 4);
+    }).then(null, () => {
+      done();
```
dplewis: `done()` should be in the `then()` not the error.

Your test are suppose to pass but throwing an error. 

Check all your tests.

integration/test/ParseQueryTest.js
```diff
@@ -1537,4 +1537,110 @@ describe('Parse Query', () => {
...
+    });
+  });
+
+  it('can perform a full text search with case sensistive options', (done) => {
```
dplewis: *case sensitive

integration/test/ParseQueryTest.js
```diff
@@ -1537,4 +1537,110 @@ describe('Parse Query', () => {
...
+      objects.push(obj);
+    }
+    Parse.Object.saveAll(objects).then(() => {
+      const q = new Parse.Query(TestObject);
+      q.fullText('comment', 'Preparar', { language: "portuguese", caseSensistive: true });
+      return q.find();
+    }).then((results) => {
+      assert.equal(results.length, 1);
+    }).then(null, () => {
+      done();
```
dplewis: should not throw an error

integration/test/ParseQueryTest.js
```diff
@@ -1537,4 +1537,110 @@ describe('Parse Query', () => {
...
+      objects.push(obj);
+    }
+    Parse.Object.saveAll(objects).then(() => {
+      const q = new Parse.Query(TestObject);
+      q.fullText('comment', 'cafe', { language: "portuguese", diacriticSensitive: true });
+      return q.find();
+    }).then((results) => {
+      assert.equal(results.length, 1);
+    }).then(null, () => {
+      done();
```
dplewis: should not throw an error

src/ParseQuery.js
danielbm: oops =|

src/ParseQuery.js
danielbm: Like this?

```
/**
  * Adds a constraint for finding string values that contain a provided
  * string. This may be slow for large datasets. Requires Parse-Server > 2.5.0
  *
  * In order to sort you must use select and ascending ($score is required)
  *  <pre>
  *   query.fullText('field', 'term');
  *   query.ascending('$score');
  *   query.select('$score');
  *  </pre>
  *
  * To retrieve the weight / rank
  *  <pre>
  *   object->get('score');
  *  </pre>
  *
  * You can define optionals by providing an object as a third parameter
  *  <pre>
  *   query.fullText('field', 'term', { language: "spanish", diacriticSensitive: true });
  *  </pre>
  *
  * @param {String} key The key that the string to match is stored in.
  * @param {String} value The string to search
  * @param {Object} options (Optional) See https://docs.mongodb.com/manual/reference/operator/query/text/#text
  * @param {String} options.language The language that determines the list of stop words for the search and the rules for the stemmer and tokenizer.
  * @param {Boolean} options.caseSensitive A boolean flag to enable or disable case sensitive search.
  * @param {Boolean} options.diacriticSensitive A boolean flag to enable or disable diacritic sensitive search.
  * @return {Parse.Query} Returns the query, so you can chain this call.
  */
```

src/ParseQuery.js
danielbm: Actually it was indented improperly before, and I fixed it.

integration/test/ParseQueryTest.js
```diff
@@ -1537,4 +1537,110 @@ describe('Parse Query', () => {
...
+    for (const i in subjects) {
+      const obj = new TestObject({ comment: subjects[i] });
+      objects.push(obj);
+    }
+    Parse.Object.saveAll(objects).then(() => {
+      const q = new Parse.Query(TestObject);
+      q.fullText('comment', 'preparar', { language: "portuguese" });
+      return q.find();
+    }).then((results) => {
+      assert.equal(results.length, 4);
```
dplewis: Can you write this like? It kinda confusing in the tests.

```
 }).then((results) => {
  assert.equal(results.length, 4);
  done();
});
```

integration/test/ParseQueryTest.js
```diff
@@ -1537,4 +1537,110 @@ describe('Parse Query', () => {
...
+      'preparação de cafe com leite',
+    ];
+    const objects = [];
+    for (const i in subjects) {
+      const obj = new TestObject({ comment: subjects[i] });
+      objects.push(obj);
+    }
+    Parse.Object.saveAll(objects).then(() => {
+      const q = new Parse.Query(TestObject);
+      q.fullText('comment', 'Preparar', { language: "portuguese", caseSensistive: true });
```
dplewis: I don't know how this test is passing when `caseSensistive` is spelled wrong.

dplewis:
@danielbm I added a few more comments on your test cases

cleever:
Any news on this?

danielbm:
@cleever Not really. I'm just short of time to fix this PR.

dplewis:
@danielbm I can finish off this PR for you, if thats cool.

danielbm:
@dplewis Sure! Feel free to do it!

flovilmart:
@dplewis where are we at on this one?

dplewis:
@danielbm how does this look?

danielbm:
@dplewis @flovilmart thanks!

flovilmart:
Thanks @danielbm for getting it started!

PR #579, opened 2018-05-28
Throw an error if there are not `rejected` handlers

alexander-alzate:
It is quite annoying trying to figure out what is the problem when you don't get any errors just because you don't properly define a `rejected` handler.

Using chained promises is also a problem because you have to define a catch for any `ParsePromise` returned.


alexander-alzate:
This get fixed on v2 by using native `Promise`

PR #582, opened 2018-06-12
Add nearsphere convenience methods

mtrezza:
Adding convenience methods for queries with `geoWithin.nearSphere`.

Depending on https://github.com/parse-community/parse-server/pull/4825

### DIFF
--- a/src/__tests__/ParseQuery-test.js
+++ b/src/__tests__/ParseQuery-test.js
@@ -573,7 +573,7 @@ describe('ParseQuery', () => {
 
   it('can generate near-geopoint queries with ranges', () => {
     var q = new ParseQuery('Shipment');
-    q.withinRadians('shippedTo', [20, 40], 2);
+    q.withinRadians('shippedTo', [20, 40], 2, true);
     expect(q.toJSON()).toEqual({
       where: {
         shippedTo: {
@@ -587,7 +587,7 @@ describe('ParseQuery', () => {
       }
     });
 
-    q.withinMiles('shippedTo', [20, 30], 3958.8);
+    q.withinMiles('shippedTo', [20, 30], 3958.8, true);
     expect(q.toJSON()).toEqual({
       where: {
         shippedTo: {
@@ -601,7 +601,7 @@ describe('ParseQuery', () => {
       }
     });
 
-    q.withinKilometers('shippedTo', [30, 30], 6371.0);
+    q.withinKilometers('shippedTo', [30, 30], 6371.0, true);
     expect(q.toJSON()).toEqual({
       where: {
         shippedTo: {
@@ -616,19 +616,15 @@ describe('ParseQuery', () => {
     });
   });
 
-  fit('can generate near-geopoint queries without sorting', () => {
+  it('can generate near-geopoint queries without sorting', () => {
     var q = new ParseQuery('Shipment');
-    q.withinRadiansUnsorted('shippedTo', [20, 40], 2);
+    q.withinRadians('shippedTo', new ParseGeoPoint(20, 40), 2, false);
     expect(q.toJSON()).toEqual({
       where: {
         shippedTo: {
           $geoWithin: {
             $centerSphere: [
-              {
-                __type: 'GeoPoint', 
-                latitude: 20,
-                longitude: 40
-              },
+              [40, 20], // This takes [lng, lat] vs. ParseGeoPoint [lat, lng].
               2
             ]
           }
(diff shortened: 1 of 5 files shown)


src/__tests__/ParseQuery-test.js
```diff
@@ -616,6 +616,63 @@ describe('ParseQuery', () => {
     });
   });
 
+  fit('can generate near-geopoint queries without sorting', () => {
```
flovilmart: Oops, needs to replace by ‘it’

mtrezza:
@flovilmart good idea, introduced `sorted` parameter.
Should I make the `sorted` parameter nullable and default to `true` so that existing code does not break?

src/__tests__/ParseQuery-test.js
mtrezza: done

flovilmart:
Ok, tests are failing for now as they are run against an unpublished version of parse-server

flovilmart:
I’ll do a release of the server later so we can get going with this one.

flovilmart:
@mtrezza I’m wondering, can we make those tests succeed without the new parse-server? This way, we’ll be able to ship in a single server release with the JS SDK?

mtrezza:
@flovilmart Tests against the generated query syntax in `src/__tests__/ParseQuery-test.js` pass without server update. Tests against the query results in `integration/test/ParseGeoPointTest.js` would require the server update AFAIK.

I think the actual query result tests are an important QA step so I suggest updating the server first. The update brings the full functionality with `withJSON` anyway, this PR is just adding convenience methods.

mtrezza:
@flovilmart Yes, we could remove the tests against the query results (for now) in `integration/test/ParseGeoPointTest.js` and only keep the tests against the generated query syntax in `src/__tests__/ParseQuery-test.js`.

The query results are already tested in https://github.com/parse-community/parse-server/pull/4825 so if the query syntax is correct and passes the tests, it should yield the same query results. The integration tests behave a bit like chicken-and-egg it seems.

However, I think there should also be a test against query results of the convenience methods, so I'd add them ASAP later on. What do you suggest?

flovilmart:
We can do the integrations tests in parse server instead, for this one or mark them xit or pending / optional

mtrezza:
Yes, I already added the tests in parse server with https://github.com/parse-community/parse-server/pull/4827.

mtrezza:
Yes, I already added the tests in parse server with https://github.com/parse-community/parse-server/pull/4827.

mtrezza:
@flovilmart done

dplewis:
> I’m wondering, can we make those tests succeed without the new parse-server? This way, we’ll be able to ship in a single server release with the JS SDK?

@flovilmart Can we do something similar to what we did with the PHP SDK and have it always go against the latest?

dplewis:
@flovilmart Have you seen my last comment https://github.com/parse-community/Parse-SDK-JS/pull/582#issuecomment-397150539 

What do you think about using the latest parse server for testing?

flovilmart:
We can definitely do that :) easier to catch regessions!

mtrezza:
What do I have to change in code?

dplewis:
@mtrezza You can start by changing the version of parse server used for test to the latest.

https://github.com/parse-community/Parse-SDK-JS/blob/master/integration/package.json#L6

to `github:parse-community/parse-server#latest`

Then remove your `xit` from test

Also I'm confused by change to the existing test cases if `sorted` is optional why are you forcing true? I feel like I'll have to change all my calls to sorted = true.

mtrezza:
Change to the existing test cases was to make it apparent that there is a sorted parameter and to explicitly test with sorted=true. But it would surely be enough to have one case with sorted=true. So I’ll revert the existing test cases and add one for sorted=true.

dplewis:
> So I’ll revert the existing test cases and add one for sorted=true.

That’s what I was going for. It looked like it was required

mtrezza:
@dplewis Can you please have a look at the failing integration test, I think it is because parse-server#latest failed to build?

dplewis:
@flovilmart According to the docs 
https://github.com/parse-community/parse-server#want-to-ride-the-bleeding-edge

I have to do `npm install parseplatform/parse-server.git#latest`

It currently fails for me 

```
npm ERR! prepareGitDep > babel src/ -d lib/ --copy-files
npm ERR! prepareGitDep 
npm ERR! prepareGitDep 
npm ERR! prepareGitDep 2> npm WARN install Usage of the `--dev` option is deprecated. Use `--only=dev` instead.
npm ERR! prepareGitDep src/ doesn't exist
```

`src/` folder doesn't exist in the latest branch

using `npm install parseplatform/parse-server.git#master` works

should the docs be updated or am I missing something

dplewis:

On a side note should we pin the node packages? Don't want a repeat of https://github.com/parse-community/parse-server/issues/2040

flovilmart:
@dplewis see my comment. I’ll have a look tomorrow too

flovilmart:
@dplewis that sounds good! Thanks to this new NPM feature, one can use any branch and test against it. npm will fetch the dev deps and do it's thing! Wow :)

dplewis:
@flovilmart thats pretty awesome now that I think about it.

flovilmart:
@mtrezza Thanks again for the PR!

mtrezza:
@flovilmart Sure thing, thanks to you and dplewis for the support.

flovilmart:
Any time!

Issue #592, opened 2018-07-04
GeoPoint storing as strings if constructed by a string array


Krilwya:
I encountered an issue using near after migrating string gps positions into GeoPoints.

So basically when you do something like : 
new Parse.GeoPoint(["48.84023", "2.33334"])
it returns :
ParseGeoPoint {_latitude: "48.84023", _longitude: "2.33334"}

So it passes through the constuctor of the Geopoint, it shall be more efficient in the _validate method of the Geopoint to check typeof latitude == "number" and same with longitude or try to cast those into parseFloat or whatever.

Knowing all of that, having string geopoints break all near, withinKilometers, and so on functions as mongodb would look for floats.

flovilmart:
@Krilwya are you willing to open a PR with the failing test and add additional validation?

Krilwya:
Sure, I runned the jest test on my changes locally, that succeed them, if you could explain to me the procedure to ask for a pull request as it's my first time on github that'd be great. I am only used to vsts.

flovilmart:
Did you check the github pull request guide? https://help.github.com/articles/creating-a-pull-request/

Krilwya:
Thank you for your answer, I think I did all the requirement to do so. If you need anything else to verify or something seems wrong I'd be pleased to hear from you.

dplewis:
Closing via https://github.com/parse-community/Parse-SDK-JS/pull/671

PR #598, opened 2018-07-09
Add casting to integer on parse geopoint constructor. More flexibilit…

Krilwya:
Adding a support to construct geopoint from string and correct the issue I opened #592
Now it rightly create a geopoint with valid float numbers in order for the queries near, withinKilometers, etc.. works fine.

you can construct GeoPoint with string as followed : 
new GeoPoint("30", "30");
new GeoPoint(["30", "30"]);
new GeoPoint({latitude: "30", longitude: "30"});

all those equals the following number constructed geopoint ; 
new GeoPoint(30, 30);

It also add more flexibility if you retrieve your gps data from an existing db or a REST/API.


dplewis:
@flovilmart I think we have a flaky test here. Can you take a look?

dplewis:
@flovilmart Is this good?

flovilmart:
I'm not sure about this one, this looks like a total mess of possible contructors, while it can be solved externally with helpers.

flovilmart:
Closing at it clutters the codebase and it can be achieved with your own helper functions if really needed.

PR #600, opened 2018-07-11
Fix for the formatting of the ParseFile comment that has

acinader:
Currently, for me, the [ParseFile](http://parseplatform.org/Parse-SDK-JS/api/v1.11.1/Parse.File.html) Constructor API docs has a <pre> tag that is not rendering:
screen shot 2018-07-11 at 4 17 20 pm

this fix, while not ideal, is a good improvement.
screen shot 2018-07-11 at 4 14 00 pm

I'm game to try someother approaches?


flovilmart:
@acinader that looks great! Thanks!

PR #601, opened 2018-07-11
Fix documentation for include and select

MatrixFrog:
These both can accept either a string, or an array of strings

### DIFF
--- a/src/ParseQuery.js
+++ b/src/ParseQuery.js
@@ -1344,10 +1344,10 @@ class ParseQuery {
   /**
    * Includes nested Parse.Objects for the provided key.  You can use dot
    * notation to specify which fields in the included object are also fetched.
-   * @param {String} key The name of the key to include.
+   * @param {...String|Array<String>} key The name(s) of the key(s) to include.
    * @return {Parse.Query} Returns the query, so you can chain this call.
    */
-  include(...keys: Array<string>): ParseQuery {
+  include(...keys: Array<string|Array<string>>): ParseQuery {
     keys.forEach((key) => {
       if (Array.isArray(key)) {
         this._include = this._include.concat(key);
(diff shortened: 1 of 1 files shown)


MatrixFrog:
Ah, thanks, I totally missed that you're using both type annotations in comments, and TS annotations

MatrixFrog:
I'm not sure what you want it to say. The way it's written, you could pass multiple arrays of strings, like `include(['foo'], ['foo.bar', 'bar'])` though I don't know if you want to document that as being supported.

I was following the docs at https://www.typescriptlang.org/docs/handbook/functions.html#rest-parameters which seem to imply that the type of a `...` param is always an array. But I'm not sure if you're actually using TypeScript or Flow, and it's possible they disagree slightly on this.

flovilmart:
Right, that’s this part that’s confusing `...String|Array<string>` as it should be read:

(spread of (string or arrays of strings) ) instead of ((spread of string) or (array of strings))

In doubt that’s confusing and even if it’s supported it’s rather make the documentation unambiguous even if it means sacrificing a ‘feature’ or corner case. What do you think?

In the end, it supports only arrays, which can be strings of arrays of strings. So we should remove the spread operator from the jsdocs as it’s confusing (me) an perhaps others.

What do you think?

MatrixFrog:
Okay so you can basically call it in two ways: You can call it as though the `@param` annotation in the JSDoc was `{...String}` or you can just pass a single array as though it was `{Array<String>}`. (Other ways of calling it are possible but rare enough that we don't want to document them.)

I'm not sure how to represent that in the JSDoc. Maybe just what I have in the PR currently, except with parens: `{(...String)|Array<string>}`

flovilmart:
The parens may help with the readability.

MatrixFrog:
Thanks!

Issue #77, opened 2015-11-02
Any plans to add Local Datastore for JS SDK?


h4rm:
Are there any plans to add the local datastore to the Javascript SDK just like it happened in the Android/iOS SDK?

While I can save `object.toJSON()` to the localStorage, the JSON serialization does not recursively save all the other objects that were included in the query but renders out pointers instead.

andrewimm:
We're actively exploring methods of local storage, so that apps built on platforms like React Native can perform queries when offline. The biggest questions at this point are around how the storage would be interfaced, and whether it would live in the SDK or an addon. For interfaces, the current frontrunner is a Flux-like store based on the one from Parse+React, which fits well in to the new SDK's internal storage scheme. As for where it lives -- the new Controller-based architecture of the SDK makes it really easy to replace individual pieces, and provide functionality as external libraries that are loaded at runtime.

reidmweber:
:+1: We have an iOS app that uses parse and the local datastore.  We would like to use parse for a desktop app written with Electron. Ideally, we could use the same methods (pinInBackground, .fromLocalDatastore, etc). Would that work if it was an add-on?

wookiem:
Any update on availability of a local datastore for React Native?  

Also, are you working on any strategies to make query subscriptions more React-like?  Unless I'm mistaken, the only way to refresh on-screen data is to re-query the datastore periodically, without any knowledge of whether the underlying data has changed.  Thanks!

carbureted:
+1 on any update regarding a React Native local datastore. Trying to choose between hacking something together on SQLite or integrating parse with the hope that storage is eventually figured out. 

Even a negative timeframe like "Not in the next 3 months" would be super helpful.

yamill:
+1 I think this is much needed - especially for React Native apps.

5amfung:
Agreed with @carbureted.  If local storage is not the priority, please let us know so that we could make a sound decision of what to do.

andrewimm:
It's not a priority at the moment, because it would take some significant restructuring to provide any valuable querying over local objects. If we were to implement a LocalDatastore similar to the one seen on Android / iOS, it would probably be a redesign version that avoids some of the less performant aspects of pinning. Any API we would implement would also want to be something that works the same way on both React Native and and offline ServiceWorker web apps, so there's a lot of thinking to do here before any action is taken.

ssssssssssss:
[...]

PR #612, opened 2018-07-27
Local Datastore

dplewis:
This is based on discussion https://github.com/parse-community/Parse-SDK-JS/issues/77

This is a Work In Progress.


dplewis:
@flovilmart Can you check Travis for me? I've been having gulp issues. 

Also can you do a quick run though of what I've done so far? Its pretty close to being done.

flovilmart:
Error: Cannot find module 'utf-8-validate' from '/home/travis/build/parse-community/Parse-SDK-JS/node_modules/parse/node_modules/ws/lib'


It’s the error that I see before the integration tests, during the browserify build

dplewis:
@flovilmart I got both 

Error: Cannot find module 'utf-8-validate' from 'xxx/node_modules/ws/lib'
Error: Cannot find module 'bufferutil' from 'xxx/node_modules/ws/lib'

I tried to npm install --save them but ended up with another error

flovilmart:
Uhm, that’s odd, try reverting ws to an earlier version then :)

src/ParseQuery.js
```diff
@@ -16,6 +16,8 @@ import ParseGeoPoint from './ParseGeoPoint';
 import ParsePolygon from './ParsePolygon';
 import ParseObject from './ParseObject';
 import ParsePromise from './ParsePromise';
+import LocalDatastore from './LocalDatastore';
+import { matchesQuery } from 'parse-server/lib/LiveQuery/QueryTools';
```
dplewis: @flovilmart I think this like is my problem. I want to use this file. I get the following error.

```
GulpUglifyError: unable to minify JavaScript
Caused by: SyntaxError: Unexpected token: keyword (const)
File: /Users/Squad/Github/Parse-SDK-JS/dist/parse.js
Line: 16699
```

Looks like that file doesn't go through babel.

Should I just copy the files over or do you know of a better way.

dplewis:
@flovilmart I fixed the issue. It was a lot of circular dependency. Since localStorage isn't available I'll have to mock it along with some other tests. 

Can you have a quick look through?

flovilmart:
I’ll have a look. The diff hit for coverage is only 21.11% can we do something?

dplewis:
@flovilmart I'll try

src/LocalDatastore.js
```diff
@@ -41,29 +40,30 @@ const LocalDatastore = {
     controller.clear();
   },
 
-  _handlePinWithName(name: string, object: ParseObject) {
+  _handlePinWithName(name: string, object: any) {
```
flovilmart: Why loose the typing?

dplewis: I kept running into a 
```
Uncaught TypeError: Super expression must either be null or a function, not undefined
```
Whenever I used ParseObject in the tests. I found a fix for it so I'll add the typing back.

flovilmart: You are 💯

flovilmart:
I’ll review over the weekend

integration/test/ParseQueryTest.js
```diff
@@ -1661,4 +1663,72 @@ describe('Parse Query', () => {
...
+        query.fromPin();
+        return query.find();
+      }).then((results) => {
+        assert.equal(results.length, 3);
+        done();
+      });
+    });
+  });
+
+  describe('Parse Query Pinning (LocalStorage)', () => {
```
dplewis: @flovilmart Quick question? Since I have both in-memory and localStorage available. I want to run the test cases against both without copying them. Would putting describe inside of a for-loop make sense?

flovilmart: Yep it works, otherwise what you can do is put it into a functions and call the function from the specific describe

dplewis: I just tried putting it in a loop. It also works great with Jasmine Test Explorer too

dplewis:
@flovilmart The OfflineQuery pretty much has the same support liveQuery does. I can add more integration tests for it if needed.

I also have add that if a PinName doesn't have any objects attached then delete that PinName

flovilmart:
Perhaps we could have an offline query engine written in JS that we use in both projects? Perhaps too much trouble for now. 

Any limitation on the queries we can run? Are they properly failing so at development time one knows which queries can / cannot run locally?

flovilmart:
I’ll do the full review tomorrow. Have a good one!

src/LocalDatastore.js
```diff
@@ -0,0 +1,198 @@
...
+    return controller.unPinWithName(name);
+  },
+
+  _getLocalDatastore() {
+    const controller = CoreManager.getLocalDatastoreController();
+    return controller.getLocalDatastore();
+  },
+
+  _clear(): void {
+    var controller = CoreManager.getLocalDatastoreController();
```
flovilmart: prefer const

src/ParseObject.js
```diff
@@ -1614,10 +1666,84 @@ class ParseObject {
...
+  }
+
+  /**
+   * Removes all objects in the local datastore using a default pin name: _default.
+   *
+   * @static
+   */
+  static unPinAllObjects() {
+    const localDatastore = CoreManager.getLocalDatastore();
+    localDatastore.unPinWithName(localDatastore.DEFAULT_PIN);
```
flovilmart: we should hasve a method

```
getPinName(pinName: String?): string {
   if (!pinName) { return DEFAULT_PIN }
   return PIN_PREFIX + name;
}
```

This qwould void the probvlems of concatenating localDatastore.PIN_PREFIX + name or getting default :)

src/LocalDatastore.js
dplewis: I added 'no-var' to eslint rule and there are quite a few there. Should I fix them in this PR?

flovilmart: Another one probably :(

flovilmart: @dplewis please, in another PR for the linter fixes, now it's a mess of changes ;)

.eslintrc.json
```diff
@@ -22,6 +22,7 @@
         "no-multiple-empty-lines": 1,
         "prefer-const": "error",
         "space-infix-ops": "error",
-        "no-useless-escape": "off"
+        "no-useless-escape": "off",
+        "no-var": "error"
```
flovilmart: @dplewis the `no-var`  should go into another PR. The local datastore is a big enough PR like that. I'd rather get the linting done with `no-var` done at a later time. Here it pollutes the PR and the feature. So please put this commit into another branch for now.

dplewis: I got it.

dplewis:
Those changes have been reverted and requested changes made

integration/test/ParseObjectTest.js
```diff
@@ -1456,4 +1463,611 @@ describe('Parse Object', () => {
...
+      });
+
+      it(`${controller.name} cannot pin unsaved pointer`, () => {
+        try {
+          const object = new TestObject();
+          const pointer = new Item();
+          object.set('child', pointer);
+          object.pin();
+        } catch (e) {
+          assert.equal(e.message, 'Cannot create a pointer to an unsaved ParseObject');
```
flovilmart: 👍

integration/test/ParseObjectTest.js
```diff
@@ -1456,4 +1463,611 @@ describe('Parse Object', () => {
...
+        assert(localDatastore[obj1.id]);
+        assert(localDatastore[obj2.id]);
+
+        await obj1.destroy();
+
+        localDatastore = Parse.LocalDatastore._getLocalDatastore();
+        assert(Object.keys(localDatastore).length === 3);
+        assert.deepEqual(localDatastore[DEFAULT_PIN], [obj2.id]);
+        assert.deepEqual(localDatastore[PIN_PREFIX + 'test_pin'], [obj2.id]);
+        assert(localDatastore[obj2.id]);
```
flovilmart: I would probably have used another key for the index of the objects like `${className}_${id}`

integration/test/ParseQueryTest.js
```diff
@@ -1801,4 +1805,69 @@ describe('Parse Query', () => {
...
+      });
+
+      it(`${controller.name} can query from local datastore`, async () => {
+        const obj1 = new TestObject({ field: 1 });
+        const obj2 = new TestObject({ field: 2 });
+        const obj3 = new TestObject({ field: 3 });
+        const objects = [obj1, obj2, obj3];
+        await Parse.Object.saveAll(objects);
+
+        Parse.Object.pinAll(objects);
```
flovilmart: when you call `pinAll` does it replaces the content of the default pin or does it add elements to the existing pinned elements?

src/LocalDatastore.js
```diff
@@ -0,0 +1,210 @@
...
+      console.log('Parse.enableLocalDatastore() must be called first'); // eslint-disable-line no-console
+    }
+    return this.isEnabled;
+  }
+};
+
+function isLocalStorageEnabled() {
+  const item = 'parse_is_localstorage_enabled';
+  try {
+    localStorage.setItem(item, item);
```
flovilmart: looks harsh as a test method no? wouldn't testing `typeof localStorage.setItem` and `localStorage.removeItem`
be enough?

src/LocalDatastoreController.default.js
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @flow
+ */
+
+const memMap = {};
+const LocalDatastoreController = {
```
flovilmart: what should be the point of this local datastore? I'm not sure it has any used as on any page refresh it would be empty.

src/LocalDatastoreController.localStorage.js
```diff
@@ -0,0 +1,52 @@
...
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @flow
+ */
+
+/* global localStorage */
+
+const LocalDatastoreController = {
+  fromPinWithName(name: string): ?any {
+    const values = localStorage.getItem(name);
```
flovilmart: Shouldn't we use AsyncStorage also, because that's how react native provides it. The localDataStoreCOntroller methods should all be async, this way we can swap LocalStorage or AsyncStorage. Otherwise we'll face issues with RN right away.

If we don't implement the AsyncStorage bridge now, we should still make all those methods async as we know for sure we'll need it in the future. 

Unfortunately, we have to use promises so we don't require babel-regenrator to be integrated in the front end SDK.

src/LocalDatastoreController.localStorage.js
```diff
@@ -0,0 +1,52 @@
...
+    const objects = JSON.parse(values);
+    return objects;
+  },
+
+  pinWithName(name: string, value: any) {
+    try {
+      const values = JSON.stringify(value);
+      localStorage.setItem(name, values);
+    } catch (e) {
+      // Quota exceeded, possibly due to Safari Private Browsing mode
```
flovilmart: we should log something still.

Thie is important for the dev to know at dev time that the local storage exploded for some reason :)

src/LocalDatastoreController.localStorage.js
```diff
@@ -0,0 +1,52 @@
...
+    }
+  },
+
+  unPinWithName(name: string) {
+    localStorage.removeItem(name);
+  },
+
+  getLocalDatastore() {
```
flovilmart: it's a copy / dump method? That seems to return a LocalDataStore :)

src/ParseObject.js
```diff
@@ -1148,6 +1150,58 @@ class ParseObject {
     );
   }
 
+  /**
+   * Stores the object and every object it points to in the local datastore,
+   * recursively, using a default pin name: _default.
+   */
+  pin() {
+    const localDatastore = CoreManager.getLocalDatastore();
+    ParseObject.pinAllWithName(localDatastore.DEFAULT_PIN, [this]);
```
flovilmart: the DEFAULT_PIN should be static somehow, it's a constant, so imported, not a property on a particular instance.

src/ParseObject.js
```diff
@@ -1614,10 +1668,96 @@ class ParseObject {
...
   }
+
+  /**
+   * Stores the objects and every object they point to in the local datastore,
+   * recursively, using a default pin name: _default.
+   *
+   * @param {Array} objects A list of <code>Parse.Object</code>.
+   * @static
+   */
+  static pinAll(objects: Array<ParseObject>) {
```
flovilmart: as mentioned before, this all should be promised based (as ultimately async)

integration/test/ParseQueryTest.js
dplewis: replaces

src/LocalDatastoreController.default.js
dplewis: There are some instances when localStorage isn't available. 

When I was making this localStorage was always undefined unless I mocked it.

I wanted to have some sort of backup. Might be used on in cloud code on a server somewhere.

src/LocalDatastoreController.localStorage.js
dplewis: What text should be logged? There are a lot of potential reasons why this would fail

src/LocalDatastoreController.localStorage.js
dplewis: Yes easier for testing when you can see whats in there

flovilmart: If for testing only, rename _dump() or getAllContents()

src/LocalDatastoreController.localStorage.js
flovilmart: The error itself

integration/test/ParseObjectTest.js
dplewis: I like this, it will be easier to find objects

src/LocalDatastoreController.default.js
flovilmart: Right, that kinda sucks... https://developer.mozilla.org/en-US/docs/Web/API/Storage/LocalStorage

on the web it's been available since forever, so you're safe. 

so:

- browser: we're safe
- node: we probably want to use this in memory one or an LRU.
- react-native: we want async storage.

Because we want a different module based on the platform, we should use compilation flags to determine the proper one (through the usage of process.env.PARSE_BUILD) (like in the rest controller when it picks the right XMLHTTPRequest implementation).

dplewis:
@flovilmart Thanks for the feedback. I'm going to be busy for a long while. If you want to jump in on this PR feel free to clean it up.

flovilmart:
@dplewis I'll close the PR then, pending we have time to finish up the implementation :) Thanks 🙏  again a 1000 times for getting this off the ground!

dplewis:
@flovilmart I forgot I started this PR because I needed it for something

dplewis:
@flovilmart Merging https://github.com/parse-community/parse-server/pull/4933 is messing with the tests. Can you look into it?

dplewis:
@flovilmart I've added some documentation there are a few `TODO` questions I had.

I'll update the javascript guide. Should I add the same documentation to the API Reference like the iOS SDK does?

[iOS Api Pin](http://parseplatform.org/Parse-SDK-iOS-OSX/api/Classes/PFObject.html#/c:objc(cs)PFObject(im)pin)

flovilmart:
The API reference will be generated from the docs. If you run ‘npm run docs’ you’ll have the api reference generate in the ‘out’ folder

dplewis:
I mean adding something like this to this PR for every pin function?

```
Synchronously stores the object and every object it points to in the local datastore, recursively, using a default pin name: PFObjectDefaultPin.

If those other objects have not been fetched from Parse, they will not be stored. However, if they have changed data, all the changes will be retained. To get the objects back later, you can use a PFQuery that uses PFQuery.-fromLocalDatastore, or you can create an unfetched pointer with +objectWithoutDataWithClassName:objectId: and then call -fetchFromLocalDatastore on it.
```

flovilmart:
Uhm, would be nice, but it’s not synchronous 😬 :)

dplewis:
I just copied that from the iOS api reference.

dplewis:
@flovilmart I basically copied the ParseQueryTest.js and ran it against the LocalDatastore, I got them all to run except one. The only limitation I see is regex. Mongo regex and Javascript RegExp have different modifiers except (i) case insensitive and (m) multi lines.

LiveQuery is missing $and, $nor, $containedBy, $geoWithin, $geoIntercepts. I'll add them here first then add them to the Live Queries later.

I'm running into a weird issue with the tests, I want you to look into it.

flovilmart:
> LiveQuery is missing $and, $nor, $containedBy, $geoWithin, $geoIntercepts. I'll add them here first then add them to the Live Queries later.

We perhaps need to consolidate the offlineQuery first, perhaps we could use the one from the JS SDK in liveQuery after :)

For regex support, leave it out for now offline. 

Good Job!

Whst’s The issue you have?

dplewis:
> For regex support, leave it out for now offline.

Regex works it just limited support for modifiers.

> What’s The issue you have?

This line here. So I'm running the tests against three adapters. I can run them individually or two at a time but not all three at once. I end up with an `invalid session token` error for every test.

https://github.com/parse-community/Parse-SDK-JS/pull/612/files#diff-f42538d45a95bd688e048debdb1ec1c5R28

flovilmart:
@dplewis I'll see if i see something to do with it

flovilmart:
@dplewis updated the testfile, all pass on me side with the change

dplewis:
Sweet cleanup

flovilmart:
Patch and project diff are very low

dplewis:
I know. After everything in the todo list. I'll go back over it and clean it up.

flovilmart:
@dplewis thanks! I'll have a look!

dplewis:
@acinader I opened a PR in the docs repo https://github.com/parse-community/docs/pull/558

acinader:
man, that's seriously helpful. I have barely touched the IOS SDK, so this makes more sense now.

flovilmart:
Where do we stand on this one?

dplewis:
It should be good to go!

flovilmart:
Is it completely opt-in? And anyone not wanting the feature should be able to experience no regression?

dplewis:
Yes you have to enable it with `Parse.enableLocalDatastore()`.

GoMino:
Hello guys does updating an object while being offline, get sync automatically when network access reappear with this new feature (react-native usage) ?

Can't wait to test the localstorage feature with the upcoming [2.2.0 release](https://github.com/parse-community/Parse-SDK-JS/pull/733) 🤩

dplewis:
It currently doesn't exist. Thats something we can implement in the future. I want to play around with this in production for a bit. Feel free to open a PR.

GoMino:
Ok so if I understand correctly this version smoothly enable offline read access to already sync datas. 
I will let you know if I have time to add support for write access.

GoMino:
@dplewis is the following [doc regarding localstorage](https://github.com/parse-community/docs/blob/ccbf6226e66f22b3c9e60e068a6b826dad68212e/_includes/js/local-datastore.md) (https://github.com/parse-community/docs/pull/558)  up to date with the last [2.2.0 release ](https://github.com/parse-community/Parse-SDK-JS/pull/733) ?

dplewis:
Yes, I also opened a PR to address your sync feature. Waiting for 2.2.0 to be release to npm

GoMino:
Great I will have a look at your pull request https://github.com/parse-community/Parse-SDK-JS/pull/734 when I have time!
First I will test the 2.2.0 release to understand how this new feature is behaving (for offline read only access)...

PR #614, opened 2018-07-28
Adds ability to watch the build

flovilmart:
@dplewis hope this will help you, check this branch out and let me know if it does what your're expecting.


dplewis:
@flovilmart Thanks a lot! Might want to update that `CONTIBUTION.MD`...

dplewis:
Also Pin the package, don't want any more drama lol

PR #624, opened 2018-08-06
:zap: Release v2.0.1

flovilmart:
This PR does a few things:

1. fix the issue with the headers, when running with an older parse-server
2. updates the documentations for Cloud Code for parse-server 3.0. This is not ideal that the Cloud Code docs isn't in parse-server itself, but that's how it is, perhaps we could do something about it like a like or I dunno

### DIFF
--- a/src/CloudCode.js
+++ b/src/CloudCode.js
@@ -2,11 +2,13 @@
  * Defines a Cloud Function.
  *
  * **Available in Cloud Code only.**
+ * **Starting parse-server 3.0.0, all cloud hooks and functions are asynchronous**
+ * **All response objects are removed and replaced by promises / async/await based resolution**
  *
  * @method define
  * @name Parse.Cloud.define
  * @param {String} name The name of the Cloud Function
- * @param {Function} data The Cloud Function to register. This function should take two parameters a {@link Parse.Cloud.FunctionRequest} and a {@link Parse.Cloud.FunctionResponse}
+ * @param {Function} data The Cloud Function to register. This function can be an async function and should take one parameter a {@link Parse.Cloud.FunctionRequest}.
  */
 
 /**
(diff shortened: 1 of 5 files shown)


flovilmart:
@miconblog can you have a look there? as you went through the upgrade process, I'd like to release the fixes in the 2.0.1 as well as the upgrading guide.

miconblog:
yeb! i am already reading this guide. and migrating my project

flovilmart:
Let me know if there's anything missing or that you believe we should put :)

src/CloudCode.js
```diff
@@ -16,19 +18,19 @@
  *
  * If you want to use afterDelete for a predefined class in the Parse JavaScript SDK (e.g. {@link Parse.User}), you should pass the class itself and not the String for arg1.
  * ```
- * Parse.Cloud.afterDelete('MyCustomClass', (request) => {
+ * Parse.Cloud.afterDelete('MyCustomClass', async (request) => {
```
miconblog: i just curios about this CloudCode. Is it work in browser now? or is your long-team plan?

flovilmart: No it's in the server, but that's easier to have the whole cloud code API documentation with the rest of the JS SDK, we may move it at a later point if we generate a bigger API docs for parse-server.

miconblog: okay, i see i just confused :)

src/CloudCode.js
```diff
@@ -62,19 +64,19 @@
  *
  * If you want to use beforeDelete for a predefined class in the Parse JavaScript SDK (e.g. {@link Parse.User}), you should pass the class itself and not the String for arg1.
  * ```
- * Parse.Cloud.beforeDelete('MyCustomClass', (request, response) => {
```
miconblog: is it remove response in beforeDelete? 
i was used it before like this:
```
Parse.Cloud.beforeDelete('RedBookNote', async function (request, response) {
    response.success();
});

```

flovilmart: In parse-server 3.0 (the the docs update is for parse-server 3.0)

flovilmart: I understand this leads to a lot of confusion, I'll bring in the CLoud Code docs in parse-server instead. And just add the link here. I'll revert cloud code docs for now :)

miconblog: goood :)

Issue #558, opened 2018-03-26
Parse.File upload problems


EnricoMonese:
I've been trying this approach as described in the docs
```js
var fileUploadControl = $("#profilePhotoFileUpload")[0];
if (fileUploadControl.files.length > 0) {
  var file = fileUploadControl.files[0];
  var name = "photo.jpg";

  var parseFile = new Parse.File(name, file);
  parseFile.save();
} 

```
But when I try to do `.save()`  it throws this error a bunch of times
`error: Invalid file upload. code=130, message=Invalid file upload. `

This only happens when using `File`,  if I use base64 data it works albeit being slower.

samuelantonioli:
Most likely it has something to do with the file adapter you're using in parse-server. Add some logs of the HTTP requests or similar so we can help you better.

EnricoMonese:
I'm no longer using Parse, but the PR above seems to address the problem.

PR #628, opened 2018-08-09
Fix Parse File upload problem via ajax

miconblog:
Before this PR: 

```javascript
var parseFile = new Parse.File('filename', file);
parseFile.save().then((f)=>{
    console.log(f.toJSON())  // {__type: "File", name: undefined, url: undefined}
});
```

i think it is a bug and related this issue: #558 
RESTController.ajax method was return with ```response``` object, you can see that Implementation in [here](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/RESTController.js#L55) and [here](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/RESTController.js#L114)

### DIFF
--- a/src/ParseFile.js
+++ b/src/ParseFile.js
@@ -254,7 +254,7 @@ var DefaultController = {
       url += '/';
     }
     url += 'files/' + name;
-    return CoreManager.getRESTController().ajax('POST', url, source.file, headers);
+    return CoreManager.getRESTController().ajax('POST', url, source.file, headers).then(res=>res.response)
   },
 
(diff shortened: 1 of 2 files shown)


flovilmart:
The bug you are describing should only be present on the master branch, and only since a few days ago. Ca you confirm ?

miconblog:
yes, sure!

flovilmart:
How is it related to the bug you are describing? how did you encounter this issue in the first place?

miconblog:
yesterday I upgrade the SDK from 1.11.1 to 2.0 and i found this problem in my web project.

flovilmart:
Excellent :) Thanks!

PR #629, opened 2018-08-09
RestController  - Use GET requests when possible

ridem:
**Note: I'll be happy to edit the tests impacted if it's considered for merging**

This PR intends to use GET requests whenever possible instead of forcing POST requests.

While this probably makes very little difference for server-to-server requests, using GET requests allows for browser-caching and "CDN caching" when used in a web browser.
Right now, parse-server sends back ETag for POST requests, which is of no use for the web browser. Putting the API behind a service like Cloudfront isn't useful if all the requests made are POST requests, as they cannot be cached.

Incidentally, this makes Parse JS use its own REST API as per its specifications with regards to using the HTTP Headers (https://docs.parseplatform.org/rest/guide/)

We might want to actually properly map all the other HTTP methods properly as well. (DELETE, PUT)

### DIFF
--- a/src/RESTController.js
+++ b/src/RESTController.js
@@ -71,6 +71,21 @@ function ajaxIE9(method: string, url: string, data: any) {
   });
 }
 
+function serialize(obj, prefix) {
+  var str = [],
+    p;
+  for (p in obj) {
+    if (obj.hasOwnProperty(p)) {
+      var k = prefix ? prefix + "[" + p + "]" : p,
+        v = obj[p];
+      str.push((v !== null && typeof v === "object") ?
+        serialize(v, k) :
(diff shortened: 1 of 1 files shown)


flovilmart:
While I understand the motivation, I recall that originally, the SDK was written with POST request in order to avoid potential issues around passing headers to the server using GET requests. 

@andrewimm may have more informations about this.

I won't consider it for merging now, as:

1. There is no tests at all, integration or unit that ensures everything works fine.
2. The SDK has been working forever with the current strategy, I am not sure why it would make sense to change it now

Please open an issue first so we ca try to get ahold of the original SDK authors so they can answer the underlying question.

flovilmart:
cc @drew-gross

ridem:
Ok, thanks for your feedback! I opened an Issue as you suggested

PR #632, opened 2018-08-10
Support includeAll Query

dplewis:
https://github.com/parse-community/parse-server/pull/4838

provides shorthand for include('*')

### DIFF
--- a/src/ParseQuery.js
+++ b/src/ParseQuery.js
@@ -183,7 +183,6 @@ class ParseQuery {
   className: string;
   _where: any;
   _include: Array<string>;
-  _includeAll: boolean;
   _select: Array<string>;
   _limit: number;
   _skip: number;
@@ -217,7 +216,6 @@ class ParseQuery {
 
     this._where = {};
     this._include = [];
-    this._includeAll = false;
     this._limit = -1; // negative limit is not sent in the server request
     this._skip = 0;
     this._extraOptions = {};
@@ -282,9 +280,6 @@ class ParseQuery {
     if (this._include.length) {
       params.include = this._include.join(',');
     }
-    if (this._includeAll) {
-      params.includeAll = true;
-    }
     if (this._select) {
       params.keys = this._select.join(',');
     }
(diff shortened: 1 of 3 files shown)


flovilmart:
Can we use the '*' as an include argument? 
```
query.include('*')
```
and perhaps includeAll being a shorthand for that? This will be easier to handle on the server also instead of a new parameter.

dplewis:
@flovilmart This is what you mean right?

flovilmart:
I mean to treat include all as just using `.include('*')` which is a change that isn't required in the SDK in the end if we support it server side :)

dplewis:
I'll make an update server side. While preserving https://github.com/parse-community/parse-server/pull/4838

src/ParseQuery.js
```diff
@@ -1313,6 +1315,10 @@ class ParseQuery {
   /**
    * Includes nested Parse.Objects for the provided key.  You can use dot
    * notation to specify which fields in the included object are also fetched.
+   * 
+   * If you want to include all nested Parse.Objects pass in '*'
+   * <pre>query.include('*');</pre>
```
flovilmart: perhaps we need to put a version number here?

dplewis: Which version? 2.1.0?

flovilmart: I mean server version?

dplewis: I have no idea mate, I can try to backtrack

flovilmart: Ok ok, so basically it already works if you pass include All in the rest SDK.

dplewis: Yes that is correct. Basically this won't work until the next release unless I re-add includeAll to this SDK

flovilmart: Yeah, let’s just use the * notation then we’ll perhaps be able to expand on with reg ex etc...

flovilmart: ❤️

dplewis:
@flovilmart can this be merged?

src/ParseQuery.js
GoGross: How best can we update the guides simultaneously with the API docs

PR #634, opened 2018-08-10
Support nor Query

dplewis:
https://github.com/parse-community/parse-server/pull/4768

### DIFF
--- a/src/ParseQuery.js
+++ b/src/ParseQuery.js
@@ -237,7 +237,6 @@ class ParseQuery {
 
   /**
    * Adds constraint that all of the passed in queries match.
-   * @method _andQuery
    * @param {Array} queries
    * @return {Parse.Query} Returns the query, so you can chain this call.
    */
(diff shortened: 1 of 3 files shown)


dplewis:

src/ParseQuery.js
```diff
@@ -1422,6 +1435,24 @@ class ParseQuery {
...
+   * Constructs a Parse.Query that is the NOR of the passed in queries.  For
+   * example:
+   * <pre>const compoundQuery = Parse.Query.nor(query1, query2, query3);</pre>
+   *
+   * will create a compoundQuery that is a nor of the query1, query2, and
+   * query3.
+   * @param {...Parse.Query} var_args The list of queries to AND.
+   * @static
+   * @return {Parse.Query} The query that is the AND of the passed in queries.
+   */
```
flovilmart: Docs are wrong

src/ParseQuery.js
```diff
@@ -250,6 +249,20 @@ class ParseQuery {
     return this;
   }
 
+  /**
+   * Adds constraint that all of the passed in queries match.
```
flovilmart: Docs need a change

src/ParseQuery.js
dplewis: Is there a jsdocs checker / linter we can add?

flovilmart: I mean the text in the docs ;) it says AND as it should says NOR. :)

src/ParseQuery.js
flovilmart: It should say:

‘Adds constraint that none of the passed queries match’

src/ParseQuery.js
dplewis: The PHP SDK has a document checker. In this case the @param is wrong also

flovilmart: Not sure about that, you should look at them with `npm run docs`.

My point was on the whole documentation that it didn’t match what the function was doing

PR #635, opened 2018-08-10
Updates docs generation template

flovilmart:
before | after
---|---
capture d ecran le 2018-08-10 a 14 28 49|capture d ecran le 2018-08-10 a 14 28 34

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "xmlhttprequest": "1.8.0"
   },
   "devDependencies": {
+    "@parse/minami": "^1.0.0",
     "babel-jest": "23.2.0",
     "babel-plugin-flow-comments": "6.3.19",
     "babel-plugin-inline-package-json": "2.0.0",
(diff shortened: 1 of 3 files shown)


dplewis:
Is there a specific reason why you changed the template?

On mobile they both look the same. On desktop the Before has better padding. After has better font.

flovilmart:
The template has a fix for the generation for Parse server :) also it’s a fork on our repo so we can gradually make it look better for all docs

package.json
```diff
@@ -33,6 +33,7 @@
     "xmlhttprequest": "1.8.0"
   },
   "devDependencies": {
+    "@parse/minami": "^1.0.0",
```
dplewis: Ok I was wondering where this came from, I didn't see  the fork.

Should this build against the master branch?

flovilmart: https://github.com/nijikokun/minami/compare/master...parse-community:master

it's published already as an Npm package so we're good.

PR #638, opened 2018-08-13
eslint on src and src/__tests__

flovilmart:


src/Cloud.js
```diff
@@ -59,23 +59,22 @@ export function run(
   return CoreManager.getCloudController().run(name, data, requestOptions);
 }
 
- /**
+/**
   * Gets data for the current set of cloud jobs.
   * @method getJobsData
   * @name Parse.Cloud.getJobsData
   * @param {Object} options
```
dplewis: There are a few places where you removed options parameter but they still exist in the docs. 

https://github.com/gajus/eslint-plugin-jsdoc

or 

https://eslint.org/docs/rules/valid-jsdoc

Maybe we can use something like these plugin to validate the docs.

flovilmart: Right, gotta have a look.

flovilmart: valid-jsdoc yields an additional 200 errors ;) let's add it later ;)

dplewis: I agree everything looks good to me here

flovilmart: awesome!

PR #684, opened 2018-10-21
Save Method for Parse.Config

Moumouls:
Allow to use `.save()` on Parse.Config when a masterKey is defined to avoid the use of `Parse._request` #684 

- [x] Write Method
- [x] Write Test
- [x] Valid the `save()` then `get()` strategy @flovilmart ?
- [x] Write Documentation Guide
- [ ] Write Documentation API

### DIFF
--- a/src/ParseConfig.js
+++ b/src/ParseConfig.js
@@ -88,10 +88,10 @@ class ParseConfig {
    * @return {Promise} A promise that is resolved with a newly-created
    *     configuration object or with the current with the update.
    */
-  static save(attrs,params) {
+  static save(attrs) {
     const controller = CoreManager.getConfigController();
     //To avoid a mismatch with the local and the cloud config we get a new version
-    return controller.save(attrs,params).then(() => {
+    return controller.save(attrs).then(() => {
       return controller.get();
     },(error) => {
       return Promise.reject(error);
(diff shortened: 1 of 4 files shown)


src/ParseConfig.js
```diff
@@ -158,6 +175,37 @@ var DefaultController = {
...
+    }
+    const encodedAttrs = {};
+    for(const key in attrs){
+      encodedAttrs[key] = encode(attrs[key])
+    }
+    return RESTController.request(
+      'PUT',
+      'config',
+      { params: encodedAttrs },
+      { useMasterKey: true }
```
flovilmart: As the masterKey is forced here, I am not sure we should have the Params check above. If the masterKey is not provided at initialization the save will fail anyway

Moumouls: It's just for an UX purpose and an integrity reason about the other `.save()` method.
To show an error during a save without the master can help other developers to understand how it works and then adjust their code.
It seems good for me !

flovilmart: In this case, without the masterKey set in the SDK it will never work. So this is quite some overhead to write useMasterKey: true everywhere without providing anything in return

Moumouls:
Not passed values can't be overriden during the`.save()` according to the documentation : [Parse Rest Guide Config](https://docs.parseplatform.org/rest/guide/#config)
It's a Parse-Server test and not a Parse SDK test ?

src/__tests__/ParseConfig-test.js
```diff
@@ -142,16 +142,6 @@ describe('ParseConfig', () => {
     });
   });
 
-  it('cant save a config object without masterkey', (done) => {
-    ParseConfig.save({str: 'hello2','num':46}).then((config) => {
-      expect(config).toBe(-1)
```
flovilmart: THis should still fail.

Moumouls:
I just wrote the guide doc, the API doc is missing
[PR Docs JS Guide Config](https://github.com/parse-community/docs/pull/573)

src/__tests__/ParseConfig-test.js
Moumouls: Do I have to re-commit this test ?

Moumouls:
What's missing here for merge ?

flovilmart:
Nothing, I just merged back master and was waiting for the CI to complete!

Thanks!

PR #686, opened 2018-10-24
Ability to ignore constructor class map 

georgesjamous:
Adds the ability to ignore classMap when creating object. Object will be returned as an instance of ParseObject if ignoresClassMap is set to true.

My use case is i need to load a Session object from an already exported json object.
`Parse.Object.fromJson({ className: '_Session', sessionToken: "r:123..", ...otherStuff })`
Currently this will throw an error preventing readonly keys from being edited.

This should be safe since most restrictions are aleady being checked in server side.

### DIFF
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -1403,8 +1403,8 @@ class ParseObject {
    * @param {Object} json The JSON map of the Object's data
    * @param {boolean} override In single instance mode, all old server data
    *   is overwritten if this is set to true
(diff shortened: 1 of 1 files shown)


georgesjamous:
Its for something unrelated, I was checking the feasibilty of a readonly hook on _Session class in parse-server repo, it was failing becuase of ParseSession readonly key override, which was throwing error.

Anyway, i might have found another way. 
Closing this, will reopen if needed with additional explanation.

PR #692, opened 2018-11-01
Update job docs for removal of JobStatus parameter in job calls

fictorial:
Also, .message was moved from JobStatus into FunctionRequest.

### DIFF
--- a/src/CloudCode.js
+++ b/src/CloudCode.js
@@ -166,12 +166,12 @@
  * @property {function} message If message is called with a string argument, will update the current message to be stored in the job status.
  */
 
- /**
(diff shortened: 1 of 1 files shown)


flovilmart:
Thanks for the PR. Can you point out where the job status object has been removed?

fictorial:
In the execution of Cloud Code jobs, the job function only takes a single parameter now.

https://github.com/parse-community/parse-server/blob/master/3.0.0.md#migrating-jobs

fictorial:
I see, JobStatus still *exists*.  An object of type JobStatus is just not passed to the job function.  Also, it's JobRequest not FunctionRequest as per my first comment.

PR #699, opened 2018-11-25
VueJS Reactivity Problem #Vue #Reactive #Observe

dreadful44:
Vue.js have a [reactivity system](https://vuejs.org/v2/guide/reactivity.html). This system allows to track object changes. 

**Why**
Vue.js cannot track Parse Objects. Because objects creating out of Vue.js. Because of that reactivity system cannot understand changes. 

**What**
Parse's ObjectsState hashmap is handling objects values. But not creating or editing by Vue.js

**How**  
If Vue.js can set initial value of Parse's ObjectsState hashmap object, Vue's reactivity system can track change. Just add this code block to `main.js` in Vue.js
```javascript
import { state } from 'parse/lib/browser/SingleInstanceStateController'
Vue.set(state, 'className', '');
```
So i exported the objectState hashmap, imported in Vue app and has been set the initial value.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -22,6 +22,7 @@
         "no-multiple-empty-lines": 1,
         "prefer-const": "error",
         "space-infix-ops": "error",
-        "no-useless-escape": "off"
+        "no-useless-escape": "off",
+        "no-var": "error"
     }
 }
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,29 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 45
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+  - good first task
+  - up-for-grabs
+  - enhancement
+  - help wanted
+  - bug
+  - greenkeeper
+  - feature
+  - needs more info
+  - discussion
+  - question
+# Label to use when marking an issue as stale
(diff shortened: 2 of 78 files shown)


dreadful44:
Sorry for incomprehensible description of commit. I updated the description. @flovilmart 
thanks for help @bugra9

flovilmart:
This is very dangerous, as:

- the state object may be overridden/replaced internally by the Parse SDK
- you end up with two tracking systems, one managed by vue, one by the Parse SDK.

I am not a big fan of this overall. 

Can we have this exported as a function instead of a variable?

Also, I don’t believe Vue réactivity system is designed to track at once all your objects ever fetched by your client app, but only a selected few that are required to update your component. 

Let’s say you fetch 500 objects, all those 500 will be tracked, then you move to another component on screen and fetch another 500, and show only those extra 500, now you’re tracking changes in 1000 etc... I am not sure this is the behavior you want.

dreadful44:
Ok. I will refactor code and adding a function which is this issue  as soon as. Can you give me a time ? @flovilmart

flovilmart:
This is not what I meant

I meant 

export getState() { return objectState } 

Also, did you get a chance to read what I mentioned about this solution? Can you provide more elements regarding my concerns?

dplewis:
@dreadful44 Ping? Pong?

dreadful44:
Sorry. I can't write for a long time. if the code which is my wrote be merge to the master, vue track all object of parse. This is wrong solution. When vue track all object, browser is getting slow. I'm closing merge request. Thank you.

PR #701, opened 2018-11-29
Add batchSize to saveAll / destroyAll

dplewis:


dplewis:
I should add it to destroyAll too

dplewis:
@flovilmart Do you know why 20 was the default batchSize?

flovilmart:
No idea! Perhaps a sensible default, not to overload their servers

PR #702, opened 2018-12-01
Add Anonymous User Login

dplewis:
Similar to PHP SDK

### DIFF
--- a/integration/test/ParseUserTest.js
+++ b/integration/test/ParseUserTest.js
@@ -439,4 +439,13 @@ describe('Parse User', () => {
       done();
     });
   });
+
+  it('can login anonymous', (done) => {
+    Parse.User.enableUnsafeCurrentUser();
+    Parse.User.loginWithAnonymous().then((user) => {
+      assert(user.get('authData').anonymous.id);
+      assert.equal(user, Parse.User.current());
+      done();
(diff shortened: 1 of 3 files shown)


dplewis:
@flovilmart Should I add facebook and twitter to this?

flovilmart:
What do you mean?

dplewis:
Log in with facebook and twitter, like the php sdk

http://parseplatform.org/parse-php-sdk/classes/Parse.ParseUser.html#method_logInWithFacebook

flovilmart:
They are already provided through the facebookUtils

dplewis:
Ah I see, forgot that was there :P

cjbland:
Would it be possible to add a way to convert an anonymous account into a "real" account? This is what I do on iOS (https://docs.parseplatform.org/ios/guide/#anonymous-users) but following the same procedure in JavaScript seems to retain the anonymous `authData`.

dplewis:
@cjbland I don't see why not. iOS also has automaticUser, lazyLogin with LDS etc. 

If you want to get started on a new branch with `AnonymousUtils` and we can work on that.

dplewis:
Closing as will add AnonymousUtils in the future

PR #714, opened 2019-01-04
LiveQuery handle unset operation

dplewis:
Addresses https://github.com/parse-community/parse-server/issues/4998

### DIFF
--- a/src/__tests__/LiveQueryClient-test.js
+++ b/src/__tests__/LiveQueryClient-test.js
@@ -264,7 +264,7 @@ describe('LiveQueryClient', () => {
     expect(isChecked).toBe(true);
   });
 
-  fit('can handle WebSocket response unset field', async () => {
+  it('can handle WebSocket response unset field', async () => {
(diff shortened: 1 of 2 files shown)


flovilmart:
Thé op should be sent to the client as undefined is not serialized.

dplewis:
Thats a better solution. Instead of setting undefined like in this PR. Send undefined to the client

src/__tests__/LiveQueryClient-test.js
```diff
@@ -264,7 +264,7 @@ describe('LiveQueryClient', () => {
     expect(isChecked).toBe(true);
   });
 
-  it('can handle WebSocket response unset pointer', async () => {
+  fit('can handle WebSocket response unset field', async () => {
```
flovilmart: oops

dplewis:
@flovilmart Thanks! Did you see my last comment?

flovilmart:
It should likely update yes,

PR #716, opened 2019-01-10
Added route to reset password with token

moonion:
Implemented another route for password reset flow, that allows setting new password from client side using the user's token that he receives by email with link.

This allows setting password through Parse function on client side, instead of manually sending request to server's public API router, avoiding direct interaction between user and server. 

The pull request that adds server-side route is here: https://github.com/parse-community/parse-server/pull/5288

### DIFF
--- a/src/CoreManager.js
+++ b/src/CoreManager.js
@@ -124,6 +124,7 @@ type UserController = {
   become: (options: RequestOptions) => Promise;
   logOut: () => Promise;
   requestPasswordReset: (email: string, options: RequestOptions) => Promise;
+  resetPasswordSetNew: (username: string, token: string, new_password: string, options: RequestOptions) => Promise;
   updateUserOnDisk: (user: ParseUser) => Promise;
   upgradeToRevocableSession: (user: ParseUser, options: RequestOptions) => Promise;
   linkWith: (user: ParseUser, authData: AuthData) => Promise;
@@ -378,6 +379,7 @@ module.exports = {
       'become',
       'logOut',
       'requestPasswordReset',
+      'resetPasswordSetNew',
       'upgradeToRevocableSession',
       'linkWith',
     ], controller);
--- a/src/ParseUser.js
+++ b/src/ParseUser.js
@@ -691,6 +691,34 @@ class ParseUser extends ParseObject {
   }
 
   /**
+   * Sets new password for specified user that already requested password reset,
+   * using his token
+   *
+   * <p>Calls options.success or options.error on completion.</p>
+   *
+
+   * @param {String} username Username associated with the user that forgot
+   *     their password.
+   * @param {String} token Verification token for password reset
+   * @param {String} new_password New password, chosen by user
+   * @param {Object} options
+   * @static
+   * @returns {Promise}
+   */
+  static resetPasswordSetNew(username, token, new_password, options) {
+    options = options || {};
+    var requestOptions = {};
+    if (options.hasOwnProperty('useMasterKey')) {
+      requestOptions.useMasterKey = options.useMasterKey;
+    }
+
+    var controller = CoreManager.getUserController();
+    return controller.resetPasswordSetNew(
+      username, token, new_password
+    );
+  }
+
+    /**
    * Allow someone to define a custom User class without className
    * being rewritten to _User. The default behavior is to rewrite
    * User to _User for legacy reasons. This allows developers to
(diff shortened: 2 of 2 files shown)


flovilmart:
Thansk for thé PR, unfortunately the token is never accessible trough an API call, therefore this should never be possible. How do you plan to get the token?

moonion:
> Thansk for thé PR, unfortunately the token is never accessible trough an API call, therefore this should never be possible. How do you plan to get the token?

We are using custom reset password page on our client side, user accesses it thorugh the link in the email that he receives when requests password reset. This link contains token and we can use it on our client side. After user enters his new password, we pass it to the function along his username and token.

flovilmart:
So why don’t you make a GET request to the choose password endpoint? I see also the PR on the server which is mostly duplicating code.

moonion:
> So why don’t you make a GET request to the choose password endpoint? I see also the PR on the server which is mostly duplicating code.

Well, because the whole point is to be able to do it through Parse method from custom page with password form. Am i missing something important?

flovilmart:
you can make any form a reset password form with the current implementation. You do not need javascript, pure HTML work fine. Did you try with it?

moonion:
Yes, but this way it will be much easier to do. We have single-page application, so its easier for us to just call a Parse method and pass it the needed data and token, instead of making request ourselves. Besides, on different platforms we will have to use different ways to make requests, so having internal Parse method to do it will save us (and hopefully other people) time in the future projects.

flovilmart:
There is an endpoint existing already, if you plan to go further with this scenario, please use the existing endpoint. As for your justification, they do not justify having it in the SDK. As a simple `fetch` call or even forming a proper HTTP form just works as expected. Heck, as the change-password endpoint is a single a href would work. So I am not convinced this is simpler, nor necessary.

nashivan:
@flovilmart 

You didn't understand me.

We develop SPA without browsers redirect.

If I used the existing endpoint I won't receive errors like I want, because existing endpoint returns 302 header status and location for redirect.

current:
```javascript
function resetPassword(username, password, token) {
  const formData = new FormData();
  formData.append("token", token);
  formData.append("username", username);
  formData.append("new_password", password);
  return fetch(`${Parse.serverURL}/apps/${Parse.applicationId}/request_password_reset`, {
    method:"POST",
    body: formData
  })
    .then(result => result.json()) 
    .then(data => console.log('success', data))
    .catch(error => console.log('error', error)); // error fetch error because endpoint return 302 status for redirect, I always receive this error when data was sent correctly
} 
```

our solution:
```javascript
function resetPasswordSetNew(username, password, token) {
  return Parse.resetPasswordSetNew(username, token, new_password)
    .then(data => console.log('success', data)) 
    .catch(error => console.log('error', error)); // token was expired
}
```

flovilmart:
> If I used the existing endpoint I won't receive errors like I want, because existing endpoint returns 302 header status and location for redirect.

So what? you can probably parse the URL.

Also you are hitting the wrong endpoint (https://github.com/parse-community/parse-server/blob/master/src/Routers/PublicAPIRouter.js#L258)

Also you can determine wether an error happened as shown [here](https://github.com/parse-community/parse-server/blob/master/src/Routers/PublicAPIRouter.js#L177) by parsing the response query string.

moonion:
The endpoint  `'/apps/choose_password` returns  html form which is provided by  parse-server. [api](https://github.com/parse-community/parse-server/blob/46ac7e7f1183a1cb3ad74684f3803805ced166b0/src/Routers/PublicAPIRouter.js#L95) and [form](https://github.com/parse-community/parse-server/blob/master/views/choose_password)

when we use `/apps/:appId/request_password_reset`:  
[screensot](http://take.ms/H2wmW)


We want to render form on our SPA not on parse server with redirect to our SPA after success or error.

We want to send request reset password and change password like we use function Parse User.logIn and etc. 

We have created new endpoint and new function for SDK.

If you don't want to add new endpoint  we can 
1. Add support AJAX to  `/apps/:appId/request_password_reset` to return error without redirect
2. change `resetPasswordSetNew` function to support  `/apps/:appId/request_password_reset` because I think it is not Parse way to use next code
```javascript
 const formData = new FormData();
  formData.append("token", token);
  formData.append("username", username);
  formData.append("new_password", password);
  return fetch(`${Parse.serverURL}/apps/${Parse.applicationId}/request_password_reset`, {
    method:"POST",
    body: formData
  })
    .then(result => result.json()) 

```

flovilmart:
> We want to render form on our SPA not on parse server with redirect to our SPA after success or error.

If I visit this code [here](https://github.com/parse-community/parse-server/blob/master/src/Routers/PublicAPIRouter.js#L149)

it very well seems that you can configure both the **success page** and the r**eset password page**:
https://github.com/parse-community/parse-server/blob/master/src/Routers/PublicAPIRouter.js#L166-L189

Did you try  setting `config.passwordResetSuccessURL` to `#success`?

My main hurdle with your implementation is that 

1. It uses an endpoint that does not exist
2. This endpoint is a full copy paste of another existing one, to the exception of the response type
3. Nothing needs to belong in the SDK as it can and should be implemented on pure HTML form.

flovilmart:
Yes and copy is bad as I mentioned about 3 times! You can find a way to make it work with the current server setup and without adding any endpoint. This is my requirement. Open a new PR when you figure it out.

PR #718, opened 2019-01-11
LiveQuery override data on update

dplewis:
Since the update / event subscription is the latest version of an object we should update the server data and localDatastore. Per [discussion](https://github.com/parse-community/Parse-SDK-JS/pull/714#issuecomment-451615779)

Added a isPinned helper to LDS.


dplewis:

PR #719, opened 2019-01-14
Created an AnonymousUtils class to match what Android/iOS does

cjbland:
I could use a little help on figuring out how to remove/strip the anonymous `authData` after an anonymous user signs up.

### DIFF
--- a/src/AnonymousUtils.js
+++ b/src/AnonymousUtils.js
@@ -1,0 +1,68 @@
+import ParseUser from './ParseUser';
+
+/**
+ * Provides utility functions for working with Anonymously logged-in users. Anonymous users have
+ * some unique characteristics:
+ * <ul>
+ * <li>Anonymous users don't need a user name or password.</li>
+ * <li>Once logged out, an anonymous user cannot be recovered.</li>
+ * <li>When the current user is anonymous, the following methods can be used to switch to a
+ * different user or convert the anonymous user into a regular one:
+ * <ul>
+ * <li>signUp converts an anonymous user to a standard user with the given username and password.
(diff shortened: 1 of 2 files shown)


dplewis:
I was playing around with this over the weekend.

Here’s a branch I started 
https://github.com/parse-community/Parse-SDK-JS/compare/master...dplewis:anonymousUtils?expand=1

cjbland:
Your branch looks good, I shouldn't checked that before creating my PR. If I follow the code right you're still going to have an issue with converting an anonymous account to a "real" account. This is what I've been trying to get to work. In Android/iOS it clears out `authData` but I can't seem to figure out how to get JavaScript to do that. Clearing it out on the device/browser does not seem to propagate to the database. I'm going to step through the Android code today to figure out what it is doing differently.

dplewis:
@cjbland Sorry for the late reply

> clears out authData but I can't seem to figure out how to get JavaScript to do that.

It seems like this is already handled by the SDK.

https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseUser.js#L291

Test here

https://github.com/parse-community/Parse-SDK-JS/blob/master/src/__tests__/ParseUser-test.js#L678

cjbland:
@dplewis Ah, you're right. I feel silly now. I guess I misunderstood the JavaScript guide for when trying to convert an anonymous account to a real account. Calling `setUsername()` instead of `set('username', '')` does clear out the `authData` field. My turn to apologize for the late reply...things have been crazy around here lately. Feel free to close/remove this PR and roll with what you created because it's a lot better :)

Issue #726, opened 2019-02-05
Setting authResponse is not supported


durunvo:
Hi, I wonder if anyone else face this issue when try to login with facebook?
Also happens when trying to login with email/password for facebook authenticated users.

The `/login` request was succeed. It returns status code 200 with users data.
However, the javascript sdk rejects the request and returns an error. `Setting authResponse is not supported`

I have tested in js version 2.0.2.

durunvo:
Found out that this is an error from Facebook sdk.

durunvo:
For more information, I have comment out `Parse.FacebookUtils.init` and the problem is gone.

durunvo:
This is weird but problem seems to be a conflict between Parse.FacebookUtils and Facebook official sdk.

Please see this line of code from unminified js.

https://github.com/nfriedly/facebook-js-sdk/blob/cd1f144d334e7e200b2ba6c7ca0e51636063a9b6/debug.js#L14268

It checks for `authResponse` and returns an error.

Please also look at this line of code of this project.

https://github.com/parse-community/Parse-SDK-JS/blob/8e0f00c5a3dd6043718fec1af1ddc1106058d5e1/src/FacebookUtils.js#L61

`authResponse ` is assigned to an option intentionally.

---

I solve this by change the facebook sdk url from `//connect.facebook.net/en_US/sdk.js` to `//connect.facebook.net/en_US/all.js`

The different is `sdk.js` and `all.js`

So maybe somebody can look into this issue, Thanks~

schurch:
I'm also seeing this issue.

Jauny:
Same for me, started to see this, our facebook login is broken because of this.

flovilmart:
The PR don’t add any test and the resolution doesn’t match with what was described (changing the SDK URL). So That’s what I’d least expect. Cc @dplewis, can you then review, merge and release please?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

RadianSmile:
> I solve this by change the facebook sdk url from `//connect.facebook.net/en_US/sdk.js` to `//connect.facebook.net/en_US/all.js`

this works for me

dplewis:
Closing via https://github.com/parse-community/Parse-SDK-JS/pull/728

PR #728, opened 2019-02-06
remove assign authResponse in FB.init

durunvo:
Base on this issue https://github.com/parse-community/Parse-SDK-JS/issues/726

I'm not sure what is the usage for `newOptions.authResponse` in this library. So I remove the line that assign `authResponse` to `newOptions`.

Anyone feel free to comment.

### DIFF
--- a/src/FacebookUtils.js
+++ b/src/FacebookUtils.js
@@ -47,11 +47,6 @@ const provider = {
         (expiration.getTime() - new Date().getTime()) / 1000 :
         0;
 
-      const authResponse = {
-        userID: authData.id,
-        accessToken: authData.access_token,
-        expiresIn: expiresIn
-      };
       const newOptions = {};
       if (initOptions) {
         for (const key in initOptions) {
@@ -67,7 +62,7 @@ const provider = {
       // from a Parse User that logged in with username/password.
       const existingResponse = FB.getAuthResponse();
       if (existingResponse &&
-          existingResponse.userID !== authResponse.userID) {
+          existingResponse.userID !== authData.id) {
         FB.logout();
       }
 
--- a/src/FacebookUtils.js
+++ b/src/FacebookUtils.js
@@ -9,7 +9,6 @@
  * @flow-weak
  */
 /* global FB */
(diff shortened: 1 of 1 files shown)


src/FacebookUtils.js
```diff
@@ -58,8 +58,6 @@ const provider = {
           newOptions[key] = initOptions[key];
         }
       }
-      newOptions.authResponse = authResponse;
```
schurch: Thanks for the @durunvo. I am also seeing this issue. In Facebook's docs I can't see what passing this to `FB.init` would have been for, but maybe someone who knows more about this can explain why.

If you do remove this, perhaps there is no need for the whole `authResponse` object on line 50 and line 72 could be changed to `existingResponse.userID !== authData.id`?

durunvo: FB.init require just 4 keys
```
{
  appId      : '{your-app-id}',
  status     : true,
  xfbml      : true,
  version    : 'v2.7' // or v2.6, v2.5, v2.4, v2.3
}
```
So yeah, I also think that it is not needed. But I want some explanation first just to make sure that this is a right thing to do.

But anyway I will push another commit according to your suggestion.

Jauny:
thanks for doing this @durunvo. Who can help to get this pushed through? Our Facebook login is broken right now and it prevents us from pushing a new release to production.

dplewis:
@durunvo Thanks for the PR! I tried this in production and it works great.

I'm surprise by the all.js and sdk.js fix.

http://connect.facebook.net/en_US/sdk/debug.js
http://connect.facebook.net/en_US/all/debug.js

Both throw the authResponse is no supported error.

The FacebookUtils file doesn't have a single test case :(
I can do it in another ticket

durunvo:
Hey @dplewis, I can see you've merged my fork into master. Do you need help with anything else?

dplewis:
@durunvo We are going to do a Release for this SDK soon to include your fix.

The parse-server repo has a lot of PR that could use clean up / review if you want to help.

https://github.com/parse-community/parse-server/pulls

PR #729, opened 2019-02-06
Improve dot notation for updating nested objects

dplewis:


dplewis:
@flovilmart This test case works for PG also

flovilmart:
Well, there is to test also all the other cases: setting a value on ‘key.path’ and increment when it’s not an object stored behind, unset(‘key.path’) etc...

flovilmart:
I am very worry bad things will happen with that much decoupling but if you beliebe this is ‘needed’ go ahead 

dplewis:
@flovilmart I did run into two issues with this.

You can't set or update nested documents on new objects.

```
const obj = new TestObject();
obj.set('objectField.number', 20); // Throws internal server error because field doesn't exist
```

If field exist the local data doesn't update until saved

```
const obj = new TestObject({ objectField: { number: 5 } });
await obj.save();
// obj.get('objectField').number is 5
obj.set('objectField.number', 20);
// obj.get('objectField').number is still 5
await obj.save();
// obj.get('objectField').number is now 20
```

flovilmart:
Do what you feel is best 

dplewis:
@flovilmart I resolved both issues I stated on the client side. Can you look this over?

I found an existing bug but it’s not important.

flovilmart:
I still have bad feelings about this, but go ahead.

dplewis:
Thanks! I’m going to play around with it more. There a lot of features that use `set()` like add, addUnique etc

PR #730, opened 2019-02-08
Support setting user from JSON (hydrate)

dplewis:


flovilmart:
this gonna end up being a complete mess adding all those undocumented, un discussed and unstable features. Well have fun!

Issue #594, opened 2018-07-05
Subscribe to ParseObject attribute changes


Aetherall:
Hi there !

It there any way to subscribe to attribute changes of an Parse.Object ?

I use react and It would be awesome if we could rerender a component when an object attribute change, 
so the UI stays in sync with the cache

I wanted to override a mutating function so I can add a callback, but It seems that there is multiple functions affecting the state of the object attributes ( to handle the data from server and the optimistic data )

Also, it seems to me that there is no way to relate the object attributes to the ParseObject itself, so even if we abstract the mutating part of all the `setServerData`, `mergeFirstPendingState`, ... mutating functions into it own `applyNewData` function, we cannot access the ParseObject to let it know the state is changed

Anyone has any thoughts on this ? :)

Thanks !

samuelantonioli:
I've struggled a lot with this (using Vue). I had to build [my own solution](https://gist.github.com/samuelantonioli/50b01142ce15d8a069fc23ea8b975873) using `Object.defineProperty` (ES6-Proxy doesn't have proper IE support) which essentially proxies all the attributes.  

It's a bit complicated, but look into `ParseObject.dirtyKeys` - it gives you the keys that were changed.  
I think this needs a major rewrite of some parts in the SDK to support React, Vue and others out of the box (using POJOs and events or hooks).  

Historically, Parse used the Backbone.Model, that's the reason it's not very compatible with the current generation of web frameworks. Changing the interface would be a major thing that I don't think will happen for compatibility reasons.

Aetherall:
I Actually succeded using events and HOC's, and overloading the save / destroy / set methods of Parse.Object

If anyone needs help on this, ask here

Maybe In the future I will write a react-parse package, in the Apollo style

flovilmart:
Have you had a look at this project? https://github.com/andrewimm/parse-lite

Aetherall:
It looks nice but I was really looking for advanced cache management ( I actually used the LiveQueryTools file from server to update cached queries on client when a subscription event or rest save occurs )

samuelantonioli:
@Aetherall Would love to see your code.

Aetherall:
@samuelantonioli 

You have to create a class extending Parse.Object and add to it subscription abilities

```js
export default class Collection extends Parse.Object {
	// === === === React Compat === === ===

	// === Override mutators ===

	async save(...args) {
		const res = await super.save(...args)
		this.dispatch()
		return res
	}

[...]

PR #731, opened 2019-02-08
Subscribe to object updates (LiveObject)

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/594

Any advice would be appreciated.

### DIFF
--- a/integration/test/ParseObjectTest.js
+++ b/integration/test/ParseObjectTest.js
@@ -1702,6 +1702,16 @@ describe('Parse Object', () => {
     done();
   });
 
+  it('can subscribe error', async (done) => {
+    const object = new TestObject({ test: 'test' });
+    const subscription = object.subscribe();
+    subscription.on('error', (error) => {
+      assert.equal(error.message, 'whoops!');
+      done();
+    });
+    subscription.emit('error', new Error('whoops!'));
+  });
+
   it('can subscribe new objects', async (done) => {
     const object = new TestObject({ test: 'test' });
     const subscription = object.subscribe();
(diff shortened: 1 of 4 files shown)


dplewis:
@flovilmart @acinader I found an existing bug in parse regarding increment. It's not a big deal but I thought it should be reported.

```
object.set('number', 5);
object.increment('number', 10);
object.get('number'); // returns 15
object.get('number'); // returns 25
object.get('number'); // returns 35
await object.save();
object.get('number'); // returns 15
```

https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ObjectStateMutations.js#L136

Since we don't keep track of which pendingOps we use when estimating values, increment will always compound.

flovilmart:
How do you pass a custom liveQuery client ? I agree it’s a nice sugar, but how does it behaves in case the client disconnects? Is not connected etc...

dplewis:
I was focused on the local object changes for this PR, I didn't consider if a live query client existed. 

I believe I could reuse the current LiveQuery implementation. I'll play around with it

flovilmart:
Ah that’s confusing, really... 

flovilmart:
Also the original requester of the feature is unlikely to use this 7 month after the initial request. I understand the motivation, but it seems that it’s the wrong solution 

dplewis:
I can post a better solution that uses LiveQueryClient that works if an object has an `objectId` if there is enough interest in this. What do you think?

flovilmart:
I am not sure this feature makes sense at all in this kind of stateful SDK. If anything, an implementation of the SDK based on react hooks would be more valuable than any other event driven implementation. If you need this feature as it is, go ahead. I don’t want to prevent anyone from using the project as their own

PR #734, opened 2019-02-13
LocalDatastore can update from server

dplewis:
Allows for updating LocalDatastore with server data

Any advice would be helpful

### DIFF
--- a/src/LocalDatastore.js
+++ b/src/LocalDatastore.js
@@ -260,8 +260,7 @@ const LocalDatastore = {
     }
   },
 
-  // sync LDS with network
-  async _syncIfNeeded() {
+  async _updateFromServer() {
     if (!this.isEnabled || this.isSyncing) {
       return;
     }
--- a/src/LocalDatastore.js
+++ b/src/LocalDatastore.js
@@ -260,8 +260,17 @@ const LocalDatastore = {
     }
   },
 
-  async _updateFromServer() {
-    if (!this.isEnabled || this.isSyncing) {
+  /**
+   * Updates Local Datastore from Server
+   *
+   * <pre>
+   * await Parse.LocalDatastore.updateFromServer();
+   * </pre>
+   *
+   * @static
+   */
+  async updateFromServer() {
+    if (!this.checkIfEnabled() || this.isSyncing) {
       return;
     }
     const localDatastore = await this._getAllContents();
@@ -278,7 +287,9 @@ const LocalDatastore = {
     const pointersHash = {};
     for (const key of keys) {
       const [className, objectId] = key.split('_');
-      pointersHash[className] = pointersHash[className] || new Set();
+      if (!(className in pointersHash)) {
+        pointersHash[className] = new Set();
+      }
       pointersHash[className].add(objectId);
     }
     const queryPromises = Object.keys(pointersHash).map(className => {
--- a/src/LocalDatastore.js
+++ b/src/LocalDatastore.js
@@ -260,9 +260,9 @@ const LocalDatastore = {
     }
   },
 
-  // If network is reconnected sync LDS with network
+  // sync LDS with network
   async _syncIfNeeded() {
-    if (!this.isEnabled || this.isSynced || this.isSyncing) {
+    if (!this.isEnabled || this.isSyncing) {
       return;
     }
     const localDatastore = await this._getAllContents();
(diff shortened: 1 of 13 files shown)


flovilmart:
Did you check how the other SDKs implémented those features? I’m not sure the server is properly equipped to ‘sync’ data when it’s been off sync? 

dplewis:
> Did you check how the other SDKs implémented those features?

If by other SDK's you mean iOS and android LDS, I don't believe they have this kind of feature. They do have `saveEventually` for syncing

> I’m not sure the server is properly equipped to ‘sync’ data when it’s been off sync?

I'm not looking for a true sync feature with this PR. Just when network reconnect perform a query.

flovilmart:
That’s not what I meant, when you will get network connection again, you will attempt to save object, and may override previous operations. But I guess you already knew that.

flovilmart:
The REstController also is not the best spot to get this feature in, as it’s responsibility is solely to connect to the API. It has nothing to do with the LDS.

dplewis:
> you will attempt to save object, and may override previous operations

This attempt doesn't happen on ie9 and when it does happen it only tries for 4 seconds max

> The REstController also is not the best spot to get this feature in

When do you believe the best implementation is? I would like to the SDK to handle it. Any advice would be appreciated

dplewis:

flovilmart:
@dplewis ask @acinader :) 

src/Parse.js
```diff
@@ -225,6 +225,23 @@ Parse.dumpLocalDatastore = function() {
     return Parse.LocalDatastore._getAllContents();
   }
 }
+/**
```
acinader: why here?  why not in LocalDatastore?

src/LocalDatastore.js
```diff
@@ -259,6 +260,54 @@ const LocalDatastore = {
...
+      }
+    }
+    if (keys.length === 0) {
+      return;
+    }
+    this.isSyncing = true;
+    const pointersHash = {};
+    for (const key of keys) {
+      const [className, objectId] = key.split('_');
+      pointersHash[className] = pointersHash[className] || new Set();
```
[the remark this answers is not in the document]
dplewis: I modified this from [include()]( https://github.com/parse-community/parse-server/blob/master/src/RestQuery.js#L787)

src/Parse.js
```diff
@@ -225,6 +225,23 @@ Parse.dumpLocalDatastore = function() {
...
+ * Updates Local Datastore from Server
+ *
+ * <pre>
+ * await Parse.updateLocalDatastoreFromServer();
+ * </pre>
+ *
+ * @static
+ */
+Parse.updateLocalDatastoreFromServer = function() {
+  if (!Parse.LocalDatastore.isEnabled) {
```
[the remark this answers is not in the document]
dplewis: Good question this should fail

src/LocalDatastore.js
```diff
@@ -259,6 +260,65 @@ const LocalDatastore = {
...
+      const objects = [].concat.apply([], responses);
+      const pinPromises = objects.map((object) => {
+        const objectKey = this.getKeyForObject(object);
+        return this.pinWithName(objectKey, object._toFullJSON());
+      });
+      await Promise.all(pinPromises);
+      this.isSyncing = false;
+    } catch(error) {
+      console.log('Error syncing LocalDatastore'); // eslint-disable-line
+      console.log(error); // eslint-disable-line
```
acinader: any reason not to use parse's logger so this stuff shows up in aggregated logs?

see middlewares, MogoTransform, PromiseRouter, etc. for examples

dplewis: something like

```
const Logger = require('parse-server/lib/logger');

Logger.logger.error(`Error syncing LocalDatastore: ${error}`);
console.log('Error syncing LocalDatastore', error);
```

acinader: Any reason to not follow the exact pattern we have already?

i.e. 

https://github.com/parse-community/parse-server/blob/master/src/PromiseRouter.js#L10
```
import log from './logger';
...
log.error('message', object);
...
```

and remove the console.log and the eslint exception comment?

dplewis: Thats the parse-server repo, there isn't a logger in this repo.

acinader: ha.  riiiight.  yeah, lgtm. ;) . I don't use the client js sdk is the only lame excuse i have for that one.

Issue #717, opened 2019-01-11
ReferenceError: regeneratorRuntime is not defined


dplewis:
Using parse through CDN or `npm install parse@version` works

I get the following error `ReferenceError: regeneratorRuntime is not defined`

If I try to run get the bleeding edge in a node project

`npm install parse-community/Parse-SDK-JS.git#master --save`

or 

build it myself and copy parse.min.js from the dist folder into a simple html file.

I don't know much about babel to fix the issue

farhansyed9:
Also have this issue, did you find a solution?

dplewis:
I haven't you can try playing around with babel to see if you can get it to work

agoldis:
Facing the same issue with [Parse@2.2.0](https://www.npmjs.com/package/parse)

https://github.com/parse-community/Parse-SDK-JS/blob/master/gulpfile.js#L18 specifies `@babel/plugin-transform-runtime` [configuration option](https://babeljs.io/docs/en/babel-plugin-transform-runtime#regenerator):
```
regenerator: false
```

That instructs babel to transpile `async` functions into a code that assumes existence of globally defined `regeneratorRuntime`.

When you start to use the SDK in a project that *does not* define a global `regeneratorRuntime` you'll bump into the exception.

There are some workarounds, but appropriate solution highly depends on your project's specific configuration>

1. Manually include `regeneratorRuntime`
Use https://www.npmjs.com/package/regenerator-runtime and make sure the following code is executed **before** Parse is initialized:

```
import regeneratorRuntime from 'regenerator-runtime/runtime';
window.regeneratorRuntime = regeneratorRuntime;
```

2. Use babel to try and inject the variable
Use [`@babel/polyfill`](https://babeljs.io/docs/en/babel-polyfill) together with [`@babel/preset-env`](https://babeljs.io/docs/en/babel-preset-env) + customize the option [`useBuildIns`](https://babeljs.io/docs/en/babel-preset-env#usebuiltins) to inject required polyfills. That's quite tricky, though and very implicit.

### For maintainers
If there's no specific reason to use this option [here](https://github.com/parse-community/Parse-SDK-JS/blob/master/gulpfile.js#L18), I would suggest to set it to `true`. That will cause babel to use `@babel/runtime/regenerator` instead of a global variable, which is much easier to set up (as described in [`generatorAliasing`](https://babeljs.io/docs/en/babel-plugin-transform-runtime#regenerator-aliasing))

[...]

PR #740, opened 2019-02-21
(fix) regeneratorRuntime is not defined

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/717

If you are in a project that doesn't have a global `regeneratorRuntime` defined i.e Browser an error is thrown.

This fix causes babel to use `@babel/runtime/regenerator` instead of a global variable.

`regenerator` is used whenever async / generator functions are used

Thanks @agoldis


dplewis:
@acinader Thanks, we should do another release

acinader:
Yes.  2.2.1?

dplewis:
Yeah

PR #750, opened 2019-03-18
Support for Anonymous Users

dplewis:
Also fixed issue with sessionToken not being passed when saving users.

### DIFF
--- a/src/AnonymousUtils.js
+++ b/src/AnonymousUtils.js
@@ -58,6 +58,7 @@ const AnonymousUtils = {
    *     The user must be logged in on this device.
    * @return {Boolean} <code>true</code> if the user has their account
    *     linked to an anonymous user.
+   * @static
    */
   isLinked(user) {
     const provider = this._getAuthProvider();
@@ -70,6 +71,7 @@ const AnonymousUtils = {
    * @method logIn
    * @name Parse.AnonymousUtils.logIn
    * @returns {Promise}
+   * @static
    */
   logIn() {
     const provider = this._getAuthProvider();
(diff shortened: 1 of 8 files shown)


src/AnonymousUtils.js
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * @flow-weak
+ */
+import ParseUser from './ParseUser';
+const uuidv4 = require('uuid/v4');
+
+let registered = false;
+
+const authenticationProvider = {
+  authData: {
```
acinader: can you name the top level 'provider'? little confused as to why we would have authData: authData:

src/AnonymousUtils.js
```diff
@@ -0,0 +1,113 @@
...
+   * @returns {Promise}
+   * @static
+   */
+  link(user) {
+    const provider = this._getAuthProvider();
+    return user._linkWith(provider.getAuthType(), provider.authData);
+  },
+
+  _getAuthProvider() {
+    const provider = authenticationProvider;
```
acinader: this is a no-op. You could just take out this line and change 106 to:

```
ParseUser._registerAuthenticationProvider(authenticationProvider);
```

but even better is moving the declaration of `authenticationProvider` into this function so that the `uuid` is only generated IF it is needed.  As you have it here, any time this file is required, it'll generate the `uuid` which is relatively expensive.  Make it lazy so it is only generated if needed.

src/AnonymousUtils.js
```diff
@@ -0,0 +1,113 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @flow-weak
+ */
+import ParseUser from './ParseUser';
+const uuidv4 = require('uuid/v4');
+
+let registered = false;
```
acinader: would be nice if this were exposed on the parse user.....

acinader:
super nice.

acinader:
@dplewis you wanna cook up a release?

Issue #747, opened 2019-03-11
JSON.parse() error on Object.destroy()


howardwkim:
parse@2.2.1
react-native@0.58.6

When I call `.destroy()`, I receive `SyntaxError: Unexpected token b in JSON at position 0`.
The object is destroyed on the server and the app still functions as expected, but I keep getting the error.

The error comes from a `JSON.parse()` located at `parse/lib/react-native/LocalDatastoreController.react-native.js:65`.
The key for `Parse/MY-PARSE-APP-ID/installationId` returns a string without curly brackets (like  `b94c0ed9-4a24-49af-b57b-d48e6a230db1` ).
 
Is this a bug in the code? Am I setting something incorrectly? Thanks!

Some of my code for setting up Parse:
```
import Parse from 'parse/react-native';
import { AsyncStorage } from 'react-native';
...

Parse.setAsyncStorage(AsyncStorage);
```

vrazn:
Getting the same error when calling `Parse.dumpLocalDatastore()`.
It looks like `async getAllContents()` also gets my Redux data stored in the Async storage when calls `const keys = await RNStorage.getAllKeys();` and fails to parse them later.

dplewis:
Thanks for reporting. Can you guys debug and submit a PR?

PR #753, opened 2019-03-20
LocalDatastore fixes for React-Native

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/747

Adds a new Prefix on stored object keys. This prevents retrieving and clearing unnecessary objects in the redux data store.

* Batches remove requests whenever possible
* Batches get requests whenever possible
* Improve Docs
* Improve Typings
* Clean up tests by removing hardcoded keys
* Add more tests
* Create Utils to remove dependency cycles
* Throws error if LDS is not enabled


src/LocalDatastore.js
```diff
@@ -13,10 +13,30 @@ import CoreManager from './CoreManager';
...
+import { DEFAULT_PIN, PIN_PREFIX, OBJECT_PREFIX } from './LocalDatastoreUtils';
 
-const DEFAULT_PIN = '_default';
-const PIN_PREFIX = 'parsePin_';
-
+/**
+ * Provides a local datastore which can be used to store and retrieve <code>Parse.Object</code>. <br />
+ * To enable this functionality, call <code>Parse.enableLocalDatastore()</code>.
+ *
+ * To store objects you can pin them.
```
acinader: use more formal.

"Pin object to add to local datastore"

src/LocalDatastore.js
```diff
@@ -13,10 +13,30 @@ import CoreManager from './CoreManager';
...
+ *
+ * <pre>await object.pin();</pre>
+ * <pre>await object.pinWithName('pinName');</pre>
+ *
+ * You can query your objects by changing the source of your query.
+ *
```
acinader: Query Object by changing query source

src/LocalDatastore.js
```diff
@@ -13,10 +13,30 @@ import CoreManager from './CoreManager';
...
+ * <pre>query.fromLocalDatastore();</pre>
+ * <pre>query.fromPin();</pre>
+ * <pre>query.fromPinWithName();</pre>
+ *
+ * Once your source is changed, you can query your objects from LocalDatastore
```
acinader: remove

src/LocalDatastore.js
```diff
@@ -313,15 +342,14 @@ const LocalDatastore = {
       await Promise.all(pinPromises);
       this.isSyncing = false;
     } catch(error) {
-      console.log('Error syncing LocalDatastore'); // eslint-disable-line
-      console.log(error); // eslint-disable-line
+      console.log('Error syncing LocalDatastore: ', error); // eslint-disable-line
```
acinader: we should just remove no console from the eslint config

src/LocalDatastoreController.default.js
```diff
@@ -8,34 +8,50 @@
...
  * @flow
  */
+import { isLocalDatastoreKey } from './LocalDatastoreUtils';
 
 const memMap = {};
 const LocalDatastoreController = {
-  fromPinWithName(name: string): ?any {
-    if (memMap.hasOwnProperty(name)) {
-      return memMap[name];
+  fromPinWithName(name: string): Promise {
```
acinader: is any of this async?  why returning promise??

src/LocalDatastoreController.default.js
```diff
@@ -8,34 +8,50 @@
...
+    if (!memMap.hasOwnProperty(name)) {
+      return Promise.resolve(null);
```
acinader: why resolving with null here but () below?

src/LocalDatastoreController.default.js
```diff
@@ -8,34 +8,50 @@
...
     }
-    return null;
+    const objects = JSON.parse(memMap[name]);
+    return Promise.resolve(objects);
   },
 
-  pinWithName(name: string, value: any) {
-    memMap[name] = value;
+  pinWithName(name: string, value: any): Promise {
```
acinader: same as above, is any of this async?

src/LocalDatastoreController.default.js
```diff
@@ -8,34 +8,50 @@
...
+    }
+    return Promise.resolve(LDS);
+  },
+
+  getRawStorage(): Promise {
+    return Promise.resolve(memMap);
   },
 
-  clear() {
+  clear(): Promise {
```
acinader: same

src/LocalDatastoreController.react-native.js
```diff
@@ -38,18 +39,52 @@ const LocalDatastoreController = {
...
-    return RNStorage.clear();
+  async getRawStorage(): Promise {
+    const keys = await RNStorage.getAllKeys();
+    const storage = {};
+    const results = await RNStorage.multiGet(keys);
+    results.map((pair) => {
+      const [key, value] = pair;
+      storage[key] = value;
+    });
+    return Promise.resolve(storage);
```
acinader: you can just return storage here as an async will auto wrap it.

src/LocalDatastoreController.react-native.js
```diff
@@ -38,18 +39,52 @@ const LocalDatastoreController = {
...
+    const results = await RNStorage.multiGet(batch);
+    results.map((pair) => {
+      const [key, value] = pair;
+      try {
+        LDS[key] = JSON.parse(value);
+      } catch (error) {
+        LDS[key] = null;
+      }
+    });
     return Promise.resolve(LDS);
```
acinader: can just return LDS async will make it a promise.

src/ParseObject.js
```diff
@@ -1237,15 +1240,18 @@ class ParseObject {
...
-      }
+    if (!localDatastore.isEnabled) {
+      return Promise.reject('Parse.enableLocalDatastore() must be called first');
+    }
+    const objectKey = localDatastore.getKeyForObject(this);
+    const pin = await localDatastore.fromPinWithName(objectKey);
+    if (pin) {
+      return Promise.resolve(true);
     }
     return Promise.resolve(false);
```
acinader: no need to wrap in promise in an async func

src/ParseObject.js
```diff
@@ -1291,20 +1299,23 @@ class ParseObject {
...
-      const pinned = await localDatastore._serializeObject(objectKey);
-      if (!pinned) {
-        throw new Error('Cannot fetch an unsaved ParseObject');
-      }
-      const result = ParseObject.fromJSON(pinned);
-      this._finishFetch(result.toJSON());
-
-      return Promise.resolve(this);
+    if (!localDatastore.isEnabled) {
+      return Promise.reject('Parse.enableLocalDatastore() must be called first');
```
acinader: just throw here

src/ParseObject.js
```diff
@@ -1291,20 +1299,23 @@ class ParseObject {
...
     }
+    const objectKey = localDatastore.getKeyForObject(this);
+    const pinned = await localDatastore._serializeObject(objectKey);
+    if (!pinned) {
+      throw new Error('Cannot fetch an unsaved ParseObject');
+    }
+    const result = ParseObject.fromJSON(pinned);
+    this._finishFetch(result.toJSON());
+
+    return Promise.resolve(this);
```
acinader: no need to wrap here.

src/ParseObject.js
```diff
@@ -1821,15 +1834,19 @@ class ParseObject {
...
-    if (localDatastore.checkIfEnabled()) {
-      for (const object of objects) {
-        await localDatastore._handlePinWithName(name, object);
-      }
+    if (!localDatastore.isEnabled) {
+      return Promise.reject('Parse.enableLocalDatastore() must be called first');
+    }
+    let promise = Promise.resolve();
+    for (const object of objects) {
+      promise = promise.then(() => localDatastore._handlePinWithName(name, object));
```
acinader: this doesn't seem ok.  I'd think you'd want:
```
const promises = objects.map(object =>  localDatastore._handlePinWithName(name, object);
return Promise.all(promises);
```

as you have it you just keep writing over promise.  Also not sure why you'd resolve a promise to just promise.then() and add a new promise....or i'm missing something...

src/ParseObject.js
```diff
@@ -1859,15 +1878,19 @@ class ParseObject {
...
-    if (localDatastore.checkIfEnabled()) {
-      for (const object of objects) {
-        await localDatastore._handleUnPinWithName(name, object);
-      }
+    if (!localDatastore.isEnabled) {
+      return Promise.reject('Parse.enableLocalDatastore() must be called first');
+    }
+    let promise = Promise.resolve();
+    for (const object of objects) {
+      promise = promise.then(() => localDatastore._handleUnPinWithName(name, object));
```
acinader: same as above....

src/StorageController.react-native.js
```diff
@@ -66,6 +66,30 @@ const StorageController = {
...
+    });
+  },
+
+  multiRemove(keys: Array<string>): Promise {
+    return new Promise((resolve, reject) => {
+      this.getAsyncStorage().multiRemove(keys, function(err) {
+        if (err) {
+          reject(err);
+        } else {
+          resolve();
```
acinader: return the keys?

src/LocalDatastoreController.react-native.js
dplewis: Thanks for clearing that up

src/ParseObject.js
dplewis: I tried that but data was getting overridden. I can try a better implementation

acinader: or 

```
const promises = objects.map(localDatastore._handlePinWithName.bind(null, name));
return Promise.all(promises);
```

:).

dplewis: I changed it to handle all pins instead of them individually, this allows batching the objects since they all have the same name pin.

src/LocalDatastore.js
```diff
@@ -13,10 +13,28 @@ import CoreManager from './CoreManager';
...
+/**
+ * Provides a local datastore which can be used to store and retrieve <code>Parse.Object</code>. <br />
+ * To enable this functionality, call <code>Parse.enableLocalDatastore()</code>.
+ *
+ * Pin object to add to local datastore
+ *
+ * <pre>await object.pin();</pre>
+ * <pre>await object.pinWithName('pinName');</pre>
+ *
+ * Query objects by changing query source
```
acinader: Query pinned objects 

(sorry)

src/LocalDatastore.js
```diff
@@ -38,44 +56,69 @@ const LocalDatastore = {
...
+    if (!objects || objects.length === 0) {
+      return;
+    }
     const pinName = this.getPinName(name);
-    const objects = this._getChildren(object);
-    objects[this.getKeyForObject(object)] = object._toFullJSON();
-    for (const objectKey in objects) {
-      await this.pinWithName(objectKey, objects[objectKey]);
+    const promises = [];
+    const objectKeys = [];
```
acinader: def not asking you to do it, but this would be a good use for reduce....

src/LocalDatastore.js
```diff
@@ -38,44 +56,69 @@ const LocalDatastore = {
...
+      const children = this._getChildren(parent);
+      const parentKey = this.getKeyForObject(parent);
+      children[parentKey] = parent._toFullJSON();
+      for (const objectKey in children) {
+        objectKeys.push(objectKey);
+        promises.push(this.pinWithName(objectKey, children[objectKey]));
+      }
     }
+    await Promise.all(promises);
     const pinned = await this.fromPinWithName(pinName) || [];
```
acinader: would be good to use 

```
const pinPromises = this.fromPinWithName(pinName) || promise.resolve([]);
const objectKeyPromises = Promise.all(promises);
const [objectKeys, pinned] = await Promises.all([objectKeyPromises, pinPromises]);
```
so that both operations happen concurrently.  As you have it, they go serially even though they are not dependent on each other.

disclaimer, my code above is likely just mostly right as it would take some more restructuring.

unfortunately, i think this should be fixed.

src/LocalDatastore.js
```diff
@@ -38,44 +56,69 @@ const LocalDatastore = {
...
-    const toPin = [...new Set([...pinned, ...objectIds])];
+    const toPin = [...new Set([...pinned, ...objectKeys])];
     await this.pinWithName(pinName, toPin);
   },
 
   // Removes object and children keys from pin name
   // Keeps the object and children pinned
-  async _handleUnPinWithName(name: string, object: ParseObject) {
+  async _handleUnPinAllWithName(name: string, objects: Array<ParseObject>) {
+    if (!objects || objects.length === 0) {
```
acinader: is this necessary?

src/LocalDatastore.js
dplewis: Just out of curiosity how would you do it with reduce :)

src/LocalDatastore.js
```diff
@@ -313,15 +360,14 @@ const LocalDatastore = {
       await Promise.all(pinPromises);
       this.isSyncing = false;
     } catch(error) {
-      console.log('Error syncing LocalDatastore'); // eslint-disable-line
-      console.log(error); // eslint-disable-line
+      console.error('Error syncing LocalDatastore: ', error);
```
acinader: do all browsers support console.error (that's a question, I dunno).

src/LocalDatastoreController.browser.js
```diff
@@ -40,14 +41,36 @@ const LocalDatastoreController = {
...
     for (let i = 0; i < localStorage.length; i += 1) {
       const key = localStorage.key(i);
-      const value = localStorage.getItem(key);
-      LDS[key] = JSON.parse(value);
+      if (isLocalDatastoreKey(key)) {
+        const value = localStorage.getItem(key);
+        LDS[key] = JSON.parse(value);
+      }
     }
     return Promise.resolve(LDS);
```
acinader: i'm confused still on where the promise is here?  why are you wrapping lds in a promise, why not just return LDS?

is it because this needs to have the same signature as another class of controller that does return promises?

src/LocalDatastoreController.react-native.js
```diff
@@ -10,15 +10,16 @@
...
 
 const RNStorage = require('./StorageController.react-native');
+import { isLocalDatastoreKey } from './LocalDatastoreUtils';
 
 const LocalDatastoreController = {
   async fromPinWithName(name: string): Promise {
     const values = await RNStorage.getItemAsync(name);
     if (!values) {
-      return Promise.resolve(null);
+      return null;
```
acinader: might be that what you want is `return []`

if the none base case is that an array of things come back?


if that is the case then the return flow type could be `Promise<array(Parse.Object)>` or the right way of saying that

src/ParseObject.js
```diff
@@ -1237,17 +1240,20 @@ class ParseObject {
    * <pre>
    * const isPinned = await object.isPinned();
    * </pre>
+   *
+   * @return {Promise} A boolean promise that is fulfilled if object is pinned.
```
acinader: if its a boolean, declare it as such: `Promise<boolean>`

src/ParseObject.js
```diff
@@ -1821,15 +1834,15 @@ class ParseObject {
    *
    * @param {String} name Name of Pin.
    * @param {Array} objects A list of <code>Parse.Object</code>.
+   * @return {Promise} A promise that is fulfilled when the pin completes.
    * @static
    */
-  static async pinAllWithName(name: string, objects: Array<ParseObject>): Promise {
+  static pinAllWithName(name: string, objects: Array<ParseObject>): Promise {
```
acinader: Array<Parse.Object>, right?

src/ParseQuery.js
```diff
@@ -1542,28 +1543,35 @@ class ParseQuery {
 
   /**
    * Changes the source of this query to all pinned objects.
+   *
+   * @return {Parse.Query} Returns the query, so you can chain this call.
    */
-  fromLocalDatastore() {
-    this.fromPinWithName(null);
+  fromLocalDatastore(): ParseQuery {
```
acinader: Parse.Query

src/ParseQuery.js
```diff
@@ -1542,28 +1543,35 @@ class ParseQuery {
...
   }
 
   /**
    * Changes the source of this query to a specific group of pinned objects.
+   *
+   * @param {String} name The name of query source.
+   * @return {Parse.Query} Returns the query, so you can chain this call.
    */
-  fromPinWithName(name: string) {
+  fromPinWithName(name: string): ParseQuery {
```
acinader: same

src/ParseObject.js
dplewis: Parse.Object doesn't exist only the current class `ParseObject`

src/ParseQuery.js
acinader: k.

src/LocalDatastore.js
dplewis: yeah same as warn and log

src/LocalDatastoreController.browser.js
dplewis: I just realized that flow typing isn't used in this project. I guess I just do it out of habit.

acinader: :)

src/LocalDatastoreController.react-native.js
dplewis: The value return is object, array, null.

I really should add flow typing to this SDK.

src/LocalDatastore.js
acinader: is this to the point:

https://runkit.com/arthur/reduce-example

src/LocalDatastoreController.browser.js
dplewis: I actually have flow for VS Code and it shows 300+ type errors in the project.

src/LocalDatastore.js
```diff
@@ -264,15 +261,16 @@ const LocalDatastore = {
         if (pinned.includes(objectKey)) {
           pinned = pinned.filter(item => item !== objectKey);
           if (pinned.length == 0) {
-            await this.unPinWithName(key);
+            promises.push(this.unPinWithName(key));
```
acinader: didn't catch this in the first pass, but yeah, if you have an await in a loop...

our lint in parse-server will complain.

acinader:
@dplewis lmk or assign to me when ready for review.

dplewis:
Its ready for review

acinader:
looking, but you'll need to rebase.

acinader:
let's see if i can just push the button to update.

src/.flowconfig
```diff
@@ -1,11 +1,10 @@
...
 
 [include]
-../package.json
 
 [libs]
-interfaces/
 
 [options]
-unsafe.enable_getters_and_setters=true
+suppress_comment= \\(.\\|\n\\)*\\@flow-disable-next
```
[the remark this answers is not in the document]
dplewis: linter should have caught that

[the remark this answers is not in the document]
acinader: linter is configured for the file extension and this file doesn't have one.

src/StorageController.react-native.js
```diff
@@ -66,6 +66,30 @@ const StorageController = {
     });
   },
 
+  multiGet(keys: Array<string>): Promise {
```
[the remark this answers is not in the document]
dplewis: Looking at the RN docs Promise<Array<Array<string>>>

src/LocalDatastoreController.default.js
```diff
@@ -8,31 +8,44 @@
...
+import { isLocalDatastoreKey } from './LocalDatastoreUtils';
 
 const memMap = {};
 const LocalDatastoreController = {
-  fromPinWithName(name: string): ?any {
-    if (memMap.hasOwnProperty(name)) {
-      return memMap[name];
+  fromPinWithName(name: string) {
+    if (!memMap.hasOwnProperty(name)) {
+      return null;
```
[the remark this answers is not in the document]
dplewis: You raised a lot good points. I've added those changes to make values consistent with arrays.

dplewis:
> Working this with you has been good. Hope it isn't too painful :).

We should do this again sometime

src/LocalDatastoreController.browser.js
```diff
@@ -10,44 +10,65 @@
...
-      LDS[key] = JSON.parse(value);
+      storage[key] = value;
     }
-    return Promise.resolve(LDS);
+    return storage;
   },
 
-  clear(): Promise {
-    return Promise.resolve(localStorage.clear());
+  clear(): void {
```
acinader: you sure you don't want to return a promise here?  if local.Storage.removeItem(key); fails, is it going to crash?

```
clear(): Promise {
  ....
  const promises = toRemove.map(localStorage.removeItem);
  return Promise.all(promises);
}
```

src/LocalDatastoreController.react-native.js
```diff
@@ -27,29 +28,63 @@ const LocalDatastoreController = {
...
+    return LDS;
   },
 
-  clear(): Promise {
-    return RNStorage.clear();
+  async getRawStorage(): Promise {
+    const keys = await RNStorage.getAllKeys();
+    const storage = {};
+    const results = await RNStorage.multiGet(keys);
+    results.map((pair) => {
```
acinader: can use forEach here instead of map cause you're not capturing the returned array.

map returns an array with one element per element in the original array
forEach return nothing.

so forEach is more memory efficient if you're not returning something.

map composes very nicely with filter.

src/LocalDatastoreController.react-native.js
```diff
@@ -27,29 +28,63 @@ const LocalDatastoreController = {
...
+      } catch (error) {
+        LDS[key] = null;
+      }
+    });
+    return LDS;
   },
 
-  clear(): Promise {
-    return RNStorage.clear();
+  async getRawStorage(): Promise {
```
acinader: could `Promise<object>`

src/LocalDatastore.js
```diff
@@ -38,44 +56,62 @@ const LocalDatastore = {
...
     }
-    const pinned = await this.fromPinWithName(pinName) || [];
-    const objectIds = Object.keys(objects);
-    const toPin = [...new Set([...pinned, ...objectIds])];
-    await this.pinWithName(pinName, toPin);
+    const fromPinPromise = this.fromPinWithName(pinName);
+    const [pinned] = await Promise.all([fromPinPromise, toPinPromises]);
+    const toPin = [...new Set([...(pinned || []), ...objectKeys])];
+    return this.pinWithName(pinName, toPin);
   },
```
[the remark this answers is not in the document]
dplewis: I tried that but for some reason to get the toPins I would have to call `Promise.all` twice

src/LocalDatastoreController.browser.js
dplewis: I don't see how removeItem would throw an error, based on the documentation.

src/LocalDatastore.js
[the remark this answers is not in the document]
acinader: I'd have to be able to step through this with a debugger to see.

One thing I'm curious about is why `this.pinWithName(objectKey, [toPin[objectKey]])) `

I'm assuming that `toPin[objectKey]` is already an array, so not sure why it needs to be put in an array.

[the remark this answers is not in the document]
dplewis: getChildren returns key -> object and not key -> [object]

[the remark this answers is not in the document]
dplewis: Having another pair of eyes debugging this is really helpful

yelounak:
@dplewis I just want to say thank you and the whole Parse contributors for your efforts!
Really appreciated.

PR #754, opened 2019-03-20
Remove flovilmart's email from COC

Tom-Fox1:
As flovilmart is no longer a core maintainer his email is being removed in favour of admin@parseplatform.org (as agreed with @acinader).


Tom-Fox1:
Changed email to codeofconduct@parseplatform.org

PR #758, opened 2019-03-22
Fixes LiveQuery behavior on multiple subscriptions

### LINKED ISSUE #278: Live queries, multiple subscription at once, is it me or it's not implemented at the moment?
Hello everyone, I'm using the **JS-SDK(v1.8.4)** with **Parse-server(v2.2.9)**, I tried to `subscribe` to multiple queries at the same app page, I've noticed that for example:

If I `subscribe` only to Query A, I can later `unsubscribe`(as in using the method) from Query A and all is good(no further events will emitted), but If I `subscribe` to Query A, then I again `subscribe` to Query B, and now I want to `unsubscribe`from Query A, **I can't**, events will still emitted.

If I `subscribe` to A then B, now I `unsubscribe` from B, no no further events will emitted for B, but I can't unsubscribe from A, I can call unsubscribe but events for A will emitted.

I don't know if this supposed to work like that(not implemented feature), anyway I just wanted to know if its a valid behavior?, for now the **workaround** is to delete the event from the subscription object, for example:

```
let subscription = (new Parse.Query('ObjectName')).subscribe();
subscription.on('create', data => {})
delete subscription ._events.create;
```

In the A, B case, you could use `unsubscribe` for B and `delete` for A.

**Update:**

To be more specific, if I have multiple queries subscribed to the same class, I can `unsubscribe` from both of them successfully, so it reduces the problem to a case where I have 2 queries with 2 different classes, if ever implemented?

Amphaal:
Fixes #278

### DIFF
--- a/src/CoreManager.js
+++ b/src/CoreManager.js
@@ -393,10 +393,9 @@ module.exports = {
 
   setLiveQueryController(controller: any) {
     requireMethods('LiveQueryController', [
-      'subscribe',
-      'unsubscribe',
-      'open',
-      'close',
+      'setDefaultLiveQueryClient',
+      'getDefaultLiveQueryClient',
+      '_clearCachedDefaultClient',
     ], controller);
     config['LiveQueryController'] = controller;
   },
--- a/src/LiveQuerySubscription.js
+++ b/src/LiveQuerySubscription.js
@@ -110,6 +110,7 @@ class Subscription extends EventEmitter {
     return CoreManager.getLiveQueryController().getDefaultLiveQueryClient().then((liveQueryClient) => {
       liveQueryClient.unsubscribe(this);
       this.emit('close');
+      return Promise.resolve(this);
     });
   }
 }
--- a/src/LiveQuerySubscription.js
+++ b/src/LiveQuerySubscription.js
@@ -101,16 +101,19 @@ class Subscription extends EventEmitter {
     this.id = id;
     this.query = query;
     this.sessionToken = sessionToken;
+
+    // adding listener so process does not crash
+    // best practice is for developer to register their own listener
+    this.on('error', () => {});
   }
 
   /**
-   * closes the subscription
+   * Closes the subscription
    */
-  unsubscribe() {
-    return CoreManager.getLiveQueryController().getDefaultLiveQueryClient().then((liveQueryClient) => {
-      liveQueryClient.unsubscribe(this);
-      this.emit('close');
-    });
+  async unsubscribe(): Promise {
+    const liveQueryClient = await CoreManager.getLiveQueryController().getDefaultLiveQueryClient();
+    liveQueryClient.unsubscribe(this);
+    this.emit('close');
   }
 }
 
--- a/src/LiveQuerySubscription.js
+++ b/src/LiveQuerySubscription.js
@@ -110,7 +110,6 @@ class Subscription extends EventEmitter {
     return CoreManager.getLiveQueryController().getDefaultLiveQueryClient().then((liveQueryClient) => {
       liveQueryClient.unsubscribe(this);
       this.emit('close');
-      return Promise.resolve(this);
     });
   }
 }
--- a/src/LiveQuerySubscription.js
+++ b/src/LiveQuerySubscription.js
@@ -108,12 +108,13 @@ class Subscription extends EventEmitter {
   }
 
   /**
-   * Closes the subscription
+   * Close the subscription
    */
-  async unsubscribe(): Promise {
-    const liveQueryClient = await CoreManager.getLiveQueryController().getDefaultLiveQueryClient();
-    liveQueryClient.unsubscribe(this);
-    this.emit('close');
+  unsubscribe(): Promise {
+    return CoreManager.getLiveQueryController().getDefaultLiveQueryClient().then((liveQueryClient) => {
+      liveQueryClient.unsubscribe(this);
+      this.emit('close');
+    });
   }
 }
 
--- a/src/ParseLiveQuery.js
+++ b/src/ParseLiveQuery.js
@@ -12,6 +12,7 @@
 import EventEmitter from './EventEmitter';
 import LiveQueryClient from './LiveQueryClient';
 import CoreManager from './CoreManager';
+const url = require('url');
 
 function getLiveQueryClient(): LiveQueryClient {
   return CoreManager.getLiveQueryController().getDefaultLiveQueryClient();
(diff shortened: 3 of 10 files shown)


dplewis:
This is awesome! Thank you for the PR.

Test cases are needed. Check our the [Contributors Guide](
https://github.com/parse-community/Parse-SDK-JS/blob/master/CONTRIBUTING.md)

I can test this in production. I would like this in the next release.

My only concern is breaking changes

Amphaal:
As I peeked into the existing tests for LiveQuery, I'm a bit concerned that my PR is still a hack in it's current state. 

ParseLiveQuery being a mere proxy class for LiveQueryClient, i feel like it would be more relevant to implement the Promise-based workflow directly in the LiveQueryClient class, resolving promises after the server acknoledges any instructions. 

You are right to assume it might also break things for certain users, but I wonder how any developer could have been able to use the LQ functionality with the framework so far, considering how restricted the working usecase is. 

Maybe feedback from developers with actual experience with this functionality is required beforehand ? IMO, Promises are mandatory.

dplewis:
@Amphaal I added a few tests to this. Can you check them and add anything you think is missing.

I agree with making it promise based. Even unsubscribe is promised based and isn't documented as such. We can go over this feature with a fine tooth comb and clean up where possible.

@acinader  thoughts?

src/ParseQuery.js
```diff
@@ -1481,7 +1481,7 @@ class ParseQuery {
    * @return {LiveQuerySubscription} Returns the liveQuerySubscription, it's an event emitter
    * which can be used to get liveQuery updates.
    */
-  subscribe(): any {
+  subscribe(): Promise {
```
dplewis: Currently its returning Promise<EventEmitter> but the documentation above states different.

src/ParseLiveQuery.js
```diff
@@ -201,12 +201,13 @@ const DefaultLiveQueryController = {
         subscription.on('error', (object) => {
           subscriptionWrap.emit('error', object);
         });
+
+        return subscriptionWrap;
```
dplewis: I've always been curious as to why the subscription needs to be wrapped.

I believe we could erase it all just do 

```
// Prevent crashing
subscription.on('error' () => {})
return subscription;
```

And document the Error Event (currently undocumented)

acinader: I'll take a look.

dplewis:
@Amphaal @acinader I did some clean up;

I changed the LiveQueryController to only handles getting / setting the LiveQueryClient.

subscribe logic moved to query.subscribe

This makes it much easier to understand how LiveQuery works.

src/LiveQuerySubscription.js
```diff
@@ -101,16 +101,19 @@ class Subscription extends EventEmitter {
...
+   * Closes the subscription
    */
-  unsubscribe() {
-    return CoreManager.getLiveQueryController().getDefaultLiveQueryClient().then((liveQueryClient) => {
-      liveQueryClient.unsubscribe(this);
-      this.emit('close');
-    });
+  async unsubscribe(): Promise {
+    const liveQueryClient = await CoreManager.getLiveQueryController().getDefaultLiveQueryClient();
+    liveQueryClient.unsubscribe(this);
```
dplewis: Should we wait for this to unsubscribe or just let it fire off in the background?

```
this.connectPromise.then(() => {
  this.socket.send(JSON.stringify(unsubscribeRequest));
});
```
I don't see any issues with the client itself but wanted feedback because `this.connectPromise.then()` is used a lot in the client.

Amphaal: In my old prototype, I waited for all subs to unsubscribe, then called Parse.LiveQuery.close(), I guess it was useless after all, since Parse.LiveQuery.close() is suposed to do it by itself.

src/CoreManager.js
```diff
@@ -393,10 +393,9 @@ module.exports = {
 
   setLiveQueryController(controller: any) {
     requireMethods('LiveQueryController', [
```
acinader: can you help me understand what this change is please?

src/LiveQuerySubscription.js
```diff
@@ -101,16 +101,19 @@ class Subscription extends EventEmitter {
...
     this.sessionToken = sessionToken;
+
+    // adding listener so process does not crash
+    // best practice is for developer to register their own listener
+    this.on('error', () => {});
   }
 
   /**
-   * closes the subscription
+   * Closes the subscription
```
acinader: Close

src/LiveQuerySubscription.js
acinader: I'd think you'd wait for the promise to resolve before firing the event, unless there's a goog reason not to?

src/ParseLiveQuery.js
```diff
@@ -57,158 +51,84 @@ const LiveQuery = new EventEmitter();
...
-        CoreManager.set('LIVEQUERY_SERVER_URL', liveQueryServerURL);
+    // If we can not find Parse.liveQueryServerURL, we try to extract it from Parse.serverURL
+    if (!liveQueryServerURL) {
+      const tempServerURL = CoreManager.get('SERVER_URL');
+      let protocol = 'ws://';
+      if (tempServerURL.indexOf('https') === 0) {
+        protocol = 'wss://'
       }
+      const host = tempServerURL.replace(/^https?:\/\//, '');
+      liveQueryServerURL = protocol + host;
```
acinader: 106,7 could use the node url lib...

src/CoreManager.js
dplewis: Those function (subscribe, unsubscribe, open, close) already exists in the LiveQueryClient.

In the LiveQueryController those were just wrapper functions. Instead of using wrapper functions call the ones from the LiveQueryClient.

setDefaultLiveQueryClient, getDefaultLiveQueryClient, _clearCachedDefaultClient already existed but weren't documented.

src/LiveQuerySubscription.js
dplewis: I'm going to revert this function. Since a socket is being used we have no way to tell if it went through or not (it will eventually). 

As I have it, the caller is now forced to wait for the unsubscribe to finish (which wasn't the case before).

src/ParseLiveQuery.js
```diff
@@ -98,13 +99,9 @@ const DefaultLiveQueryController = {
...
-      const tempServerURL = CoreManager.get('SERVER_URL');
-      let protocol = 'ws://';
-      if (tempServerURL.indexOf('https') === 0) {
-        protocol = 'wss://'
-      }
-      const host = tempServerURL.replace(/^https?:\/\//, '');
-      liveQueryServerURL = protocol + host;
+      const serverURL = url.parse(CoreManager.get('SERVER_URL'));
+      const protocol = serverURL.protocol === 'http:' ? 'ws://' : 'wss://';
+      liveQueryServerURL = protocol + serverURL.host + serverURL.pathname;
```
acinader: hot!

dplewis:
@Amphaal Thanks for opening the PR. Sorry if I kinda took it over. We can look more deeply into the LiveQueryClient to see if there are improvements to be made.

src/ParseLiveQuery.js
acinader: uh @dplewis...wait a sec...is this on the client?  will the `url` class be available?

I'm so used to all the parse stuff being on the server, I get a little confused on what's where when it comes to the JS SDK...😬

Amphaal:
@dplewis You did great, I honestly don't have much time right now to dig back into Parse... This fix is a great addition to future release !
But I was happy to help, even in the smallest way. I hope this will encourage users to go further with LQ :)

src/ParseLiveQuery.js
dplewis: I forgot about that too. I keep forgetting about other environments besides node.

To answer your question I don't think so. I'll revert it in a PR.

acinader: or https://www.npmjs.com/package/url-polyfill

the code is sooooo much better this way.

acinader: https://developer.mozilla.org/en-US/docs/Web/API/URL

not sure what compatibility we're looking for.

PR #760, opened 2019-03-26
Add Issue Templates

dplewis:


Tom-Fox1:
@dplewis I like the getting help template - perhaps it could also say use the `parse-platform` and `parse-javascript-sdk` tags on stack overflow for code level questions.

We could add this to other repos too.

PR #764, opened 2019-03-27
add email address for npm

acinader:


dplewis:
Adding an environment variable `$NPM_EMAIL` would be easier to maintain.

acinader:
righto.

acinader:
hm. actually. after looking at it, i'm going to disagree.  this is public and is not a secret, why the indirection?  this way anyone can a) see what it is and b) change it if needed through the pr process....

what do you htink?

dplewis:
Thats what we have in parse-server. 

> this way anyone can a) see what it is

Without this PR I would have never known the email

> change it if needed through the pr process

No PR needed just change it in travis settings.

I'll leave it up to you

acinader:
my point would be that everything that isn't a secret should go through pr so it is visible to the community. 

so my take would be that if it aint a secret and needs to be a secret, like publishing tokens and such, it should go through pr and be in a git repo

if you leave it up to me, i'm good as it is.

acinader:
i'll move the parse-server and conform it to the same email address for the forthcoming release too.

PR #765, opened 2019-03-27
(Browser) Support for File Uri Upload

dplewis:
`http` and `https` modules don't exist in browser

### DIFF
--- a/src/ParseFile.js
+++ b/src/ParseFile.js
@@ -317,6 +317,9 @@ const DefaultController = {
     if (XHR) {
       return this.downloadAjax(uri);
     }
+    if (process.env.PARSE_BUILD !== 'node') {
+      return Promise.reject('Cannot make a request: No definition of XMLHttpRequest was found.');
+    }
     return new Promise((resolve, reject) => {
       let client = require('http');
       if (uri.indexOf('https') === 0) {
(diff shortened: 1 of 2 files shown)


src/ParseFile.js
```diff
@@ -317,6 +317,9 @@ const DefaultController = {
     if (XHR) {
       return this.downloadAjax(uri);
     }
+    if (process.env.PARSE_BUILD !== 'node') {
```
acinader: maybe i don't understand what process.env.PARSE_BUILD is, but the way I'm reading this is:

if we're not in node, complain that we don't have XMLHttpRequest, but XMLHttpRequest is in the browser, not node.  set me straight here :).

dplewis: For browsers that don't support ajax. 

https://www.tutorialspoint.com/ajax/ajax_browser_support.htm

dplewis: PARSE_BUILD is set when building the sdk. 

https://github.com/parse-community/Parse-SDK-JS/blob/master/build_releases.sh

src/ParseFile.js
```diff
@@ -311,10 +314,16 @@ const DefaultController = {
   },
 
   download: function(uri) {
+    if (XHR) {
```
acinader: I think I get it.  Fine as is.

re-writting below how I would formultate.  You think this is functionally the same?

no need to change, just curious.
```
if(process.env.PARSE_BUILD === 'browser') {
  if(!XHR) {
     return Promise.reject('Cannot make a request: No definition of XMLHttpRequest was found.');
  }

  return return this.downloadAjax(uri);
}
// We're in node
return new Promise((resolve, reject) => {
    const client = uri.indexOf('https') === 0 
        ? require('https')
        : require('http');
...
```

dplewis: Functionality will be the same

dplewis: Actually xmlhttprequest exists in react-native

PR #767, opened 2019-03-29
Add options to _linkWith(provider) for masterKey

acinader:
1. without the master key it is not possible to link a user with an auth provider in cloud code
2. remove a recent addition to pass the sessionToken which was added for anonymous auth, but not necessary and causing problems.  Also remove the related test.

### DIFF
--- a/integration/server.js
+++ b/integration/server.js
@@ -1,22 +1,27 @@
 const express = require('express');
 const ParseServer = require('parse-server').ParseServer;
 const app = express();
+const CustomAuth = require('./test/CustomAuth');
 
-// Specify the connection string for your mongodb database
-// and the location to your Parse cloud code
 const api = new ParseServer({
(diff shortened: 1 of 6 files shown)


acinader:
@dplewis parse-server unit tests, my cloud code unit tests and the specific issue I was having (linking a user to a custom OAuth provider in cloud) all work with these changes, but I want to cover and provide passing tests with the master key and test that throw when there is no master key, but I am having trouble getting it to work.  gonna sleep on it.  If you see anything obvious or have any guidance, I'm all ears :).

dplewis:
@acinader Thanks for getting this started. I fixed the tests. 

I realized the docs needs to be updated for this. [Unlinking](https://docs.parseplatform.org/js/guide/#unlinking)

On the client side - you can link with current user / sessionToken
On server side - since current user doesnt exist you have to use master key / sessionToken

PR #769, opened 2019-03-29
Add two unit tests for linking

acinader:
1. test loggin in with the js sdk to verify documentation change
2. add a negative test that proves we need the master key to link


dplewis:
LGTM! The more tests the better

Issue #773, opened 2019-03-30
Failed to get object from local datastore


FahamuTech:
### Issue Description

When try to get pinned object from local datastore, results in JSON.parse error 

### Steps to reproduce

1. Pin object to local datastore
2. Query object from local datastore

### Expected Results

Expected list of object from local datastore

### Actual Outcome

SyntaxError: Unexpected token b in JSON at position 0
    at JSON.parse (<anonymous>)
    at Object.getAllContents (LocalDatastoreController.browser.js:59)
    at Object._getAllContents (LocalDatastore.js:61)
    at Object.<anonymous> (LocalDatastore.js:313)
    at tryCatch (runtime.js:62)
    at Generator.invoke [as _invoke] (runtime.js:288)
    at Generator.prototype.(:4200/anonymous function) [as next] 

### Environment Setup

- **Server**
  - parse-server version: 3.1.3
  - Operating System:    Ubuntu Desktop 18.04
  - Localhost or remote server? localhost

- **JS SDK**
  - JS SDK version: 2.2.1
  - Application? : Browser

### Logs/Trace
SyntaxError: Unexpected token b in JSON at position 0
    at JSON.parse (<anonymous>)
    at Object.getAllContents (LocalDatastoreController.browser.js:59)
    at Object._getAllContents (LocalDatastore.js:61)
    at Object.<anonymous> (LocalDatastore.js:313)
    at tryCatch (runtime.js:62)
    at Generator.invoke [as _invoke] (runtime.js:288)
    at Generator.prototype.(:4200/anonymous function) [as next] (http://localhost:4200/vendor.js:142466:21)
    at asyncGeneratorStep (asyncToGenerator.js:3)
    at _next (asyncToGenerator.js:25)
    at asyncToGenerator.js:32

FahamuTech:
I found out in localstorage of browser  there is an installationId which is not JSON object.
So i edit LocalDatastoreController( js file of parse sdk from nodemodule ) : getAllContents (LocalDatastoreController.browser.js:59). i rewrite `LDS[key] = JSON.parse(value);` with `try {
        LDS[key] = JSON.parse(value);
       } catch (e) {
         console.warn('Parse localdatastore warn: key==>' + key+ ', value==>'+ value+ ', error ==>'+ e )
      }`. to ignore installationId of the device

dplewis:
@FahamuTech I can submit a PR. Can you post the exact payload that threw the error.

PR #774, opened 2019-03-30
Update LocalDatastoreController.browser.js

FahamuTech:
solve issue #773 of JSON.parse error when try to get pinned objects


dplewis:
Thanks for the fix. Can you write a test case?

FahamuTech:
Yes I will write a test.

FahamuTech:
i am closing  a pull to write a test and i will open again

PR #776, opened 2019-04-01
Fix #775

### LINKED ISSUE #775: The package at "node_modules/parse/lib/react-native/ParseFile.js" attempted to import the Node standard library module "https"
### Issue Description
With changes come with #765 causes this issue. There was a similiar issue caused by `http` package. Please see #768 . It supposed to be fixed but now i have this.

There is no problem with version `2.2.1`.

### Steps to reproduce

Try to use `parse/react-native` with version >= 2.3

### Actual Outcome

> The package at "node_modules/parse/lib/react-native/ParseFile.js" attempted to import the Node standard library module "https". It failed because React Native does not include the Node standard library.

### Environment Setup

- **Server**
  - parse-server version: `2.8.4`
  - Operating System:     [FILL THIS OUT]
  - Hardware:             [FILL THIS OUT]
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): `Back4App`

- **JS SDK**
  - JS SDK version: `2.3.1`
  - Application? (Browser, Node, React-Native, etc): `React Native`

dplewis:
Closes: #775 

Prevent modules from compiling in react-native.

@alioguzhan Can you check if this fix works for you?

### DIFF
--- a/src/ParseFile.js
+++ b/src/ParseFile.js
@@ -316,11 +316,7 @@ const DefaultController = {
   download: function(uri) {
     if (XHR) {
       return this.downloadAjax(uri);
-    }
-    if (process.env.PARSE_BUILD === 'browser' && !XHR) {
-      return Promise.reject('Cannot make a request: No definition of XMLHttpRequest was found.');
-    }
-    if (process.env.PARSE_BUILD === 'node') {
+    } else if (process.env.PARSE_BUILD === 'node') {
       return new Promise((resolve, reject) => {
         const client = uri.indexOf('https') === 0
           ? require('https')
@@ -337,6 +333,8 @@ const DefaultController = {
           });
         }).on('error', reject);
       });
+    } else {
+      return Promise.reject('Cannot make a request: No definition of XMLHttpRequest was found.');
     }
   },
 
--- a/src/ParseFile.js
+++ b/src/ParseFile.js
@@ -317,25 +317,27 @@ const DefaultController = {
     if (XHR) {
       return this.downloadAjax(uri);
     }
-    if (process.env.PARSE_BUILD === 'browser') {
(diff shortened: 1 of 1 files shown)


src/ParseFile.js
```diff
@@ -317,25 +317,27 @@ const DefaultController = {
     if (XHR) {
       return this.downloadAjax(uri);
     }
-    if (process.env.PARSE_BUILD === 'browser') {
+    if (process.env.PARSE_BUILD === 'browser' && !XHR) {
```
acinader: you could just change the if to and `else if` and drop the `&& !XHR`

src/ParseFile.js
```diff
@@ -317,25 +317,27 @@ const DefaultController = {
...
-        : require('http');
-      client.get(uri, (resp) => {
-        resp.setEncoding('base64');
-        let base64 = '';
-        resp.on('data', (data) => base64 += data);
-        resp.on('end', () => {
-          resolve({
-            base64,
-            contentType: resp.headers['content-type'],
+    if (process.env.PARSE_BUILD === 'node') {
```
acinader: is it possilbe for this if to be false and neither of the two ifs above to be true?

if so, how to handle?  if not, then we shouldn't need this test?

dplewis: The first if handles react-native and browser
The second for unsupported browsers
The third is for node only. (Note babel trick here. This won't end up in the react-native build)

You can check by forcing this branch. `npm run build` and check `/lib/xxx/ParseFile.js` and notice the difference

alioguzhan:
Thanks for the quick fix! I will check it today and will let you know. @dplewis

alioguzhan:
Sorry @dplewis , I couldn't have the chance to test this fix. But looks fine to me.  👍

PR #777, opened 2019-04-02
Fixed object save method params documentation

pocketcolin:
The documentation for the params required for Object's save method describes 1 param (options), but you cannot actually pass just options as an object to the save method. The correct use of the save method is described with examples in the documentation, but the additional 2 params are not documented. As the save method does accept 3 arguments (in the code as value: function (arg1, arg2, arg3) {), these params need to be added to the documentation and not just mentioned in the examples.

### DIFF
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -1081,6 +1081,8 @@ class ParseObject {
    * You can either call it as:<pre>
    *   object.save();</pre>
    * or<pre>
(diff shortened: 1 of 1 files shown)


pocketcolin:
https://github.com/parse-community/parse-server/issues/5456

pocketcolin:
Fixes https://github.com/parse-community/parse-server/issues/5456

Issue #678, opened 2018-10-16
Fetch and getData on Parse.File - FEATURE REQUEST


vince1995:
Hey,

would be great to run `.fetch()` and `.data()` on `Parse.File`.

The background is that I have some data which I save in `Parse.File`s. But to cache (before save and after download) them I need to have it in an extra value or overwrite the `get()` function of `Parse.Object` (or my class which extends `Parse.Object`).

Something like the `.getData()` function in the [Parse SDK for Android](https://parseplatform.org/Parse-SDK-Android/api/com/parse/ParseFile.html#getData--).

vince1995:
Thanks, that's exactly what I've been loking for. :)

PR #780, opened 2019-04-02
Parse.File getData

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/678

Adds similar functionality for `.getData()` function in the [Parse SDK for Android](https://parseplatform.org/Parse-SDK-Android/api/com/parse/ParseFile.html#getData--) and [Parse SDK for PHP](http://parseplatform.org/parse-php-sdk/classes/Parse.ParseFile.html#method_getData)

Returns base64 (not full base64 with `data:...`)


src/ParseFile.js
```diff
@@ -189,10 +212,19 @@ class ParseFile {
...
-        this._previousSave = controller.saveUri(this._name, this._source, options).then((res) => {
+        this._previousSave = controller.download(this._source.uri).then((result) => {
+          const newSource = {
+            format: 'base64',
+            base64: result.base64,
+            type: result.contentType,
+          };
+          this._data = result.base64;
+          return controller.saveBase64(this._name, newSource, options);
+        }).then((res) => {
```
acinader: I know it isn't your code, but I'm surprised that this function doesn't always return a promise!  

if !this._previousSave, then this wont return a promise and anything that then's this func will error.

maybe in practive there is always a previous state when this is called???

dplewis: I don't know why thats there. Been there since the beginning of the SDK. But this always returns a Promise.

dplewis: Want me to clean it up or leave it?

acinader: definitely your call

the class could definitely use some factoring out in any number of places (base64 comma location stuff for instance), but your changes look good to me and it works as is.

PR #790, opened 2019-04-11
Improve Flow Layout

dplewis:
I believe we can break this up into multiple PR's

This PR covers half of the flow errors (out of 375) without functionality changes.

The next half will need to be reviewed with a fine tooth comb


dplewis:
I removed it in this case. Its used when you can't figure out flow type errors or there are too many to handle at the present moment.

Issue #778, opened 2019-04-02
LiveQueryClient subscribe with token - no updates to ACL secured objects


kirilledelman:
### Issue Description

Live query subscription to an ACL secured object doesn't work, even when sessionToken is passed.

### Steps to reproduce

```
// user is logged in, there's a Patient object in the DB, whose ACL is set to Read+Write by same user
var lq = Parse.CoreManager.getLiveQueryController();
var cu = Parse.User.current();
var pq = new Parse.Query( 'Patient' );
pq.equalTo( 'userId', cu.id );
var plq = lq.subscribe( pq, cu.getSessionToken() );
plq.on('update', (object) => {
	console.log('object updated', object );
});
```
Update a field in Patient object on the server.

### Expected Results

"object updated" handler should be called

### Actual Outcome

nothing happens.

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : ^3.1.3
  - Operating System:     CentOS
  - Hardware:            AWS instance
  - Localhost or remote server? AWS remote server

- **JS SDK**
  - JS SDK version: 2.2.1
  - Application: Chrome

dplewis:
Improvements were made to LiveQuery https://github.com/parse-community/Parse-SDK-JS/pull/758/files

Can you update to the latest version?

If it still doesn't work can you write a failing test? There are examples in the PR if you need it.

`Note: await query.subscribe() gets the current user session internally`

kirilledelman:
@dplewis, I've updated my JS SDK to the latest PR and server side `parse-server` node packages, but I'm still getting the same behavior. I can see that `Subscription` object I get from subscribing to a query has the correct `sessionToken` property set. But when I subscribe to a query that returns an object whose ACL is set to the current user, I get no `update` event. ACL set to public read+write posts update events correctly.

So this must be the server side problem, right?

dplewis:
There is a issue opened on the server side https://github.com/parse-community/parse-server/issues/5393

Would you like to provide a fix?

kirilledelman:
I don't think #5393 is the same to what I'm experiencing. I'm not using roles and my object is secured with an ACL containing a single user id, and I'm not getting updates. Unfortunately, I'm not familiar enough with Parse project to be able to fix this. I was hoping for a known workaround.

dplewis:
Can you write a failing test?

dplewis:
[...]

PR #791, opened 2019-04-11
LiveQuery subscribe with sessionToken

dplewis:
Adds sessionToken parameter to `query.subscribe()`
As described in the guide for Advanced API https://docs.parseplatform.org/js/guide/#sessiontoken

This is also a failing test for https://github.com/parse-community/Parse-SDK-JS/issues/778

### DIFF
--- a/integration/test/ParseLiveQueryTest.js
+++ b/integration/test/ParseLiveQueryTest.js
@@ -137,7 +137,7 @@ describe('Parse LiveQuery', () => {
     assert.equal(count, 1);
   });
 
-  it('can subscribe to ACL', async () => {
+  it('can subscribe to ACL', async (done) => {
     const user = await Parse.User.signUp('ooo', 'password');
     const ACL = new Parse.ACL(user);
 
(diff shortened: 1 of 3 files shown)


integration/test/ParseLiveQueryTest.js
```diff
@@ -135,4 +136,26 @@ describe('Parse LiveQuery', () => {
...
+    const query = new Parse.Query(TestObject);
+    query.equalTo('objectId', object.id);
+    const subscription = await query.subscribe(user.getSessionToken());
+    let count = 0;
+    subscription.on('update', object => {
+      count++;
+      assert.equal(object.get('foo'), 'bar');
+    })
+    await object.save({ foo: 'bar' });
+    await sleep(1000);
```
acinader: I don't think we should introduce second long waits if we can avoid it.

you could use done in this case, and in the update handler, you can do all the asserts, logout and then use done.

acinader: i know that this is just modeling other tests, but i don't like introducing another second wait

PR #794, opened 2019-04-13
Support for Wechat applet

einsqing:
Support for Wechat applet


einsqing:
- [x] wx.Parse

acinader:
@einsqing is there any reason you can't use wx and fly outside of parse entirely?  

I'm not sure how wechat app works.  Is it possible that you can manipulate the request with something like: ```$.ajaxPrefilter(function(options, originalOptions, jqXHR) {``` that would allow you to use parse without having to make the changes here?


I don't know what flyio is or what it does.

einsqing:
微信小程序发送请求参考如下，不能使用xhr，另外使用flyio是因为支持微信小程序，并且可以使用拦截器

The reference for sending requests from the Wechat miniprograme is as follows: XHR cannot be used, and flyio is used because it supports the Wechat miniprograme and interceptors can be used.

```
wx.request({
  url: 'test.php', // 仅为示例，并非真实的接口地址
  data: {
    x: '',
    y: ''
  },
  header: {
    'content-type': 'application/json' // 默认值
  },
  success(res) {
    console.log(res.data)
  }
})
```

PR #803, opened 2019-04-24
LDS: Improve Error Handling

dplewis:
Fixes: https://github.com/parse-community/Parse-SDK-JS/issues/773

Should we log key, values?

Console.log vs warn vs error?


acinader:
test failing

dplewis:
@acinader After much trial and error, I discovered with the latest jest / jsdom you can't spy after mocking localStorage. It is already set in window.localStorage by jest. No need to mock it

Issue #819, opened 2019-05-09
the build system is only for mac/linux, not Windows


oallouch:
Hi,
I'd like to contribute to this project, but I code on Windows.
What can I do ?

Thx,
Olivier

dplewis:
You are right, I'm looking into it right now.

PR #820, opened 2019-05-09
Support testing on Windows

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/819

* Add .gitattributes file to default to LF line ending
* Add cross-env to handle env variables
* Default watch to browser and added the following commands:

`npm run watch:node`
`npm run watch:browser`
`npm run watch:react-native`

@oallouch Can you try out this branch? I believe with watch, test, and integration should be enough to start contributing.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,15 @@
+* text=auto eol=lf
+
+*.js text
+*.html text
+*.less text
+*.json text
+*.css text
+*.xml text
+*.md text
+*.txt text
+*.yml text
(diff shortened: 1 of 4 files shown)


oallouch:
Test and watch work perfectly.
My lib folder gets updated.
ReadPreference, here I come !

dplewis:
@oallouch Glad it worked. Will try to get this in the next release.

@acinader How does this look?

PR #837, opened 2019-06-13
Allow any Blob to be uploaded

LinusU:
`File` is a subclass of `Blob` and thus this will still accept any `File` as before.

The benefit of also allowing `Blob` is that those are user-creatable, and you can get them from a `Canvas` if you are e.g. resizing images in the browser before uploading them.


LinusU:
Thanks for the quick merge 🙌

Issue #831, opened 2019-06-04
Parse.FacebookUtils.logIn() always have existed() true


fatbattk:
### Issue Description

New Facebook account signups are always set as an existing users. Hence there is no way to differentiate a new sign up vs a returning log in.

http://docs.parseplatform.org/js/guide/#login--signup

### Steps to reproduce

```
Parse.FacebookUtils.logIn("email").then(function(user) {
  console.log(user.existed());  // always returns true.
});
```
- log in as a new Facebook user.
- check console logs.

### Expected Results

returns `false` when new Parse user.
returns `true` when existing Parse user.

### Actual Outcome

returns `true` on both cases.

### Environment Setup

- **JS SDK**
  - JS SDK version: 2.4.0
  - Application? (Browser, Node, React-Native, etc): Browser

dplewis:
Thanks for reporting. Would you like to submit a PR to address this issue?

fatbattk:
Finally had some time to dive into this and found the problem.

`ParseObject._handleSaveResponse()` never gets any `status` because
`RESTController.request()` returns `response` only.
https://github.com/parse-community/Parse-SDK-JS/blob/2.4.0/src/RESTController.js#L259

Had to modify a `ParseUser` unit test because a new option was introduced as part of fix while not affecting other unit tests too much.

PR #845, opened 2019-07-03
Fix Facebook login so it properly distinguishes between new and existing users.

fatbattk:
fixes: #831

### DIFF
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -2202,9 +2202,11 @@ const DefaultController = {
             batchReady.push(ready);
             const task = function() {
               ready.resolve();
-              return batchReturned.then((responses, status) => {
+              return batchReturned.then((responses) => {
                 if (responses[index].hasOwnProperty('success')) {
                   const objectId = responses[index].success.objectId;
+                  const status = responses[index].status;
+                  delete responses[index].status;
                   mapIdForPin[objectId] = obj._localId;
                   obj._handleSaveResponse(responses[index].success, status);
                 } else {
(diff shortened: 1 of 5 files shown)


dplewis:
Thanks for looking into this. Can you write tests cases for this?

Here is an example for the issue you were having https://github.com/parse-community/Parse-SDK-JS/blob/master/integration/test/ParseUserTest.js#L697

src/ParseObject.js
```diff
@@ -2228,7 +2231,9 @@ const DefaultController = {
                 return params;
               })
             }, options);
-          }).then((response, status) => {
+          }).then((response) => {
+            const status = response.status;
+            delete response.status;
             batchReturned.resolve(response, status);
```
dplewis: Great catch! It does look weird that a promise will resolve to more than one parameter.

`batchReturned.resolve(response, status);` shouldn't have 2 parameters. Can you write a test case to ensure that the status works.

fatbattk:
Just pushed with requested changes.

davimacedo:
Seems good to me. @dplewis could you also review again?

src/RESTController.js
```diff
@@ -255,8 +256,12 @@ const RESTController = {
       }
 
       const payloadString = JSON.stringify(payload);
-      return RESTController.ajax(method, url, payloadString, {}, options).then(({ response }) => {
-        return response;
+      return RESTController.ajax(method, url, payloadString, {}, options).then(({ response, status })=>{
+        if (options.returnStatus) {
+          return { ...response, status };
```
dplewis: The only issue I see is that if the response has a status field it would get overridden.

`object.set('status', 'OK')` would no longer work.

perhaps  `return { ...response, _status: status };` and delete _status

src/ParseObject.js
```diff
@@ -2228,8 +2233,8 @@ const DefaultController = {
                 return params;
               })
             }, options);
-          }).then((response, status) => {
-            batchReturned.resolve(response, status);
+          }).then((response) => {
+            batchReturned.resolve(response);
```
acinader: could simplify to:

`.then(batchReturned.resolve)`

src/RESTController.js
acinader: good catch and I like the solution too.

Issue #833, opened 2019-06-05
Stateless user controller for server-side user handling


yagotome:
**Is your feature request related to a problem? Please describe.**
Yes. I am using this SDK on a NodeJS app which is an API Gateway that handles multiple user sessions in a stateless way (server side). Therefore, it would be nice if the SDK provided a stateless `UserController`, however the `UserController` seems to have been designed to be used in client side handling single user session.

**Describe the solution you'd like**
Methods related to `currentUser` shouldn't be part of a general `UserController`. Methods like `logout` should receive `sessionToken` as parameters. Perhaps `DefaultController` should be chosen according to platform (browser or node).

**Describe alternatives you've considered**
It is also possible to have at least another kind of User Controller. What do you think?

**Additional context**
- _UserController_:
https://github.com/parse-community/Parse-SDK-JS/blob/05e8c79e17c8d7f30280303c41848dcd15b6d5cb/src/CoreManager.js#L116-L130

- _DefaultController_:
https://github.com/parse-community/Parse-SDK-JS/blob/d4eb736ce9ab6cc2fd3c4c0221fba1e0a42763d1/src/ParseUser.js#L810

- Related issue:
https://github.com/parse-community/Parse-SDK-JS/issues/424

dplewis:
Current user doesn't exist on server-side by default. 

Does `become` with sessionToken work for you?

You could also create your own UserController and use `CoreManager.setUserController()`

yagotome:
@dplewis `become` works, but I think `logOut` should work similarly as a convenience method so I could use `singUp`, `logIn`, `become` and `logOut` on server-side instead of handle HTTP requests to _Parse Server_. Would it be possible?

I understand `CoreManager.setUserController()` is available, however as mentioned earlier some method signatures are not general enough (e.g. `logOut` does not receives `sessionToken`, `currentUser` doesn't apply to a server-side app).

yagotome:
@dplewis adding more information. I've tested `become` (from `DefaultController` defined in `ParseUser`) in a server environment and received this exception:

```log
Error: It is not memory-safe to become a user in a server environment
    at Function.become (/mystic-market/node_modules/parse/lib/node/ParseUser.js:762:13)
    at Object.sessionTokenToUser (/mystic-market/build/utils/parse.js:227:45)
    at app.use (/mystic-market/build/app.js:22:46)
[...]

PR #846, opened 2019-07-03
Stateless UserController

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/833

There aren't any CurrentUser in a server environment. This PR allows for use of sessionTokens and prevent errors from being thrown.

Let me know if more tests are needed and any feedback


dplewis:
> My suggestion is to leave the become as is now and create a new function (maybe called "me"?).

`Parse.User.me(sessionToken)` @acinader What do you think?

> Shouldn't we take care of "signUp" and "logIn" as well?

I believe those are already taken care of in a node js environment.

acinader:
ugh.  I can't quite wrap my head around this -- probably just soft after my holiday!

Most of the work I've done with the JS-SDK is cloud code, so I am familiar with the territory, but I could use a use case of what we're trying to accomplish (in the form of an integration test or two) to show what the goal of this change is more clearly.

I read #833 twice, but I can't quite piece together what we're solving and why.

davimacedo:
@acinader currently, there is no method in the Node.js SDK to log out or validate a session token. That's the problem @dplewis is solving. This PR does two different things:
- Include the option of passing a session token argument in the logOut() method, so it can be used in Node.js;
- Changes the become() method to be used in Node.js - for this one, in my opinion, I think we should create a new method maybe called me();

acinader:
ok.  I get the logout fix and I buy it.  Not so sure about the become/me.  

What use case would you use this become/me for?  And where would the session Id be coming from?

When I've wanted server code to be able to 'log in', I have made an auth adapter so I can create a session using login.

davimacedo:
Currently the same UserController is used in two different scenarios:
- Running in web browser:
1) Client uses signUp(username, password), logIn(username, password), or become(sessionToken) functions
2) SDK validates credentials/sessionToken in Parse Server and receive back a valid session token and the current user data
3) SDK stores the session token, current user data and automatically uses the session token in any future api call
4) Client uses logOut() to destroy the session token
- Running in server
1) Client calls some kind of signUp or logIn function in a custom Node.js application
2) Node.js application use uses SDK's signUp(username, password) or logIn(username, password) functions
3) SDK validates credentials in Parse Server and receive back a valid session token and the current user data (it does not store it, differently when used in the web browser)
4) Node.js application returns session token and current user data to client
5) Client stores session token and current user data no matter how
6) For a future request (for instance, find() objects), client send to Node.js the session token
7) Node.js calls SDK's find({ sessionToken }) function
8) SDK sends the operation to Parse Server together with sessionToken and receive back the result
9) Node.js receive result from SDK and send to client

The flow works, except:
- There is no convenient method for logging out, because current SDK's logOut() function does not have sessionToken option - I understand this point we've already pacified
- There is no convenient method for validating a session token and receiving back the user data (take a look in this thread here: https://community.parseplatform.org/t/session-token-validation-with-nodejs/87/27) - This is the become/me function; I agree with @dplewis that it should exist, I am not so comfortable with the name become() though, because it can cause some confusion to the developers as they may think that, after using this function, the next api calls will use the sessionToken automatically (and that's not true for the Node.js SDK). I suggested me, because it is actually the name of the REST endpoint that performs the job in the Parse Server side (https://github.com/parse-community/parse-server/blob/master/src/Routers/UsersRouter.js#L424).

I agree with you that we could have two controllers but I'd tackle this on a separate issue.

acinader:
@davimacedo 

whew.  Thank you for laying that out so crystal clearly.  

the only real purpose of me/become in the node environment would be to validate a sessionToken? or is there any other use case?

If that's the only use, I'm good with it.  My preference is also `me` over become now that I understand you rational @davimacedo

davimacedo:
> the only real purpose of me/become in the node environment would be to validate a sessionToken? or is there any other use case? I can't think of one.

I think it is the only one

> On the two controllers, just asking for your view. After reading your response, the thread you cite and playing with some code, I don't think that breaking into an abstract class with two subclasses would really be worth overhead it would take to figure out what is going on. Though I don't have a strong opinion either way :).

In my personal opinion, I think it is easier to understand if we go with the subclasses instead of the current approach in which I have to read the whole code and understand the "if" clauses. I think it is a low priority issue though.

src/__tests__/ParseUser-test.js
```diff
@@ -675,6 +672,71 @@ describe('ParseUser', () => {
...
+      ajax() {}
+    });
+
+    const user = await ParseUser.me('123abc');
+    expect(user.id).toBe('uid3');
+    expect(user.isCurrent()).toBe(false);
+    expect(user.existed()).toBe(true);
+  });
+
+  it('can retreive a user with sessionToken and masterKey(me)', async () => {
```
acinader: Is there a use case for this?  In other words, how would this request differ from the above?  The master key is included, but does it behave it any differently?  I looked at the router code, but not sure...

PR #855, opened 2019-07-10
Including read preference options in parse query

davimacedo:
See https://github.com/parse-community/parse-server/issues/5515

### DIFF
--- a/src/ParseQuery.js
+++ b/src/ParseQuery.js
@@ -1519,7 +1519,7 @@ class ParseQuery {
    * @param {String} subqueryReadPreference The read preference for the sub queries.
    * @return {Parse.Query} Returns the query, so you can chain this call.
    */
-  readPreference(readPreference: string, includeReadPreference: string, subqueryReadPreference: string): ParseQuery {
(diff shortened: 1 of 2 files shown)


src/ParseQuery.js
```diff
@@ -1482,6 +1512,20 @@ class ParseQuery {
...
   }
 
+  /**
+   * Changes the read preference that the backend will use when performing the query to the database.
+   * @param {String} readPreference The read preference for the main query.
+   * @param {String} includeReadPreference The read preference for the queries to include pointers.
+   * @param {String} subqueryReadPreference The read preference for the sub queries.
+   * @return {Parse.Query} Returns the query, so you can chain this call.
+   */
+  readPreference(readPreference: string, includeReadPreference: string, subqueryReadPreference: string): ParseQuery {
```
dplewis: Are all theses fields required? For typing you can do:

`includeReadPreference?: string, subqueryReadPreference?: string`

davimacedo: just fixed

PR #860, opened 2019-07-11
Add options for AnonymousUtils

dplewis:
This PR also addresses an issue.

```
const user1 = await Parse.User.signUp('anon-not', '1234'); // is current user
const user2 = await Parse.AnonymousUtils.logIn(); // request are made with user1 sessionToken
user2.set('foo', 'bar');
// user2 doesn't have a sessionToken because of https://github.com/parse-community/parse-server/pull/5801
// LinkWith sessionToken won't generate new session token
user2.save();
```


dplewis:
I’m going to write a few test them draft a release

PR #874, opened 2019-07-23
Support for Wechat applet

einsqing:
Support for Wechat applet，https://developers.weixin.qq.com/miniprogram/en/dev/

### DIFF
--- a/src/Parse.js
+++ b/src/Parse.js
@@ -267,6 +267,10 @@ if (process.env.PARSE_BUILD === 'node') {
   Parse.Hooks = require('./ParseHooks');
 }
 
+if (process.env.PARSE_BUILD === 'weapp') {
+  wx.Parse = Parse;
+}
+
 // For legacy requires, of the form `var Parse = require('parse').Parse`
 Parse.Parse = Parse;
 
--- a/src/Parse.js
+++ b/src/Parse.js
@@ -267,10 +267,6 @@ if (process.env.PARSE_BUILD === 'node') {
   Parse.Hooks = require('./ParseHooks');
 }
 
-if (process.env.PARSE_BUILD === 'weapp') {
-  wx.Parse = Parse;
-}
-
 // For legacy requires, of the form `var Parse = require('parse').Parse`
 Parse.Parse = Parse;
 
--- a/src/RESTController.js
+++ b/src/RESTController.js
@@ -38,6 +38,58 @@ if (typeof XMLHttpRequest !== 'undefined') {
 if (process.env.PARSE_BUILD === 'node') {
   XHR = require('xmlhttprequest').XMLHttpRequest;
 }
+if (process.env.PARSE_BUILD === 'weapp') {
+  class WeappXhr {
+    constructor() {
+      this.header = {};
+      this.readyState = 4;
+      this.status = 0;
+      this.responseText = "";
+      this.responseHeader = {};
+      this.method = "";
+      this.url = "";
+      this.onreadystatechange = () => {}
+    }
+
+    getAllResponseHeaders() {
+      let header = "";
+      for(const key in this.responseHeader){
+        header += key + ':' + this.getResponseHeader(key) + '\r\n'
+      }
+      return header;
+    }
+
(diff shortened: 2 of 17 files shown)


src/RESTController.js
```diff
@@ -38,6 +38,58 @@ if (typeof XMLHttpRequest !== 'undefined') {
 if (process.env.PARSE_BUILD === 'node') {
   XHR = require('xmlhttprequest').XMLHttpRequest;
 }
+if (process.env.PARSE_BUILD === 'weapp') {
+  class WeappXhr {
```
dplewis: Move this to a new file `WeAppXHR.js`? 
Why can't you use XHR? Are there some restrictions?

src/Parse.js
```diff
@@ -267,6 +267,10 @@ if (process.env.PARSE_BUILD === 'node') {
   Parse.Hooks = require('./ParseHooks');
 }
 
+if (process.env.PARSE_BUILD === 'weapp') {
+  wx.Parse = Parse;
```
dplewis: Parse is already global, would you need this?

src/StorageController.weapp.js
```diff
@@ -0,0 +1,50 @@
...
+ * Copyright (c) 2015-present, Parse, LLC.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @flow
+ */
+localStorage = {
```
dplewis: Is there a reason why you are overriding localStorage? Can you use wx.getStorageSync and others directly below?

src/RESTController.js
einsqing: The Wechat miniprogram does not support XHR

src/Parse.js
einsqing: The Wechat miniprogram does not support global，And use wx instead of global

src/StorageController.weapp.js
einsqing: The Wechat miniprogram does not support localStorage，https://developers.weixin.qq.com/miniprogram/en/dev/api/storage/wx.setStorageSync.html

dplewis: You can remove this and just use wx directly no? It’s basically a wrapper.

src/Parse.js
dplewis: That’s ok as long as wx doesn’t use export a Parse Class in the future, you could override it.. Would wx._Parse make more sense? It’s up to you if you want to keep it.

src/RESTController.js
dplewis: Do queries etc work with this wx.request? If so you can ignore my comment.

dplewis: I see they have a web socket api, you could implement that as well. Do files upload correctly?

einsqing: Supporting queries, socket with my support when I have time

src/Parse.js
einsqing: It's customary to use wx.Parse in this way, or use export a Parse。

src/RESTController.js
einsqing: Livequery has been supported

dplewis:
@einsqing I updated the readme can you have a look? Also if you have time, you can add LocalDatastore support as well.

einsqing:
@dplewis LocalDatastore has been supported

dplewis:
Is there anything you want to add or change in the readme like documentation or usage with WeChat?

einsqing:
@dplewis Wechat minprogram is limited to 2m in size. Generally, we use dist/parse.weapp.min.js directly. You can also require ("parse/weapp"), but you need to turn on building NPM packages

image

dplewis:
Since you don’t use wx.Parse can you remove it from the PR?

Can you also update the readme with npm run build and copy /dist/parse.weapp.js

einsqing:
@dplewis Update readme and remove wx.Parse

PR #878, opened 2019-07-31
Ensure LiveQuery subscribes

dplewis:
Its a known issue that query.subscribe instantly returns a subscription without waiting until it subscribes. This PR fixes that.

This PR will make it easier to write Live Query tests on Parse Server.

### DIFF
--- a/src/LiveQueryClient.js
+++ b/src/LiveQueryClient.js
@@ -178,9 +178,9 @@ class LiveQueryClient extends EventEmitter {
    *
    * @param {Object} query - the ParseQuery you want to subscribe to
    * @param {string} sessionToken (optional)
-   * @return {Promise<LiveQuerySubscription>} subscription
+   * @return {LiveQuerySubscription} subscription
    */
-  subscribe(query: Object, sessionToken: ?string): Promise<LiveQuerySubscription> {
+  subscribe(query: Object, sessionToken: ?string): LiveQuerySubscription {
     if (!query) {
       return;
     }
@@ -208,9 +208,8 @@ class LiveQueryClient extends EventEmitter {
     this.connectPromise.then(() => {
       this.socket.send(JSON.stringify(subscribeRequest));
     });
-    return subscription.subscribePromise.then(() => {
-      return subscription;
-    });
+
+    return subscription;
   }
 
   /**
@@ -440,7 +439,7 @@ class LiveQueryClient extends EventEmitter {
 
     // handle case when both close/error occur at frequent rates we ensure we do not reconnect unnecessarily.
     // we're unable to distinguish different between close/error when we're unable to reconnect therefore
-    // we try to reonnect in both cases
+    // we try to reconnect in both cases
     // server side ws and browser WebSocket behave differently in when close/error get triggered
 
     if (this.reconnectHandle) {
--- a/src/LiveQueryClient.js
+++ b/src/LiveQueryClient.js
@@ -360,7 +360,6 @@ class LiveQueryClient extends EventEmitter {
       break;
     case OP_EVENTS.SUBSCRIBED:
       if (subscription) {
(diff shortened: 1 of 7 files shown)


src/LiveQueryClient.js
```diff
@@ -178,9 +178,9 @@ class LiveQueryClient extends EventEmitter {
    *
    * @param {Object} query - the ParseQuery you want to subscribe to
    * @param {string} sessionToken (optional)
-   * @return {Object} subscription
+   * @return {Promise<LiveQuerySubscription>} subscription
    */
-  subscribe(query: Object, sessionToken: ?string): Object {
```
davimacedo: Since it is a breaking change, I was wondering if perhaps would be better to continue returning the subscription object. The ones interested in waiting it to resolve can just do subscription.subscribePromise.then(() => { ... });

dplewis: This isn't a breaking change as `query.subscribe()` already returns `Promise<LiveQuerySubscription>`. As you can see by the [integration tests](https://github.com/parse-community/Parse-SDK-JS/blob/master/integration/test/ParseLiveQueryTest.js) I didn't have to change anything.

src/LiveQueryClient.js
```diff
@@ -358,12 +360,15 @@ class LiveQueryClient extends EventEmitter {
...
       if (subscription) {
+        subscription.subscribed = true;
+        subscription.subscribePromise.resolve();
         subscription.emit(SUBSCRIPTION_EMMITER_TYPES.OPEN);
       }
       break;
     case OP_EVENTS.ERROR:
       if (data.requestId) {
         if (subscription) {
+          subscription.subscribePromise.resolve();
```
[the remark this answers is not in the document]
dplewis:
```
subscription.on('error', (error) => {}) // should be used if you want error.
```
I used `subscribePromise.resolve()` only just so that subscribe() doesn't hang and always returns subscription.

src/LiveQueryClient.js
davimacedo: I think the `await` in the integration tests are unnecessary. It returns a LiveQuerySubscription object as you can see [here](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/LiveQueryClient.js#L205)

davimacedo: The docs do not show the `await` as well: http://docs.parseplatform.org/js/guide/#create-a-subscription

dplewis: Docs should be updated.

You have to look [query.subscribe()](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseQuery.js#L1611)

davimacedo: so agree with you that `query.subscribe()` is currently already returning a `Promise` and the related docs should be fixed, by `client.subscribe()` is currently not returning a `Promise` and the [docs](http://docs.parseplatform.org/js/guide/#subscribe) are right. It will be a breaking change for the `client.subscribe()`. Maybe you should do the trick [here](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseQuery.js#L1620), leaving the client as is now and awaiting the subscription promise to resolve, before returning the subscription.

dplewis: Oh I see

src/LiveQueryClient.js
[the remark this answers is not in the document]
davimacedo: let's say I do `query.subscribe().then(()=> { ... }).catch(() => { ... });`, what will happen if the subscription fails to complete?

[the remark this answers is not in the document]
dplewis: The chances of that happening are unlikely. unless something like https://github.com/parse-community/parse-server/pull/5871 where the json is invalid. But I get where you are going with this.

PR #881, opened 2019-08-04
feat(ParseObject): Add option `cascadeSave` to save()

### LINKED ISSUE #864: Option to prevent cascade saving
**Is your feature request related to a problem? Please describe.**
When saving an object in which properties there's a pointer to another object, the SDK launches a batch operation to save all modifications made to all nested objects, which is great but... when you don't have write access to any of those (for example when you're saving a Message object pointing to another user's profile) you need to **make sure nested objects are not `dirty`** or call `rever()` in order **not to get the `Object not found` error**.

**Describe the solution you'd like**
A parameter in the saving options, something like `ParseObject.save(null, {cascadeSave: false})` would address this issue.

**Describe alternatives you've considered**
A workaround for now is to call `Parse.Object.revert()` on all objects to which you know this user does not have write permissions.

RaschidJFR:
The new feature can be used to prevent cascade saving in nested objects (Only ignores objects that have already been saved previously):

```
parentObject.save(null, { cascadeSave: false });  // Child objects will not be saved (again)
```

Closes #864

### DIFF
--- a/integration/test/ParseObjectTest.js
+++ b/integration/test/ParseObjectTest.js
@@ -934,6 +934,52 @@ describe('Parse Object', () => {
     });
   });
 
+  it('can skip cascade saving as per request', async(done) => {
+    const Parent = Parse.Object.extend('Parent');
+    const Child = Parse.Object.extend('Child');
+
+    const parent = new Parent();
+    const child1 = new Child();
+    const child2 = new Child();
+    const child3 = await new Child().save();
+
+    child1.set('name', 'rob');
+    child2.set('name', 'sansa');
+    child3.set('name', 'john');
+    parent.set('children', [child1, child2]);
+    parent.set('bastard', child3);
+
+    await parent.save(null, { cascadeSave: false });
+    const results = await new Parse.Query(Child)
+      .ascending('name')
+      .find();
+
+    assert.equal(results.length, 3);
+    expect(results[0].get('name')).toBeUndefined();
+    assert.equal(results[1].get('name'), 'rob');
+    assert.equal(results[2].get('name'), 'sansa');
+
+    parent.set('dead', true);
+    child1.set('dead', true);
+    await parent.save(null);
+    const rob = await new Parse.Query(Child)
+      .equalTo('name', 'rob')
+      .first();
+
+    expect(rob.get('dead')).toBe(true);
+
+    parent.set('lastname', 'stark');
+    child3.set('lastname', 'stark');
+    await parent.save(null, { cascadeSave: false });
+    const john = await new Parse.Query(Child)
+      .doesNotExist('lastname')
+      .first();
+
+    expect(john.get('lastname')).toBeUndefined();
+
+    done();
+  });
+
   it('can do two saves at the same time', (done) => {
     const object = new TestObject();
     let firstSave = true;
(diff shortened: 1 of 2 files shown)


davimacedo:
@RaschidJFR thanks for the PR. Can you please take a look in the failing tests?

RaschidJFR:
I'm checking it. It seems it could easily be fixed by adding a safe condition [here](https://github.com/RaschidJFR/Parse-SDK-JS/blob/96f0a171d902d75eb9052fed0e26c1b43d3da265/src/ParseObject.js#L1210), but it seems strange to me as [this line](https://github.com/parse-community/Parse-SDK-JS/blob/4023bdd56652f3bf8658806f64dc2c33f39dc9bf/src/unsavedChildren.js#L49) should always return a defined array. So what do you think @davimacedo , should I just do:

```
let unsaved = unsavedChildren(this);  // <- This is supposed to always return a defined array
unsaved = unsaved && unsaved.filter(o => o._localId || options.cascadeSave !== false) || []
```

davimacedo:
I'd do something like this:

```
    let cascadePromise = Promise.resolve();
    if (options.cascadeSave !== false) {
      cascadePromise = controller.save(unsavedChildren(this), saveOptions);
    }
    return cascadePromise.then(() => {
      return controller.save(this, saveOptions);
    });
```

RaschidJFR:
> I'd do something like this:
> 
> ```
>     let cascadePromise = Promise.resolve();
>     if (options.cascadeSave !== false) {
>       cascadePromise = controller.save(unsavedChildren(this), saveOptions);
>     }
>     return cascadePromise.then(() => {
>       return controller.save(this, saveOptions);
>     });
> ```

Thanks for the suggestion @davimacedo , but if I did that new objects wouldn't be saved at all with this option. So I just added the safe check. It seems to work know. Let me know.

davimacedo:
hum... so your idea is, when cascadeSave set to false, not save updates anymore but still save the creates? It is very confusing in my opinion.

RaschidJFR:
Well yes. I think new objects must always be saved, otherwise you may run into unexpected broken binds later on. The idea here is to prevent Parse from saving objects you do not explicitly intend to save, but keeping relational data working.

Maybe it could become less confusing if we try different more explicit name for the parameter?

davimacedo:
I don't know. In my opinion it should either save everything (if cascade set to true, which is the default) or save only the object which `save()` method is invoked (is cascade set to false). Having other objects been saved with cascade set to false is weird in my opinion. For the relationships you can still call the `save()` method for each of the pointers.

For example:
```
const a1 = new Parse.Object('A');
await a1.save({ cascade: false });
const a2 = new Parse.Object('A');
a1.set('somefield', 'somevalue');
a2.set('somefield', 'somevalue');
const b = new Parse.Object('B');
b1.set('pointer1', a1);
b1.set('pointer2', a2);
await b1.save({ cascade: false });
```

Now `a1` has `somefield` set to `undefined` and `a2` has `somefield` set to `somevalue`. That's really a strange behavior for me. For this case, in my opinion, ideally, the developer should receive an error "a2 should be saved before used as a pointer".

RaschidJFR:
It makes sense. It should be enough throwing the default error then.

davimacedo:
@RaschidJFR thanks for the contribution!!! Would you be willed to also add some note about this new option in JS Docs?

RaschidJFR:
Sure. I'm already working on it. What about the api reference? is that automatically generated from the code here?

davimacedo:
Nice! The API reference is already updated: https://parseplatform.org/Parse-SDK-JS/api/master/Parse.Object.html#save :)

PR #883, opened 2019-08-09
Adds _batchCount and _batchIndex properties for #5667

stevestencil:
this allows for [#5667](https://github.com/parse-community/parse-server/issues/5667) on parse-server. There is a separate pull request for parse-server

### DIFF
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -309,6 +309,8 @@ class ParseObject {
     for (attr in pending[0]) {
       json[attr] = pending[0][attr].toJSON();
     }
+    json._batchIndex = this._batchIndex || 0;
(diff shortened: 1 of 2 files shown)


src/ParseObject.js
```diff
@@ -309,6 +309,8 @@ class ParseObject {
     for (attr in pending[0]) {
       json[attr] = pending[0][attr].toJSON();
     }
+    json._batchIndex = this._batchIndex || 0;
```
davimacedo: Will we start sending `batchIndex=0`, `batchCount=1` when it is not even a batch?

stevestencil: I put this here so that you could tell that it was only 1 item being saved

PR #898, opened 2019-08-21
Support object.exists()

dplewis:
Returns a promise boolean if the object exists on the server

See https://github.com/parse-community/parse-server/issues/5950

### DIFF
--- a/integration/test/ParseObjectTest.js
+++ b/integration/test/ParseObjectTest.js
@@ -80,6 +80,14 @@ describe('Parse Object', () => {
     });
   });
 
+  it('can check if object exists', async () => {
+    const object = new TestObject();
+    await object.save();
+    assert.equal(await object.exists(), true);
+    await object.destroy();
+    assert.equal(await object.exists(), false);
+  });
+
   it('can find objects', (done) => {
     const object = new TestObject({ foo: 'bar' });
     object.save().then(() => {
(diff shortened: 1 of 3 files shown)


integration/test/ParseObjectTest.js
```diff
@@ -80,6 +80,14 @@ describe('Parse Object', () => {
     });
   });
 
+  it('can check if object exists', async () => {
+    const object = new TestObject();
```
davimacedo: Could you also include a test here? I mean calling `await object.exists()` before saving for the first time.

src/ParseObject.js
```diff
@@ -933,6 +933,27 @@ class ParseObject {
...
+   *       behalf of a specific user.
+   * </ul>
+   * @return {Promise<boolean>} A boolean promise that is fulfilled if object exists.
+   */
+  async exists(options?: RequestOptions): Promise<boolean> {
+    try {
+      await this.fetch(options);
+      return true;
+    } catch (e) {
+      return false;
```
davimacedo: I think we should check `e.code === 101` and only return false in this case. Otherwise `throw e`. I am afraid that a network issue could be interpreted as an inexistent object.

davimacedo: I was thinking here again. Maybe do a query by id would be better here? I am not sure if the developer, when calling an `exists()` function, is expecting its data to be updated.

dplewis: Thats a good point.

PR #900, opened 2019-08-23
Support fetchAllIfNeededWithInclude

dplewis:
See parse-community/parse-server#5950


acinader:
can you increase coverage?

PR #907, opened 2019-09-04
Add support to get parse config with master key

mtrezza:
Adds support to retrieve the parse config with the master key. Allows to retrieve config parameters that require the master key to be retrieved.

Related to
- https://github.com/parse-community/parse-server/issues/5930
- https://github.com/parse-community/parse-server/pull/5954


mtrezza:
Don't merge yet, integration test missing.

How can I add an integration test for getting config parameters that require the master key?
Currently the JS SDK does not support setting a parameter to require the master key. Can I write directly to the DB and flag a parameter to require the master key for a test?

davimacedo:
You can:
- test if the rest command is sent correctly when passing the master key option; and
- add one test to the [integration tests](https://github.com/parse-community/Parse-SDK-JS/tree/master/integration/test) in which you could first use a REST API call to create the config and then use the SDK to retrieve the config.

mtrezza:
@davimacedo, the test `Parse Config cannot get internal config parameter without masterkey` is failing, but locally it passes. Could it be that the tests are not running against parse server 3.8?

dplewis:
@mtrezza This repo runs against the latest commit. Please do the following.

1. Delete package-lock.json
2. npm install
3. Commit new package-lock.json

mtrezza:
@dplewis Thanks for cleaning this up so quickly!

PR #910, opened 2019-09-05
Add master-key-only option when saving config parameter

mtrezza:
This allows to save a config parameter that can be read with master key only using `Parse.Config.save(parameters, masterKeyOnly)`.

An additional argument can be added to specify which of the parameters should be retrievable with master key only, for example:

```
Parse.Config.save(
    { internal: "i", public: "p" },
    { internal: true }
 );
```

A config parameter is retrievable without master key if it is not added in the `masterKeyOnly` argument or is added with value `false`.


src/ParseConfig.js
```diff
@@ -95,10 +95,10 @@ class ParseConfig {
    * @return {Promise} A promise that is resolved with a newly-created
    *     configuration object or with the current with the update.
    */
-  static save(attrs: { [key: string]: any }) {
+  static save(attrs: { [key: string]: any }, masterKeyOnly: { [key: string]: any }) {
```
dplewis: I realized that attrs and masterKeyOnly aren't documented in the comment above. Can you add it and run `npm run docs` and check if they are generated?

mtrezza:
The code coverage diff seems to be a false positive? The integration tests seem to cover all the changed code.

src/ParseConfig.js
```diff
@@ -95,10 +95,10 @@ class ParseConfig {
...
     const controller = CoreManager.getConfigController();
     //To avoid a mismatch with the local and the cloud config we get a new version
-    return controller.save(attrs).then(() => {
+    return controller.save(attrs, masterKeyOnly).then(() => {
       return controller.get();
```
dplewis: You should pass useMasterKey into get(). That ways after you save you will get the current version. You can write a test case for this as well.

src/ParseConfig.js
```diff
@@ -184,16 +184,20 @@ const DefaultController = {
...
+  save(attrs: { [key: string]: any }, masterKeyOnly: { [key: string]: any }) {
     const RESTController = CoreManager.getRESTController();
     const encodedAttrs = {};
+    const encodedMasterKeyOnly = {};
     for(const key in attrs){
       encodedAttrs[key] = encode(attrs[key])
     }
+    for(const key in masterKeyOnly){
+      encodedMasterKeyOnly[key] = encode(masterKeyOnly[key])
+    }
```
dplewis: Is this needed?

dplewis:
@mtrezza The code coverage here refers to the unit tests and not the integration tests.

src/ParseConfig.js
mtrezza: Removed

src/ParseConfig.js
mtrezza: Added test case

src/ParseConfig.js
mtrezza: added docs, generated successfully

package-lock.json
```diff
@@ -10052,7 +10050,7 @@
       "dev": true
     },
     "parse-server": {
-      "version": "github:parse-community/parse-server#ac40bf66f7cdeeb578fd716a8e3dd540470209ca",
+      "version": "github:parse-community/parse-server#bde96a9002a3286c417efaae79856264b8788cb1",
```
dplewis: Can you revert this package-lock.json? It points to an old commit and your tests will fail.

https://github.com/parse-community/parse-server/commit/bde96a9002a3286c417efaae79856264b8788cb1

mtrezza: Done, I wonder why it always points to the old commit after I run npm install.

dplewis: You have to delete your node module folder

mtrezza: Because package.json says `"parse-server": "github:parse-community/parse-server#master"`

mtrezza: That was it!

mtrezza:
What should I do about the decreased coverage? Should I add unit test even though the integration tests cover it already?

dplewis:
Even tho the coverage is a 100% now. A few unit test is recommended

Add tests here https://github.com/parse-community/Parse-SDK-JS/blob/master/src/__tests__/ParseConfig-test.js#L103

You can check out the ParseObject-test.js

Since you are passing in parameters, you want to check that those parameters get passed into the restController

mtrezza:
> Since you are passing in parameters, you want to check that those parameters get passed into the restController

How would I do this, like this?

```
it('can save a config object with a parameter that be retrieved with masterkey only', async () => {
    ParseConfig.save(
      {internal: 'i', number: 12},
      {internal: true});
    expect(CoreManager.getRESTController().request.mock.calls[0]).toEqual([
      'PUT',
      'config',
      { params: {internal: 'i', number: 12}, masterKeyOnly: {internal: true} },
      { useMasterKey: true }]);
  });
```

dplewis:
There are a few ways, you can do it

```
fit('can get a config object with master key', async () => {
    CoreManager.setRESTController({
      request(method, path, body, options) {
        expect(method).toBe('GET');
        expect(path).toBe('config');
        expect(body).toEqual({});
        expect(options).toEqual({ useMasterKey: true });
        return Promise.resolve({
          params: {
            str: 'hello',
            num: 45
          }
        });
      },
      ajax() {}
    });
    const config = await ParseConfig.get({ useMasterKey: true });
    expect(config.get('str')).toBe('hello');
    expect(config.get('num')).toBe(45);
    const path = Storage.generatePath('currentConfig');
    expect(JSON.parse(Storage.getItem(path))).toEqual({
      str: 'hello',
      num: 45,
    });
  });
```

mtrezza:
The unit test fails I added fails because `encode` results in `undefined` for all values: https://github.com/parse-community/Parse-SDK-JS/blob/2fc8da088b53a4bfde36bb0b6cfab8aebc7b868e/src/ParseConfig.js#L190-L192

Strange, because that logic was not changed in this PR.

mtrezza:
Amazing! Thanks @dplewis

mtrezza:
@dplewis Could you do a release so we can close https://github.com/parse-community/parse-server/issues/5930?

Issue #911, opened 2019-09-06
`matchesKeyInQuery` not working in `liveQuery`


JeromeDeLeon:
### Issue Description

I'm using `matchesKeyInQuery` and I got initial result when doing `query.find()` but when integrating `liveQuery`, I'm not able to get any data whenever I updated some fields on the object. It worked when I tried commenting out the `matchesKeyInQuery`.

### Steps to reproduce

1. Create 3 Classes (Sample, Test, Tester)
2. Fill them
```
Tester: {
    firstName: "Tester1",
}

Test: {
    name: "testName",
    tester: Pointer to Tester
}

Sample: {
    active: true,
    tester: Pointer to Tester
}
```

3.  Run this code (parse is already imported, and all classes were specified in Server config)
```js
const getSampleQuery = () => {
    const query = new Parse.Query("Sample");
    query.equalTo("active", true);

    return query;
}

const getQuery = () =>  {
    const query = new Parse.Query("Test");
    query.contains("name", "test");
    query.matchesKeyInQuery("tester", "tester", getSampleQuery());

    return query;
}

const test = async () => {
    const query = await getQuery();
    const subscription = await query.subscribe();

    console.log("initial", await query.find());

    subscription.on("open", () => {
      console.log("subscription opened");
    });

    subscription.on("close", () => {
      console.log("subscription closed");
    });

    subscription.on("enter", object => {
      console.log("object entered", object);
    });

    subscription.on("create", object => {
      console.log("object created", object);
    });

    subscription.on("update", object => {
      console.log("object updated", object);
    });

    subscription.on("leave", object => {
      console.log("object left", object);
    });
}

test();
```

4. Set the value of `name` in `Test` class to empty
### Expected Results

You would receive a `leave` event or `update` event when you update `name` in `Test`

### Actual Outcome

None

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.8.0
  - Operating System:     Windows 10
  - Hardware:             [FILL THIS OUT]
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): localhost

- **JS SDK**
  - JS SDK version: 2.7.0
  - Application? (Browser, Node, React-Native, etc): Browser

### Logs/Trace

After initial run, I got this on console:
```
initial [ParseObject]
subscription opened
```

Network WS log:
```
[...]

PR #913, opened 2019-09-06
Add failing test to `matchesKeyInQuery`

JeromeDeLeon:
https://github.com/parse-community/Parse-SDK-JS/issues/911
@dplewis

### DIFF
--- a/integration/test/ParseLiveQueryTest.js
+++ b/integration/test/ParseLiveQueryTest.js
@@ -7,6 +7,7 @@ const sleep = require('./sleep');
 
 const TestObject = Parse.Object.extend('TestObject');
 const DiffObject = Parse.Object.extend('DiffObject');
+const SampleObject = Parse.Object.extend('SampleObject');
 
 describe('Parse LiveQuery', () => {
   beforeEach((done) => {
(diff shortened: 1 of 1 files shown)


JeromeDeLeon:
Same result I got when I ran my own test on it. I even tried to use other events, but no luck.

integration/test/ParseLiveQueryTest.js
```diff
@@ -32,6 +33,35 @@ describe('Parse LiveQuery', () => {
...
+    const subscripton = await query.subscribe();
+
+    subscripton.on('update', object => {
+      assert.equal(object.get('name'), 'tester');
+      done();
+    });
+
+    objectA.set({ name: 'tester' });
+    await objectA.save();
+  });
```
dplewis: Saving object A with tester doesn’t trigger the live query. 

The trigger occurs when the save matches the query you subscribe to. 

I can look more in-depth shortly

JeromeDeLeon: I changed the event to `leave`

dplewis:
Can you check Travis build and lint your code?

PR #924, opened 2019-09-16
feat(ParseObject): enable subclasses to set initial values for props

RaschidJFR:
Sobclasses of `ParseObject` where unable to set initial values for their props and maintain them after fetching from the server. The function `ParseObject.fromJSON()` has been modified to allow this feature. Two test cases have been added.

It may close parse-community/parse-server#5521


src/ParseObject.js
```diff
@@ -1701,7 +1701,7 @@ class ParseObject {
       throw new Error('Cannot create an object without a className');
     }
     const constructor = classMap[json.className];
-    const o = constructor ? new constructor() : new ParseObject(json.className);
+    const o = constructor ? new constructor(json) : new ParseObject(json.className, Object.assign(json, { className: undefined }));
```
dplewis: `const o = constructor ? new constructor(json) : new ParseObject(json.className)` should be fine for your test to work.

There are other issues this brings up. There are other Subclasses of ParseObject like `_User` and `_Session` that will try to set readonly fields. You could try to clone the json and delete the readonly fields before passing to the constructor but I don't know what other issue this would cause.

dplewis: @RaschidJFR Thoughts?

RaschidJFR: Yes, sorry I didn’t reply earlier. I’ve already done that with good results. Haven’t  pushed though.

 I’ve got almost all tests passing now, just a couple left.  What I’m trying to figure out now is a way to prevent pending ops from overwriting the properties set in the constructor. 

Tomorrow I’ll push my progress so anyone can have a look.

RaschidJFR:
@dplewis So I pushed my last commit... it turns out that it was just what you suggested. I havn't continued working on it since a couple of weeks so I can't remember exactly where I am, I'll retake it hopefully soon. 

For now there are only a couple of tests failing for `ParseObject`: `'conver from JSON'` and `retrieve subclass objects' (seems like we're back to square one). I just need to figure out how to keep the attributes from [_finishFetch()](https://github.com/parse-community/Parse-SDK-JS/blob/8c9bf87537129a8300a075edd42c36ecc451193a/src/ParseObject.js#L1740) over the PendingOps.

src/ParseObject.js
```diff
@@ -1700,8 +1700,23 @@ class ParseObject {
     if (!json.className) {
       throw new Error('Cannot create an object without a className');
     }
+
     const constructor = classMap[json.className];
-    const o = constructor ? new constructor() : new ParseObject(json.className);
+
+    // Remove reserved fields
+    const attributes = {}
+    Object.assign({}, json);
```
VladStepanov, line 1708: if you wanted to clone `json`, just `const clonedJson = Object.assign({}, json)` and after refer to `clonedJson`. Otherwise you are mutating `json`

victorx98:
@RaschidJFR @dplewis  Any update on this PR?  Are we going to merge it anytime soon? Thanks!

RaschidJFR:
Hi @victorx98 Sorry, work is keeping me busy so I won't be able to re-take it at least for the next 4 weeks. I see there's been increasing activity around this topic lately so I will dive into it when I'm out of homework. Meanwhile anyone how wishes to retake the subject is welcome to give a hand.

dplewis:
Closing as this feature already exists. You can override the initialize function.

```
const Subclass = Parse.Object.extend('Subclass', {
 initialize(_attributes, _options, number) {
   this.number = number || -1;
 },
});
const object = new Subclass({}, {}, 57);
object.number // 57
```

mtrezza:
@dplewis That looks interesting, have we documented this use of `Parse.Object.extend` anywhere?

dplewis:
https://parseplatform.org/Parse-SDK-JS/api/6.0.0/Parse.Object.html#.extend

Should probably remove the docs for Backbone compatibility

Issue #936, opened 2019-09-30
Unable to resolve module 'EventEmitter' - React Native 0.61.1


kekami:
### Issue Description
After upgrading from react-native v0.60.4 to v0.61.1 I receive a red screen where parse is unable to location EventEmitter in the project. 

Using parse version 2.7.1.

Nothing else was changed between 0.60.4 => 0.61.1

### Steps to reproduce

1. Upgrade react native using https://react-native-community.github.io/upgrade-helper/
2. yarn
3. pod install
4. react-native run-ios
5. Start the packager

### Expected Results

Manage to start the app

### Actual Outcome

![Screenshot 2019-09-30 at 13 40 29](https://user-images.githubusercontent.com/24361798/65875748-31ad2900-e388-11e9-89da-a0028d92e07d.png)
### Environment Setup

- **JS SDK**
  - JS SDK version: 2.7.1
  - Application? (Browser, Node, React-Native, etc): RN0.61.1

renanogueira:
same here

VenomApps:
Whats the status on this is there a fix?

OmidRReza:
this fixed my problem:

go to the file:  

> node_modules\parse\lib\react-native\EventEmitter.js


and change this line:

`const EventEmitter = require('EventEmitter');`

to this:

`const EventEmitter = require('../../../react-native/Libraries/vendor/emitter/EventEmitter');`

dplewis:
Closing via https://github.com/parse-community/Parse-SDK-JS/pull/946

BalalRaza:
When will we publish this update?

jacinto123:
Yeah we really need this, I hope we can get this published soon.

dplewis:
There are a few PR's opened, once they are reviewed we can do a release.

I'll draft one right now.

DurandSacha:
I have a same problem here.

dplewis:
@DurandSacha can you open a new issue and fill out the template as much as possible

pastean:
`TypeError: undefined is not an object (evaluating '_$$_REQUIRE(_dependencyMap[0], "../../../react-native/Libraries/vendor/emitter/EventEmitter").prototype.addListener')`

I've been stuck on this for a couple hours. Can anyone point me in the right direction? Worked before adding @types/parse. Not working after removing the package though.

maidrissi:
> 
> 
[...]

PR #946, opened 2019-10-09
Fix EventEmitter crash in react-native for v0.61.+ version

KarlosQ:
This fix the crash when try to require EventEmitter on 0.61 react-native version

Related issue #936


KarlosQ:
> LGTM!
> 
> Would this support almost all versions of React Native?

Yes, all versions (at least 0.50 or above) have EventEmitter in this path.

dplewis:
Thanks for the PR!

Issue #925, opened 2019-09-16
Error when doing `npm install` on Windows


JeromeDeLeon:
### Issue Description
The installation stopped when doing `npm run build` and accessing `./build_releases.sh`.

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : [FILL THIS OUT]
  - Operating System:     Windows 10
  - Hardware:             [FILL THIS OUT]
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): [FILL THIS OUT]

- **JS SDK**
  - JS SDK version: [FILL THIS OUT]
  - Application? (Browser, Node, React-Native, etc): [FILL THIS OUT]

### Logs/Trace

```
> parse-server@github:parse-community/parse-server#ac40bf66f7cdeeb578fd716a8e3dd540470209ca postinstall D:\Project\ReactJS\Parse-SDK-JS\node_modules\parse-server
> node -p 'require("./postinstall.js")()'

require(./postinstall.js)()
npm WARN rollback Rolling back node-pre-gyp@0.12.0 failed (this is probably harmless): EPERM: operation not permitted, scandir 'D:\Project\ReactJS\Parse-SDK-JS\node_modules\fsevents\node_modules'

> parse@2.7.1 prepare D:\Project\ReactJS\Parse-SDK-JS
> npm run build


> parse@2.7.1 build D:\Project\ReactJS\Parse-SDK-JS
> ./build_releases.sh

'.' is not recognized as an internal or external command,
operable program or batch file.
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! parse@2.7.1 build: `./build_releases.sh`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the parse@2.7.1 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\Admin\AppData\Roaming\npm-cache\_logs\2019-09-16T15_22_28_135Z-debug.log
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! parse@2.7.1 prepare: `npm run build`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the parse@2.7.1 prepare script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

JeromeDeLeon:
[...]

PR #947, opened 2019-10-09
Support compiling on Windows

dplewis:
Converts bash script to node script.

No longer requires gulp to be installed globally.

Fixes: https://github.com/parse-community/Parse-SDK-JS/issues/925

Let me know if this can be improved.

@JeromeDeLeon Can you test if this works?


JeromeDeLeon:
I was able to install it successfully. Thanks 😄

PR #951, opened 2019-10-12
Support Parse.File getData cancel

dplewis:
Allow to cancel downloading file data or uploading via uri (uri is first downloaded to base64 then uploaded to the server).

### DIFF
--- a/src/ParseFile.js
+++ b/src/ParseFile.js
@@ -259,6 +259,9 @@ class ParseFile {
     }
   }
 
+  /**
+   * Aborts the request if it has already been sent.
+   */
   cancel() {
     if (this._requestTask && typeof this._requestTask.abort === 'function') {
       this._requestTask.abort();
@@ -403,7 +406,6 @@ const DefaultController = {
           base64: ParseFile.encodeBase64(bytes),
           contentType: xhr.getResponseHeader('content-type'),
         });
-      };
       };
       options.requestTask(xhr);
       xhr.send();
--- a/src/ParseFile.js
+++ b/src/ParseFile.js
@@ -166,8 +166,11 @@ class ParseFile {
     if (!this._url) {
       throw new Error('Cannot retrieve data for unsaved ParseFile.');
     }
+    const options = {
+      requestTask: (task) => this._requestTask = task,
+    };
     const controller = CoreManager.getFileController();
-    const result = await controller.download(this._url);
+    const result = await controller.download(this._url, options);
     this._data = result.base64;
     return this._data;
   }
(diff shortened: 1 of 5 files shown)


src/ParseFile.js
```diff
@@ -348,33 +355,43 @@ const DefaultController = {
...
           ? require('https')
           : require('http');
-        client.get(uri, (resp) => {
+        const req = client.get(uri, (resp) => {
           resp.setEncoding('base64');
           let base64 = '';
           resp.on('data', (data) => base64 += data);
           resp.on('end', () => {
+            if (aborted) {
+              return resolve({});
```
acinader: this is returning a resolved Promise to event distributor, does it actually do anything?  I doubt it.

acinader: on first reading, I got confused by the return here.  I think it would be more clear if you just `resolve({})` without the `return`.

also, instead of having a variable `abort`, couldn't you just `resolve` in the `abort` event handler on [381.](https://github.com/parse-community/Parse-SDK-JS/pull/951/files#diff-67ccbb663bacb4c7556216ddb313067cR381)?

src/ParseFile.js
```diff
@@ -383,12 +400,19 @@ const DefaultController = {
         if (xhr.readyState !== 4) {
```
acinader: we should really use `XMLHttpRequest.DONE` instead of `4`

src/ParseFile.js
```diff
@@ -383,12 +400,19 @@ const DefaultController = {
...
           return;
         }
+        if (aborted) {
+          return resolve({});
```
acinader: what is the value of `this.response` here?

src/ParseFile.js
dplewis: I believe when you abort, the abort handler and end event both gets called. If I call resolve in abort would end still get called?

acinader: I believe it would, but easy enough to test?

dplewis:
@acinader I have made the requested changes

src/ParseFile.js
```diff
@@ -224,13 +227,17 @@ class ParseFile {
           return this;
         });
       } else if (this._source.format === 'uri') {
-        this._previousSave = controller.download(this._source.uri).then((result) => {
+        this._previousSave = controller.download(this._source.uri, options).then((result) => {
+          if (!result || !result.base64) {
```
acinader: I think that 

`!(result && result.base64)`

is easier to read...

src/Xhr.weapp.js
```diff
@@ -1,7 +1,13 @@
 module.exports = class XhrWeapp {
   constructor() {
+    this.UNSENT = 0;
```
acinader: ah.  if you just want to use the numbers then, i'm ok with that too.

dplewis:
Thanks for reviewing!

Issue #941, opened 2019-10-06
Cannot get schema in Parse SDK using Session Token


chromonav:
### Steps to reproduce
- Initialize Parse SDK with client key, session token, 
run the following:
```
    Parse.Schema.all({
      useMasterKey: false,
      sessionToken: '' //a session token here 
    }).then(r => {
      console.log(r)
    })
```
### Expected Results
Should give schema

### Actual Outcome
Throws error: 
![image](https://user-images.githubusercontent.com/5052568/66272884-95ee5200-e88c-11e9-988b-846bef4d8701.png)

### Environment Setup

- **Server**
  - parse-server version: 3.9.0
  - Operating System:   OSX
  - Hardware:  16gb RAM, i5 processor
  - Localhost or remote server? Back4app

- **JS SDK**
  - JS SDK version: 2.7.1
  - Application: React

### Logs/Trace
![image](https://user-images.githubusercontent.com/5052568/66272944-2167e300-e88d-11e9-869d-51676645b383.png)

santiagosemhan:
Hi @chromonav , **master key** is required to fetch the schema. I think you should not store the master key on your client-side. Can you give more details about what are you trying to do?

https://docs.parseplatform.org/js/guide/#schema

chromonav:
@santiagosemhan according to documentation reference `Parse.Schema.get` should be allowed from client-side using session token without using MasterKey:

Please checkout: http://parseplatform.org/Parse-SDK-JS/api/2.7.1/Parse.Schema.html#get
Which is not working.

santiagosemhan:
I recommend you to code a cloud function that return to frontend your schema data. So in your cloud function you could make additional controls or process the response and deliver only the required data and not all Parse schema data.

PS: Masterkey is required, use it in your cloud function.

dplewis:
@chromonav This is my fault. When I added Parse.Schema, copied from Parse.Object documentation.

MasterKey is required. MasterKey would override any sessionToken you pass anyway.

I'll submit a PR to fix this.

PR #959, opened 2019-10-15
Fix: Remove options from Parse.Schema

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/941

Fixes Documentation.

`useMasterKey: true` is sent with every Parse.Schema request so passing it and sessionToken as a parameter is redundant.

https://github.com/parse-community/parse-server/blob/master/src/Routers/SchemasRouter.js#L124

### DIFF
--- a/src/ParseSchema.js
+++ b/src/ParseSchema.js
@@ -10,7 +10,6 @@
  */
 
 import CoreManager from './CoreManager';
-import type { RequestOptions, FullOptions } from './RESTController';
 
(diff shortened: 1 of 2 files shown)


src/ParseSchema.js
```diff
@@ -425,34 +363,30 @@ class ParseSchema {
...
       params,
-      requestOptions
+      { useMasterKey: true }
     );
   },
 
-  get(className: string, options: RequestOptions): Promise {
-    return this.send(className, 'GET', {}, options);
+  get(className: string): Promise {
+    return this.send(className, 'GET', {});
```
acinader: what happens if you just do `this.send(className, 'GET')`
or `this.send(className, 'DELETE')`

would be nice if send set the default param of {} instead of using here?

Issue #583, opened 2018-06-14
CLP needs an API


oallouch:
Hi,
CLP is an amazing feature, but without an easy API to set it, I can't use it. It's too easy to forget to check the dashboard.
I don't know which form it can take (is it Schema-related ?), but I'm sure you'll figure it out.

What do you think about it ?

Thx,
Olivier

flovilmart:
It makes sense, a masterKey only API on the SCHEMA would be nice.

mullwaden:
Would be very nice! Currently I do something like this in lieu of a proper API: 

```
const setCLP = ({ className, clp }) => {
  if (!clp) {
    return Promise.resolve();
  }

  // this approach is used by the dashboard.
  // see: https://github.com/parse-community/parse-dashboard/blob/master/src/lib/stores/SchemaStore.js
  return Parse._request(
    'PUT',
    'schemas/' + className,
    { classLevelPermissions: clp },
    { useMasterKey: true }
  );
};
```

Where the clp object is defined something like: 

```
    clp: {
      get: { requiresAuthentication: true },
      find: {},
      create: { '*': true },
      update: { requiresAuthentication: true },
      delete: {},
      addField: {},
    },
```

kosssi:
@mullwaden I don't understand how I can set good CLP with your function I would like to make this:

![capture d ecran 2019-01-15 a 23 19 03](https://user-images.githubusercontent.com/1135513/51213781-2b6f3d80-191c-11e9-9447-1f9a7eb899cf.png)

Can you describe `clp` object for this example?

oallouch:
I really like the new protectedFields CLP.
It would still be great to have it in a future CPL API :)

PR #960, opened 2019-10-15
Set Class Level Permission via Parse.Schema

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/583

Let me know if there are any other tests needed.
I rarely use CLPs so I may miss something.

### DIFF
--- a/src/ParseSchema.js
+++ b/src/ParseSchema.js
@@ -173,9 +173,6 @@ class ParseSchema {
    * @return {Parse.Schema} Returns the schema, so you can chain this call.
    */
   setCLP(clp: { [key: string]: mixed }) {
-    if (!clp) {
-      return;
-    }
     this._clp = clp;
     return this;
   }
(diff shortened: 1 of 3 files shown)


src/ParseSchema.js
```diff
@@ -161,6 +165,21 @@ class ParseSchema {
...
 
+  /**
+   * Sets Class Level Permissions when creating / updating a Schema.
+   * EXERCISE CAUTION, running this may override CLP for this schema and cannot be reversed
+   *
+   * @param {Object} clp Class Level Permissions
+   * @return {Parse.Schema} Returns the schema, so you can chain this call.
+   */
+  setCLP(clp: { [key: string]: mixed }) {
+    if (!clp) {
```
acinader: why return here.  shouldn't you be able to 'unset' by sending null?

dplewis: True

dplewis:
@acinader How does this look?

Issue #930, opened 2019-09-20
How to set required and defaultValue in Schema?


kyawswarthwin:
With parse dashboard we can set required and defaultValue. I want to known how can I set those value with Parse.Schema in js sdk.

davimacedo:
There is currently no support to set required and defaultValue though the JS SDK. Would you be willed to help tackling this?

dplewis:
@davimacedo @acinader I can take this one. How do you guys see this being used?

```
  addField(name: string, type: string = 'String', required: boolean = false, defaultValue: any) {
    if (!name) {
      throw new Error('field name may not be null.');
    }
    if (FIELD_TYPES.indexOf(type) === -1) {
      throw new Error(`${type} is not a valid type.`);
    }

    this._fields[name] = { type, required, defaultValue };
    return this;
  }
```

or 

```

 addField(name: string, type: string = 'String', options: FieldOptions = {}) {
    if (!name) {
      throw new Error('field name may not be null.');
    }
    if (FIELD_TYPES.indexOf(type) === -1) {
      throw new Error(`${type} is not a valid type.`);
    }

    this._fields[name] = { 
      type,
      required: options.required,
      defaultValue: options.defaultValue,
    };
    return this;
  }

PR #961, opened 2019-10-16
Parse.Schema required fields and defaultValues

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/930

Requires Parse Server 3.7.0+

Feature: https://github.com/parse-community/parse-server/pull/5835

I added a FieldOption to addField and its counter parts. This will allow for future options like uppercase / lowercase for example.

### DIFF
--- a/src/__tests__/ParseSchema-test.js
+++ b/src/__tests__/ParseSchema-test.js
@@ -83,7 +83,7 @@ describe('ParseSchema', () => {
     done();
   });
 
-  it('can create schema fields required and default values', (done) => {
+  it('can create schema fields required and default values', () => {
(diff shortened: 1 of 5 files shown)


dplewis:
@acinader @davimacedo How does this look?

src/__tests__/ParseSchema-test.js
```diff
@@ -69,6 +83,30 @@ describe('ParseSchema', () => {
     done();
   });
 
+  it('can create schema fields required and default values', (done) => {
```
acinader: since this doesn't require any async, you could remove the done from the func declaration and remove the end done().

Issue #934, opened 2019-09-25
Missing lot's of documentation in order to create a custom authentication


rhuanbarreto:
I'm trying to create a custom authentication using Keycloak OpenID Connect Provider as a source for authentication and I'm stumbling over many docs about how to create a custom authentication in order to support it as:

- https://docs.parseplatform.org/parse-server/guide/#custom-authentication
- https://docs.parseplatform.org/js/guide/#linking-users
- https://docs.parseplatform.org/rest/guide/#signing-up-and-logging-in

And in the end, some of the code mentioned on the docs are missing documentation in the SDK, as:

- [_registerAuthenticationProvider](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseUser.js#L799) - Missing all of it
- [_logInWith](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseUser.js#L809) - Missing all of it
- [_linkWith](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseUser.js#L80) - Incomplete, as it doesn't describe the function itself, only the no need to have a _logInWith even though the function exists.

And then I needed to read all the source in order to understand how the things work. I'm still developing the custom authentication, but I'm not fully aware of the inner workings on this, so for now I can't help opening a PR and doing this myself, but it's my duty to open an issue and point this out so all the community can improve it.

dplewis:
https://docs.parseplatform.org/js/guide/#custom-authentication-module

Everything should be there to create a custom auth.

I can open a PR to document those functions.

rhuanbarreto:
I thinks it's better. It's still not clear to me that for making a login, you need to create this provider. This could be better explained on why do you need each function in the provider and how it connects internally in the SDK.

In the end I didn't need to do more than the server. I'm using the rest API now, because it seems more convenient than creating this provider. But soon I will make an extension to the SDK to support Keycloak like the FacebookUtils.

rhuanbarreto:
Also it's important to mention that an ID field is mandatory in order to make the 3rd party login work. I could understand it reading the other authentication methods.

dplewis:
You are right. id field is required and I don’t think hyphen names modules work either.

dplewis:
@rhuanbarreto Can you open a PR? We will gladly review it.

Issue #552, opened 2018-03-05
I hate the underscore in _logInWith :)


oallouch:
The underscore at the beginning of _logInWith says "it's internal, use it at your own risk" .
What do you think ?

oallouch:
Btw, should I use _logInWith or _linkWith ? (when I initially have no user)

PR #963, opened 2019-10-18
Missing documentation for Custom Auth

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/934
Addresses: https://github.com/parse-community/Parse-SDK-JS/issues/552

Adds deprecation for _linkWith and _loginWith

I can check the links works after this PR is merged.

### DIFF
--- a/src/AnonymousUtils.js
+++ b/src/AnonymousUtils.js
@@ -69,7 +69,7 @@ const AnonymousUtils = {
    */
   logIn(options?: RequestOptions) {
     const provider = this._getAuthProvider();
-    return ParseUser._logInWith(provider.getAuthType(), provider.getAuthData(), options);
+    return ParseUser.logInWith(provider.getAuthType(), provider.getAuthData(), options);
   },
 
   /**
@@ -84,7 +84,7 @@ const AnonymousUtils = {
    */
   link(user: ParseUser, options?: RequestOptions) {
     const provider = this._getAuthProvider();
-    return user._linkWith(provider.getAuthType(), provider.getAuthData(), options);
+    return user.linkWith(provider.getAuthType(), provider.getAuthData(), options);
   },
 
   _getAuthProvider() {
--- a/src/interfaces/AuthProvider.js
+++ b/src/interfaces/AuthProvider.js
@@ -1,4 +1,4 @@
-/* eslint-disable */
+/* eslint no-unused-vars: "off" */
 /**
  * Copyright (c) 2015-present, Parse, LLC.
  * All rights reserved.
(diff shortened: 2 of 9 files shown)


src/AnonymousUtils.js
```diff
@@ -69,7 +69,7 @@ const AnonymousUtils = {
    */
   logIn(options?: RequestOptions) {
     const provider = this._getAuthProvider();
-    return ParseUser._logInWith(provider.getAuthType(), provider.getAuthData(), options);
+    return ParseUser.logInWith(provider.getAuthType(), provider.getAuthData(), options);
```
acinader: YES!! nice.

dplewis:
@rhuanbarreto Anything you want to add?

src/interfaces/AuthProvider.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
acinader: why?  can you disable individual lines with the proper offending rule?  Or, can you disable the rules your concerned with here, or is there something more fundamental?  Its really good to be able to see the reason or rule in the comment....

i.e. `/* eslint-disable object-curly-spacing */`

which lets the reader know that object-curly-spacing is the rule that is an issue.

src/interfaces/AuthProvider.js
```diff
@@ -10,28 +11,30 @@
 
 /**
  * Interface declaration for Authentication Providers
+ *
+ * @interface AuthProvider
  */
-export interface AuthProvider {
+export class AuthProvider {
```
acinader: why the change from an interface to class here, just curious.

dplewis: JSDocs didn't like it

dplewis:
There is documentation in the guide https://docs.parseplatform.org/js/guide/#custom-authentication-module

Working code examples are in [ParseUserTest](https://github.com/parse-community/Parse-SDK-JS/blob/master/integration/test/ParseUserTest.js) and a working server setup [here](https://github.com/parse-community/Parse-SDK-JS/blob/master/integration/server.js#L17).

dplewis:
@acinader Perhaps we should add this to the ReadMe since this highly requested?

rhuanbarreto:
I think that this will help more the people looking for answers and it’s enough for now. I managed to implement the Keycloak login but I’m using the REST API for now. Soon I will try to commit something showing how is it on Server side.

dplewis:
@rhuanbarreto Having another set of eyes always helps as we tend to forget new users have to read through the documentation.

If you can try to use `linkWith` and provide feedback on if these changes are clear. We would greatly appreciate it.

PR #977, opened 2019-11-04
Add installationId to LiveQuery

dplewis:
Adds reconnect for error event from server (already reconnects with web socket error).
Adds backwards compatibility for additionalProperties that do not exist in the server.

This PR will require a server update.

Use Case: Determine which client the live query request was sent from.

Will open a PR separately for the following.
```
subscription.on('create', (object, options) => {
  options.installationId, options.clientId, etc
});
```

### DIFF
--- a/src/ParseLiveQuery.js
+++ b/src/ParseLiveQuery.js
@@ -86,9 +86,11 @@ const DefaultLiveQueryController = {
     if (defaultLiveQueryClient) {
       return defaultLiveQueryClient;
     }
-    const currentUser = await CoreManager.getUserController().currentUserAsync();
+    const [currentUser, installationId] = await Promise.all([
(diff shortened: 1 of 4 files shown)


src/ParseLiveQuery.js
```diff
@@ -88,6 +88,7 @@ const DefaultLiveQueryController = {
     }
     const currentUser = await CoreManager.getUserController().currentUserAsync();
     const sessionToken = currentUser ? currentUser.getSessionToken() : undefined;
+    const installationId = await CoreManager.getInstallationController().currentInstallationId();
```
acinader: you should do this async and the current user async together:

```js
const [currentUser, installationId] = await Promise.all([...]);
```

dplewis: Done. I keep forgetting about that trick.

PR #978, opened 2019-11-04
Properly store User Subclass in Storage

dplewis:
@taivo Let me know if any changes can be made to replicate your issue

### DIFF
--- a/src/ParseUser.js
+++ b/src/ParseUser.js
@@ -869,7 +869,7 @@ const DefaultController = {
   updateUserOnDisk(user) {
     const path = Storage.generatePath(CURRENT_USER_KEY);
     const json = user.toJSON();
-    json.className = '_User';
+    json.className = user.className || '_User';
     return Storage.setItemAsync(
       path, JSON.stringify(json)
     ).then(() => {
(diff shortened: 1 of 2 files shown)


dplewis:
@taivo The issue here is a fundamental one.

CustomUser -> extends Parse.User -> super('attributes') -> Parse.User extends Parse.Object -> super('_User') <- CustomUser className never gets set.

dplewis:
@acinader This is the easiest fix I can come up with.

The issue here is that subclass users don't automatically have a className (must be explicitly set or passed in [here](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseUser.js#L46) somehow). The second issue is that even if a className exist, it's hardcoded before it saves to disk.

src/ParseUser.js
```diff
@@ -869,7 +869,7 @@ const DefaultController = {
   updateUserOnDisk(user) {
     const path = Storage.generatePath(CURRENT_USER_KEY);
     const json = user.toJSON();
-    json.className = '_User';
+    json.className = user.className || '_User';
```
acinader: how about 

`json.className = this.constructor.name === 'ParseUser' ? '_User' : this.constructor.name;`

i think that'll work and then you don't have to worry about setting this.className.

dplewis: Perfect!

dplewis: Looks like we have a flaky test. I can't restart the build the button is missing, can you?

taivo:
This is good. When you guys have it merged, I can switch my use case to this and test for a few days.

Issue #967, opened 2019-10-24
Encrypt Current User - LocalStorage


macarthuror:
Hello guys :vulcan_salute:,

Right now I’m trying to encrypt the data which is storage in the browser and I’m using secure-ls , actually is working fine but only is encrypting the ping data not the Current User.

Do you have any idea of how to do it ?

Thanks and have a good day.

BTW. This is the code I’m using

```javascript
import SecureLS from 'secure-ls'
const ls = new SecureLS({ isCompression: false })

Parse.enableLocalDatastore()
Parse.setLocalDatastoreController({
  fromPinWithName: name => ls.get(name),
  pinWithName: (name, objects) => ls.set(name, objects),
  unPinWithName: name => ls.remove(name),
  getAllContents: () => ls.getAllKeys(),
  clear: () => ls.removeAll()
})
```

> I wrote it before in the [community forum](https://community.parseplatform.org/t/encrypt-current-user-localstorage/774) but no one shows up 😅

dplewis:
Sorry for the late reply. That secure-ls trick looks cool.

The currentUser isn't stored in the localDatastore.

This is the [line](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseUser.js#L865) you are looking for.

dplewis:
If you want you could do the same for setting the [StorageController](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/StorageController.browser.js)

macarthuror:
Hello @dplewis 🖖
Sorry for not answering before and thanks for the line I was looking for.

Sure let me try it. 👌😁

dplewis:
I would like to add SecureLS to the SDK but I’m skeptical. StorageController is swappable so we could have that option.

@acinader @davimacedo Thoughts?

acinader:
Off the cuff, just documenting it on the readme and maybe the js sdk guide how to use it would be a decent first step? 

alternatively,  we could make an argument to initialize or a single call to use it wheen Parse SDK is initialized?

dplewis:
I figure it could be the same as setAsyncStorage for React-Native. 

Secure LS has browser support issues i.e Microsoft Edge (insert rant here). I can open a PR for this and we can move the conversation there. @macarthuror what do you think?

macarthuror:
@acinader  I'm gonna documented it in the Readme. Thanks.

@dplewis I like it, also I think we can use other solutions to improve the security without using other repo.

[...]

PR #984, opened 2019-11-05
Reuse StorageController for LDS

dplewis:
Can be useful for https://github.com/parse-community/Parse-SDK-JS/issues/967

* Reduces the number of test and files.

Let me know if changes are needed.

Some of the promises could be written better.


src/LocalDatastoreController.js
```diff
@@ -8,40 +8,40 @@
...
   },
 
-  getAllContents(): Object {
+  async getAllContents(): Object {
+    const keys = await Storage.getAllKeysAsync();
     const LDS = {};
-    for (let i = 0; i < localStorage.length; i += 1) {
-      const key = localStorage.key(i);
+
+    for(const key of keys){
```
acinader: this is fine, but this kind of pattern I would use reduce.

src/__tests__/LocalDatastore-test.js
```diff
@@ -788,71 +789,14 @@ describe('LocalDatastore', () => {
...
-describe('LocalDatastore (BrowserDatastoreController)', () => {
-  beforeEach(async () => {
-    CoreManager.setLocalDatastoreController(BrowserDatastoreController);
+    CoreManager.setStorageController(BrowserStorageController);
+    CoreManager.setLocalDatastoreController(LocalDatastoreController);
     await LocalDatastore._clear();
   });
 
-  it('can store and retrieve values', async () => {
+  fit('can store and retrieve values', async () => {
```
acinader: oops.

dplewis: Just fixed that

src/LocalDatastoreController.js
dplewis: Please help great one. I tried using reduce but gave up.

acinader: sorry, i'm wrong, localstorage is not iterable, so you can't use map and reduce on it...

dplewis: I'm looping through a set of keys and building a json object. localStorage isn't there anymore.

acinader: right, one sec.

acinader: https://runkit.com/arthur/reduce

dplewis: What about the async / await?

acinader: make the reducer function async:

```js
array.reduce(async(accumulator...
```

dplewis: That would be cool, I also figured out how to use reduce, I forgot to return

src/LocalDatastoreController.js
```diff
@@ -23,49 +22,46 @@ const LocalDatastoreController = {
...
+    return Storage.removeItemAsync(name);
   },
 
-  getAllContents(): Object {
-    const LDS = {};
-    for (let i = 0; i < localStorage.length; i += 1) {
-      const key = localStorage.key(i);
+  async getAllContents(): Object {
+    const keys = await Storage.getAllKeysAsync();
+    return keys.reduce(async (previousPromise, key) => {
```
acinader: nice.

Brokray:
Hello @dplewis,      

I just saw you added 'getAllKeys' as a required method for a `StorageController`, could you explain me why so? Is that intended?       

It seems to me that in the definition type of a StorageController there is no getAllKeys method expected.       
Which statement should we believe?     

[Here](https://github.com/parse-community/Parse-SDK-JS/blame/681016da6b521a0060c81e1efbe31ff0d07a3cc1/src/CoreManager.js#L367) you can find the change I am talking about.

dplewis:
That is intended for dumping the LocalDatastore.

If the definition type is missing can you add it?

PR #987, opened 2019-11-06
Support query.map, query.filter, query.reduce

dplewis:


src/__tests__/ParseQuery-test.js
```diff
@@ -1580,6 +1580,28 @@ describe('ParseQuery', () => {
...
+            { objectId: 'I89', size: 'small', name: 'Product 89' },
+            { objectId: 'I91', size: 'small', name: 'Product 91' },
+          ]
+        });
+      }
+    });
+
+    const q = new ParseQuery('Item');
+
+    const results = await q.map((object) => {
```
acinader: how i'd write that...
```js
const results = await q.map(object => object.attributes.size);
```

acinader:
niiiice....but where's reduce!?  and filter!! :).

you might find this interesting to check out: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#Polyfill

dplewis:
@acinader I just thought about it. Should we use each() or find()?

acinader:
use each or find for what?  i'm not following.

src/ParseQuery.js
```diff
@@ -910,6 +910,38 @@ class ParseQuery {
...
+   *   <li>sessionToken: A valid session token, used for making a request on
+   *       behalf of a specific user.
+   * </ul>
+   * @return {Promise} A promise that will be fulfilled once the
+   *     iteration has completed.
+   */
+  async map(callback: (currentObject: ParseObject, index: number, query: ParseQuery) => any, options?: BatchOptions): Promise<Array<any>> {
+    const array = [];
+    let index = 0;
+    await this.each((object) => {
```
dplewis: @acinader This uses `.each()` i'm going to replace it with `find` since `each` has restrictions like no sort/limit/skip etc.

acinader: well, look at the implementation of each.  I think that each, map, reduce and filter all need to be able to be run against a whole collection.  so at least for a first pass, using the same mechanism as each (or actually using each as you have done) is the way to go.

what you've done just make available in the API what I have been doing a lot in my parse code, so I think what you have is a big improvement by moving the pattern into the SDK.

dplewis: Sweet, i'll add reduce and filter to this PR. That reference link really helped.

acinader:
if you wanted to do more tests, you could add reduce to an object and reduce to an array

PR #1007, opened 2019-11-22
run npm audit fix

acinader:
attempting to fix github's critical security issue by running npm audit fix.

do they agree?


acinader:
@davimacedo unfortunately, I can't reproduce the failing test on my local.

davimacedo:
@acinader sorry for the delay. I couldn't reproduce as well. Travis stores a cache of the node_modules and it can be causing the problem. I can try to figure it out but I think that @dplewis already tackled the audit issues at https://github.com/parse-community/Parse-SDK-JS/pull/1020, right?

dplewis:
@acinader the audit fixes were for dev dependencies so they aren’t really vulnerabilities in production.

Issue #1021, opened 2019-12-03
Aborting query use case


JeromeDeLeon:
I followed the test case reference to the new feature `query.cancel` for aborting the request made by @dplewis . I don't know if that was the right thing. After that didn't work, here's what I tried:

```js
query.find().then(console.log).catch(console.error)
setTimeout(() => {
    query.cancel();
}, 0);
```

I saw from the network section that the query was cancelled however, two request were made (one that got cancelled and one that went successfully) and I also saw from the console that it passes the `then` block from above. So how do I use the `cancel`? awaiting before calling abort was nonsense because the request was already completed before aborting.

I thought doing below snippet is possible:

```js
query.find().then(console.log).catch(console.error)
query.cancel();
```

dplewis:
Sorry about that I haven't updated the documentation for that yet.

The cancel query is tricky and could use improvement.

https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseQuery.js#L1849

There is an on change event for the event task that you can subscribe to so that you can know when the query has started. I think an improvement should be made to determine when to set the requestTask

JeromeDeLeon:
Is it not possible to move the onchange inside `cancel`? so it automatically listen fo on change when calling cancel?

dplewis:
I think that would work. Can you try it out?

JeromeDeLeon:
I tried your snippet but it didn't work and the request was still successful.

JeromeDeLeon:
I also tried doing it in ReactJS but no luck:

```js
componentDidMount() {
    // setup my query
    this.query = query;
    this.query.find(console.log)
}

componentWillUnmount() {
    this.query.cancel();
}
```

dplewis:
@JeromeDeLeon I think it might be related to where we set the request task.

https://github.com/parse-community/Parse-SDK-JS/blob/master/src/RESTController.js#L110

Can you try to move it after xhr.send()?

JeromeDeLeon:
I tried forking the lib but failed to run when I installed it using `npm install github...` saying that no such file or directory for `/node_modules/parse/lib/browser/ObjectStateMutations.js`

PR #1027, opened 2019-12-08
Move requestTask after sending request

JeromeDeLeon:
#1021 

@dplewis , I'm currently having problem with testing/experimenting in my project.

PR #1030, opened 2019-12-11
Support query.cancel() on Node

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/1021, https://github.com/parse-community/Parse-SDK-JS/pull/1027

Moves abort event to prevent xhr.open from calling it.
Ensure requestTask is set after the request sends.

* Improves browser support

### DIFF
--- a/integration/test/ParseQueryTest.js
+++ b/integration/test/ParseQueryTest.js
@@ -1998,4 +1998,17 @@ describe('Parse Query', () => {
     }, 0);
     assert.equal(result, 6);
   });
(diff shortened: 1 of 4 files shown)


dplewis:
@JeromeDeLeon This should fix your issue. Can you verify it?

JeromeDeLeon:
@dplewis Thanks. I was able to run it with `npm link` and successfully tested it. I don't know why does it fail with github.

PR #1035, opened 2019-12-15
Fixes logo don't being added to /out

ivnnv:
**Problem**
Parse logo is not showing when visiting[]( https://parseplatform.org/Parse-SDK-JS/api/master/) (which I find my self visiting dozens of time a day 😅 )

broken-logo

**Solution**
jsdoc-conf.js is not being told the assets folder needs to be copied into the output folder
and for it to properly work as we expect, we need to move the logo image down into /assets/img.
Readme.md needs to point to "/img/logo_large.png" to get the logo from "/out/img".

**Extra**
Replacing space by underscore character in "logo large.png" file, so its now "logo_large.png"

.
.
Working logo being shown after this fix:
finally-logo


dplewis:
Thanks for the PR. Can you check the Travis logs? It’s failing.

ivnnv:
Mmm I think it has to do with the capital A in "Assets", it should be "assets" (no capital A) to keep consistency naming across the rest of the repo, but i think you cant rename a folder in git fs just for the cases, so Ill change the config to check if travis finds the folder this way (I guess linux is more sensible to name casing)

ivnnv:
@dplewis fixed. Travis is happy now 😄

PR #1036, opened 2019-12-17
Encrypted Current User in browser

macarthuror:
## Problem
Sometimes we don't want or we don't need to show the information of the current user in the local storage

## Solution
With this implementation the information of the current user will be encrypted in the Local Storage to prevent expose the data.
Made it with the [crypto-js](https://www.npmjs.com/package/crypto-js) package.

### To implement it you only need to use the next code

```javascript
Parse.enableEncryptedUser()
Parse.secret= 'my Secrey Key'
```
#967


package.json
```diff
@@ -31,6 +31,7 @@
   "dependencies": {
     "@babel/runtime": "7.7.4",
     "@babel/runtime-corejs3": "7.7.4",
+    "crypto-js": "^3.1.9-1",
```
dplewis: Can you pin the package and resolve the conflicts? Pinning means remove the ^

src/ParseUser.js
```diff
@@ -873,7 +878,10 @@ const DefaultController = {
 
     json.className = user.constructor.name === ParseUser.name ? '_User' : user.constructor.name;
     return Storage.setItemAsync(
-      path, JSON.stringify(json)
+      path,
+      (!CoreManager.get('ENCRYPTED_USER') && process.env.PARSE_BUILD !== 'browser')
+        ? JSON.stringify(json)
```
dplewis: Can you add a check for Crypto  being null aka not browser? A test would be awesome.

I can see an issue with enabling this feature when not on browser.

src/__tests__/ParseUser-test.js
```diff
@@ -1009,4 +1010,26 @@ describe('ParseUser', () => {
     const authProvider = user.linkWith.mock.calls[0][0];
     expect(authProvider).toBe('testProvider');
   });
+
+  it('can get the current user even encrypted', async () => {
+    CoreManager.set('ENCRYPTED_USER', true);
+    CoreManager.set('ENCRYPTED_KEY', 'secret');
```
dplewis: I see encrypted key is defaulted to null. Can you write a test when encrypted user is true but you don’t set the encrypted key?

src/__tests__/ParseUser-test.js
```diff
@@ -1009,4 +1010,26 @@ describe('ParseUser', () => {
...
+    expect(u instanceof ParseObject).toBe(true);
+
+    u = new ParseUser({
+      username: 'andrew',
+      password: 'secret'
+    });
+    expect(u.get('username')).toBe('andrew');
+    expect(u.get('password')).toBe('secret');
+
+    expect(function() {
```
dplewis: I believe this can be removed.

package.json
macarthuror: Done 👌

src/ParseUser.js
macarthuror: After check some things I changed the conditional and also created a test to check if all works well.

src/__tests__/ParseUser-test.js
macarthuror: Done it, also in the conditional to encrypt the data I check if it's null.

dplewis:
I just realized if you remove the check for browser, this would still work for node and react-native no?

macarthuror:
> I just realized if you remove the check for browser, this would still work for node and react-native no?

At Node I believe there's no problem just let me check react-native compatibility.

dplewis:
I think it should work sad that package uses vanilla js. Also you have conflict.

I’ll pull this down and use this in production.

macarthuror:
Ok, thanks

macarthuror:
Let me try with this package [Forge](https://github.com/digitalbazaar/forge).

dplewis:
@macarthuror I made a few changes to this PR. There is a CryptoController that allows users to add their own implementation. I also updated the tests.

@davimacedo Anything you want to add?

macarthuror:
@dplewis Looks amazing 👌

src/CryptoController.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * Copyright (c) 2015-present, Parse, LLC.
```
acinader: this unnecessary

Issue #1045, opened 2019-12-27
[Discussion] Remove batchSize


dplewis:
**Is your feature request related to a problem? Please describe.**

Features like saveAll and destroyAll uses `/batch` endpoint. I believe this should send a single request to the server.

Currently the `/batch` endpoint send multiple requests in the with the number of objects equal to `batchSize` (default 20). I believe batchSize was added to reduce server load during the parse.com days. Having batchSize will make result in a certain amount of overhead for your client.

[batches](https://github.com/parse-community/Parse-SDK-JS/blob/6dc6e5ca7625adeac1afd3f5fe287ca7ed2f7f3e/src/ParseObject.js#L2157)

**Describe the solution you'd like**

Removing batchSize will save resources and as a bonus will allow `BeforeSaveAll` hooks and the like. There may be query improvements we can add on the server side as well for example Transactions.

**Describe alternatives you've considered**

I have tried setting batch size to a large number (10000) to remove the multiple requests.

What do you think? @acinader @davimacedo

dplewis:
I did some research, you can remove batchSize from destroyAll easily. But saveAll will have at most 2 calls to batch to prevent batch cycles.

PR #1047, opened 2019-12-30
Remove batch size from SaveAll and DestroyAll

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/1045

Adds slight improvement to `saveAll([ParseFiles])`. Should we return the parse files if only saving parse files?

Although batchSize is removed `saveAll` will still make multiple calls to the `/batch` endpoint to check for save cycles.

### DIFF
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -19,7 +19,7 @@ import ParseACL from './ParseACL';
 import parseDate from './parseDate';
 import ParseError from './ParseError';
 import ParseFile from './ParseFile';
-import { when, continueWhile } from './promiseUtils';
+import { when, continueWhile, resolvingPromise } from './promiseUtils';
 import { DEFAULT_PIN, PIN_PREFIX } from './LocalDatastoreUtils';
 
 import {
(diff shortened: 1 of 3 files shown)


davimacedo:
I understand that your idea is to completely remove the batch size option. Besides a batch of only 20 items by default looks small, we could at least keep the option for the developer to define a batch size, which can be useful and very convenient for saving a huge amount of objects. For example, I have recently used the JS SDK to load data to one of my apps and called `Parse.Object.savelAll()` to save more than 100K objects at once. It worked perfectly with the current batch size but I guess it would not work anymore if we completely remove the batch option. The reverse proxies use to have a max payload size and it would probably block my request. So, in my option, we should not remove the feature entirely.

dplewis:
I see, you have a valid point. Maybe we could do `CoreManager.set('BATCH_SIZE', 20)` to change the global default.

davimacedo:
Yes. I'd prefer something like this.

acinader:
ok, next time i'll read the convo first.

if i could dismiss my approval,  i would.

dplewis:
@acinader No problem, I'll close this out and make the changes in a separate PR.

PR #1054, opened 2020-01-08
Add support to add hint to Parse.Query

stevestencil:
issue #6288


stevestencil:
I can go back and add more test but I want to make sure this is what we're looking for?

src/ParseQuery.js
```diff
@@ -713,7 +722,8 @@ class ParseQuery {
     const controller = CoreManager.getQueryController();
     const params = {
       distinct: key,
-      where: this._where
+      where: this._where,
+      hint: this._hint ? this._hint : undefined,
```
dplewis: just `hint: this.hint` is fine

src/ParseQuery.js
```diff
@@ -921,6 +932,13 @@ class ParseQuery {
     });
   }
 
+  hint(value: mixed): ParseQuery {
+    if (typeof value === 'undefined') {
+      delete this._hint;
+    }
+    this._hint = value;
+  }
```
dplewis: This should `return this;` no?

src/ParseQuery.js
```diff
@@ -752,8 +761,10 @@ class ParseQuery {
       throw new Error('Invalid pipeline must be Array or Object');
     }
 
-    const params = { pipeline };
-
+    const params = {
+      pipeline,
+      hint: this._hint ? this._hint : undefined,
```
dplewis: Same as above,

src/ParseQuery.js
```diff
@@ -921,6 +932,13 @@ class ParseQuery {
     });
   }
 
+  hint(value: mixed): ParseQuery {
```
dplewis: Can you add documentation?

src/ParseQuery.js
stevestencil: The reason I did this was because I caused tests to fail because in the constructor this._hint is set to null (like readPreference)... so I can either take setting this._hint as null out of the constructor or keep it like this?

dplewis: You will know once you start writing tests.

src/ParseQuery.js
stevestencil: done!

src/ParseQuery.js
stevestencil: done!

src/ParseQuery.js
stevestencil: done!

src/ParseQuery.js
stevestencil: done!

dplewis: Sweet! Now lets get the server side.

stevestencil: working on it now!

dplewis:
@stevestencil Can you give me access to your branch?

stevestencil:
@dplewis Done!

TomWFox:
Thoughts on documenting this in the guide?

Issue #1049, opened 2020-01-02
Saving a file does not pass in session token


stevestencil:
### Issue Description

When saving a Parse.File from the browser, the session token is not passed into the headers of the request. This only appears to happen when uploading a file directly from the browser.

It appears this will only happen when using the File object from the browser as it calls the [saveFile()](https://github.com/parse-community/Parse-SDK-JS/blob/6dc6e5ca7625adeac1afd3f5fe287ca7ed2f7f3e/src/ParseFile.js#L221) function. All other file sources use the [saveBase64()](https://github.com/parse-community/Parse-SDK-JS/blob/6dc6e5ca7625adeac1afd3f5fe287ca7ed2f7f3e/src/ParseFile.js#L241) function which adds the session token and then calls [RESTController.ajax()](https://github.com/parse-community/Parse-SDK-JS/blob/6dc6e5ca7625adeac1afd3f5fe287ca7ed2f7f3e/src/RESTController.js#L89) function. when calling [saveFile()](https://github.com/parse-community/Parse-SDK-JS/blob/6dc6e5ca7625adeac1afd3f5fe287ca7ed2f7f3e/src/ParseFile.js#L221) it calls  [RESTController.ajax()](https://github.com/parse-community/Parse-SDK-JS/blob/6dc6e5ca7625adeac1afd3f5fe287ca7ed2f7f3e/src/RESTController.js#L89) directly without adding the session token.

### Steps to reproduce

1) create a new file using the File object selected with a file upload control from a browser
2) save the file
3) view the payload that is sent to the server

### Expected Results

Session token should be passed in with request

### Actual Outcome

Only the application id is passed in

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.7.2
  - Operating System:  OSX
  - Hardware: All
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): All

- **JS SDK**
  - JS SDK version: 2.7.1
  - Application? (Browser, Node, React-Native, etc): Browser

dplewis:
I believe you are correct but do files actually use the sessionToken or is this documented anywhere?

stevestencil:
Currently they do not use the sessionToken anywhere. However this is something I think we should change. I have read through multiple issues (#2137 and  #4672) that require the sessionToken to implement. Adding in the sessionToken to the files route is the first step to allowing things like:

- adding security to uploading files
[...]

PR #1057, opened 2020-01-08
Saving files with sessionToken

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/1049

A small optimization with saving files in saveAll.
Clean up deferred promises.

@stevestencil Can you check and see if this works for you?

Let me know if anything else is needed.

### DIFF
--- a/src/ParseFile.js
+++ b/src/ParseFile.js
@@ -208,6 +208,8 @@ class ParseFile {
    *  * Valid options are:<ul>
    *   <li>useMasterKey: In Cloud Code and Node only, causes the Master Key to
    *     be used for this request.
+   *   <li>sessionToken: A valid session token, used for making a request on
+   *     behalf of a specific user.
    *   <li>progress: In Browser only, callback for upload progress
    * </ul>
    * @return {Promise} Promise that is resolved when the save finishes.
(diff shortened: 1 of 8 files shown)


stevestencil:
I noticed that when saving the files by using request() (instead of ajax()) it [sets the session token as "_SessionToken"](https://github.com/parse-community/Parse-SDK-JS/blob/74fd98224f1a3e5cac59b0a0784a25f84f117f04/src/RESTController.js#L262) in the body. With this change you're setting the session token in the header X-Parse-Session-Token. Unless I'm missing something, shouldn't these use one or the other?

dplewis:
Looking [here](https://github.com/parse-community/parse-server/blob/985933955f502009da28fc4241ecd500ae6d44c4/src/middlewares.js) Both of these have existed since the beginning of parse. One represents the REST API with headers the other represents POST request with (headers included in data). In this case SaveFile is using REST style request so for consistency I'm using REST style header here.

I'm not really sure why they used `ajax()` instead of `request()` here, maybe there was a side-effect. We can try it in a separate PR.

For now does this work for you?

acinader:
@dplewis this change looks good to me.  I'd prefer to keep what you've done with the session in the header where we now expect it and wouldn't make a requirement for adding this important capability cleaning up a log established inconsistency (unless you want to that is).

Issue #1055, opened 2020-01-08
LiveQuery connection fails because sessionToken is null, but sessionToken is present in localStorage. Parse also sends sessionToken with some cloud function calls but not others.


akritskiy:
LiveQuery works fine most of the time, but sometimes opening the connection fails because the sessionToken in the request payload is null.

The error message on the backend is: "error: Connect message error %s"

This message comes from: /node_modules/parse-server/lib/LiveQuery/ParseLiveQueryServer.js (line 317) on the back-end. I added some console logs and the actual error message is “Invalid type: null (expected string)” and the request is
```
{ op: 'connect',
  applicationId: 'MyAppId',
  javascriptKey: 'myKey',
  sessionToken: null,
  installationId: '109b15ba-b772-4962-9dc4-ae0da8ded1c9' }
```

The request fails because the sessionToken is null. When I look in local storage, the sessionToken is there, and all other requests work fine as far as I can tell.

Another observation: In the network tab in Chrome Dev Tools, I can see Cloud functions being called. As I go down the list, it looks like the first half of the requests have the sessionToken in the request payload, but the rest don't include the sessionToken. These requests work fine, but they are just Cloud functions and I don't understand why the sessionToken gets sent with some requests but not others. It could be said that maybe these requests don't need the token, but on the other hand, the LiveQuery request definitely does need the token, and it's failing because the session token is not being sent.

### Steps to reproduce

In App.js (React) we have the Parse setup:
```
import Parse from 'parse'
Parse.initialize('MyApp', 'myKey')
Parse.serverURL = process.env.API_URL + '/parse'
Parse.CoreManager.set('REQUEST_ATTEMPT_LIMIT', 0)
```

I made a testing screen where I create a simple query for a single object with a "count" on it, subscribe, then handle updates...

```
const query = new Parse.Query('test')
query.equalTo('objectId', '7bXrkGxHFv')
query.limit(1)

const livequerySubscription = await query.subscribe()

livequerySubscription.on('open', () => {
    console.log('livequery connection opened')
})

livequerySubscription.on('update', obj => {
    console.log('livequery received update, object: ', obj.toJSON())

    let count = obj.toJSON().count
    setCount(count) // This displays the "count" in the UI
})
```

Then I have a button that changes the count and saves.

I log in with a user and navigate to my testing screen. In the console, I see "livequery connection opened." I click the button and see "livequery received update" and the UI gets updated with the new count.

I hit refresh. Same procedure works.

[...]

PR #1058, opened 2020-01-09
Fix subscribing to null sessionToken

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/1055

This PR just forces sessionToken null to be undefined when intializing the client.

If anybody knows how a sessionToken can be null instead of a string, '', or undefined, let me know.

### DIFF
--- a/integration/test/ParseLiveQueryTest.js
+++ b/integration/test/ParseLiveQueryTest.js
@@ -180,4 +180,29 @@ describe('Parse LiveQuery', () => {
     })
     await object.save({ foo: 'bar' });
   });
+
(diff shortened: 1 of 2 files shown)


dplewis:
@acinader How does this look?

acinader:
hm.  does the test case fail without the `this.sessionToken = sessionToken || undefined;`

if i've got it right, your code is changing null to undefined.  null errors, undefined doesn't?

dplewis:
The test does fail without this fix. undefined doesn’t error. 

I can’t think of a case where sessionToken could be null.

PR #1065, opened 2020-01-17
Support File Metadata

stevestencil:


src/ParseFile.js
```diff
@@ -99,11 +101,15 @@ class ParseFile {
    * @param type {String} Optional Content-Type header to use for the file. If
    *     this is omitted, the content type will be inferred from the name's
    *     extension.
+   * @param metadata {Object} Optional key value pairs to be stored with file object (S3 Only)
+   * @param tags {Object} Optional key value pairs to be stored with file object (S3 Only)
```
dplewis: These are optional and maybe used in other StorageAdapters besides S3.

Can these be retrieved somehow?

src/ParseFile.js
```diff
@@ -292,6 +300,52 @@ class ParseFile {
...
+      });
+    }
+  }
+
+  /**
+   * Sets metadata to be saved with file object. Adds to existing metadata
+   * @param {String} key
+   * @param {String} value
+   */
+  addMetadata(key: String, value: String) {
```
dplewis: This can be removed  and just use this._metadata = metadata?

src/ParseFile.js
```diff
@@ -292,6 +300,52 @@ class ParseFile {
...
+      });
+    }
+  }
+
+  /**
+   * Sets tags to be saved with file object. Adds to existing tags
+   * @param {String} key
+   * @param {String} value
+   */
+  addTag(key: String, value: String) {
```
dplewis: same as above

src/ParseFile.js
stevestencil: I added these so that you can add to existing meta data without having to get the existing and add to it

src/ParseFile.js
stevestencil: Add getMetadata() and getTags()?

dplewis: That too, I mean if I fetch a file from database would the metadata and tag be retrieved.

src/ParseFile.js
dplewis: Ok, I think that value should be generic `value: any`. The storage adapters should handle the validation if any.

src/ParseFile.js
stevestencil: no because only the name of the file is actually being stored to the database

dplewis: In S3 case we are retrieving the object from the storage adapter. Doesn't the metadata / tags return with that data? Can you check?

I know that most storage adapters with metadata support only allow setting metadata on creation. In S3 case should `setMetadata` exists? What would happen if you change the metadata and resave?

stevestencil: When retrieving object from S3 the metadata is included but the tags are not. Tags would have to be pulled in a separate request. Maybe fetching the metadata and tags could be added in the future? At least with what I have added so far the data can get to S3 so that files can be better managed view other APIs not currently built in Parse.

src/ParseFile.js
stevestencil: This has been updated

src/ParseFile.js
dplewis: Metadata would be great. I think tags can be added in the future if there is a request for it.

stevestencil: How do you suggest I implement this considering the file adapters only return the data? I propose changing the S3 adapter to return an object with the data, and metadata... and updating parse-server to check if an object is coming in, if so, parse out the data and metadata, and if a buffer is coming in ignore metadata?

dplewis: I believe so. We can take a look at it to prevent breaking changes

stevestencil: ok... can you look at the PR I have for parse-server too? This goes together with it.

src/ParseFile.js
```diff
@@ -292,6 +300,52 @@ class ParseFile {
...
+   * @param {String} key
+   * @param {Mixed} value
+   */
+  addTag(key: String, value: String) {
+    if (typeof key === 'string') {
+      this._tags[key] = value;
+    }
+  }
+
   static fromJSON(obj): ParseFile {
```
dplewis: Can you update fromJSON with _metadata and _tag? 

Just realized type was missing. I think we can skip it for now.

src/ParseFile.js
```diff
@@ -292,6 +300,52 @@ class ParseFile {
     );
   }
 
+  /**
+   * Sets metadata to be saved with file object. Overwrites existing metadata
+   * @param {Object} metadata Key value pairs to be stored with file object
+   */
+  setMetadata(metadata: Object) {
```
dplewis: We should stick the the same convention as _name and _url by making these private. file.metadata() and file.tags() should be enough, metadata is usually immutable. If you want to keep these that fine too.

src/ParseFile.js
stevestencil: I had this added at first but it broke a bunch of tests. Doing this would cause the metadata and tags to be saved to mongo. I believe leaving metadata and tags out makes sense because the they are both stored in S3 which should be the source of truth. So if you want the metadata or tags from a file that has already been saved, it should come from S3 (or whatever storage provider they're using)

src/ParseFile.js
stevestencil: I would propose these be promises that call the storage adapter to fetch the metadata and tags? If the file has not been saved we could return what is set in _metadata or _tags without making a call to the storage adapter. If so I would like to add this at a later time because it would require adding additional methods to the storage adapters.

dplewis: When you originally proposed this feature. I thought metadata would be stored in the DB. This would be way more useful and we would have more control. Is there a reason it’s not there? I don’t think it would be a breaking change.

stevestencil: The reason it does no save to the database now is because files are currently stored in the DB as just a string. I agree this would be great to add to the database but it would require storing the files as objects instead (unless you have some other solution?). My original proposal included links to the S3 documentation showing how metadata and tags are stored with files in S3 directly. My idea behind building all of this (beforeSaveFile, afterSaveFile, metadata and tags) is so that I can A) query against my S3 and have fields to query against and B) Build a new collection for any files that are saved by using the afterSaveFile hook to create a new object, with the location of the file and any other data that I want to store with it. I can then add this new "File" object as a pointer to whatever object I want. Maybe in the future we can add a default "_File" collection that does this automatically?

stevestencil: If you think the setTags and setMetadata could be misleading, I can change it so that those values can only be passed into the options object when saving the file? so something like:
```
const file = new Parse.File('some-file.txt');
await file.save({ metadata: {foo: 'bar'}, tags: {bar: 'foo'});
```

dplewis: You already have them in the constructor. I think that is enough.

stevestencil: so remove the setters and getters?

dplewis: You can leave the getters. `file.metadata()` and `file.tags()`

stevestencil: Remove the setters?

dplewis: Yes

stevestencil: @dplewis if I remove the setters than you would have no way of modifying the metadata and tags on the beforeSaveFile hook.

stevestencil: @dplewis I added some documentation with examples to show how this PR combined with my PR on parse-server will work together

Issue #1063, opened 2020-01-14
Fetching object with id "" doesn't work as expected


percypyan:
### Issue Description

As expected, when you try to fetch an object with an ID that don't exist in database, you get a 101.
But in the case your object as an empty string as ID, the fetch do not throw any error even is the returned data does not match the expected result.

Given the following object: `MyClass.createWithoutData("fakeID")`.
A fetch will trigger the following API call:
```
curl -X GET \
  -H "X-Parse-Application-Id: ${APPLICATION_ID}" \
  -H "X-Parse-REST-API-Key: ${REST_API_KEY}" \
  https://parse-server.com/parse/classes/MyClass/fakeID
```
This call return a 404.

Now, given the following object: `MyClass.createWithoutData("")`.
A fetch will trigger the following API call:
```
curl -X GET \
  -H "X-Parse-Application-Id: ${APPLICATION_ID}" \
  -H "X-Parse-REST-API-Key: ${REST_API_KEY}" \
  https://parse-server.com/parse/classes/MyClass/
```
This call return a list of the first 100 objects of the MyClass collection under the `results` key.

### Steps to reproduce

Simply fetch an object created either of this ways:
- `Object.createWithoutData("")`
- `const object = new Object(); object.id = ""`

### Expected Results

The REST API act as it should, but the SDK should prevent this call by throwing something (a 101 or anything else that you find appropriate).

The other way to handle it is to prevent the usage of `Parse.Object.createWithoutData` method with an empty string. The outcome is that we still can create an object and set is ID to "" manually.

### Actual Outcome

The object contains a new `results` attribute that contain the list of objects fetched.

### Test Case

To accurately reproduce your issue. Add your test cases [here](https://github.com/parse-community/Parse-SDK-JS/tree/master/integration/test) and read the [Contributing Guide](https://github.com/parse-community/Parse-SDK-JS/blob/master/CONTRIBUTING.md) to run the tests. 

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.3.9
  - Operating System:     macOS 10.15 / Heroku operating system (I suppose Ubuntu)
  - Hardware:             MacBook Pro / Heroku servers
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Localhost and Heroku

- **JS SDK**
  - JS SDK version: 2.7.1
  - Application? (Browser, Node, React-Native, etc): Node

### Logs/Trace

[...]

PR #1066, opened 2020-01-17
Throw error when fetching an object without ID

percypyan:
Pull Request fixing issue #1063.

Add a check for ID definition before fetching data.

### DIFF
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -2121,6 +2121,12 @@ const DefaultController = {
       if (options && options.include) {
         params.include = options.include.join();
       }
+      if (!target.id) {
+        return Promise.reject(new ParseError(
+          ParseError.MISSING_OBJECT_ID,
+          'Object does not have an ID'
+        ));
+      }
       return RESTController.request(
         'GET',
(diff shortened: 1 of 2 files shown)


percypyan:
@dplewis Maybe should I only throw for "" ID to avoid breaking code fetching with `null` or `undefined` ID?

src/ParseObject.js
```diff
@@ -179,10 +179,10 @@ class ParseObject {
    * Returns a local or server Id used uniquely identify this object
    */
   _getId(): string {
-    if (typeof this.id === 'string') {
+    if (this.id && typeof this.id === 'string') {
```
dplewis: Can you revert this change? I don't think this is needed. You can write integration test to verify.

percypyan: @dplewis as you wish I've written a unit test but it's maybe useless to add this check

dplewis: An integration tests is different from unit tests. Integration tests run against a parse server instance. Add your tests [here](https://github.com/parse-community/Parse-SDK-JS/blob/master/integration/test/ParseObjectTest.js).

PR #1067, opened 2020-01-18
Added support to delete Parse.File

stevestencil:


src/ParseFile.js
```diff
@@ -271,6 +271,25 @@ class ParseFile {
     this._requestTask = null;
   }
 
+  /**
+   * Deletes the file from the Parse cloud.
+   * @param {Object} options
```
dplewis: You can remove the options as this is masterKey only. Also document that this is master key only or to be used in cloud code.

https://github.com/parse-community/parse-server/blob/master/src/Routers/FilesRouter.js#L32

src/ParseFile.js
```diff
@@ -425,6 +444,11 @@ const DefaultController = {
 
   _setXHR(xhr: any) {
     XHR = xhr;
+  },
+
+  deleteFile: function(name, options) {
```
dplewis: Can you write an integration test?

stevestencil: I'm writing a test now and running into some blocks which might be the reason why this was never added in the first place. According to the REST API documentation you have to use a DELETE method with the X-Parse-Application-Id and X-Parse-Master-Key headers. The normal RestController.request() function is setting all of the Auth data via the body and changing everything to a POST request... I believe once this hits the server it never goes anywhere... I'm still digging...

dplewis: You will have to use RestController.ajax(). I can add it if you want.

stevestencil: I had to step away from my computer. If you can add it, please do!

dplewis:
@davimacedo @acinader Anything you guys want to add?

stevestencil:
@dplewis I would recommend adding an options argument to the destroy() function. Reason being everything else that requires the master key forces you to pass in { useMasterKey: true } into the options. Also, I know right now the master key is required to delete files (which limits it to only cloud code), but in the future I would like to add the ability to do it client side with a session token.
Also, with the addition of beforeDeleteFile and afterDeleteFile hooks (my other PR) passing in the session token when calling destroy() would allow the user object to be passed in. In a nutshell, adding an options argument keeps with existing destroy() functionality and future proofs the function.

PR #1068, opened 2020-01-19
File delete with options

stevestencil:
@dplewis here are the proposed changes to the file deletion endpoint. It requires { useMasterKey: true } to be passed in the destroy() function. Doing this keeps with the same logic as deleting Parse.Objects and allows for session tokens to be added in the future. Without this it would prevent us from ever adding client side file deletion with this function.


dplewis:
I'm going to close this for now. We can reopen in the future if client side deletion is supported. I don't want users to think client side deletion is supported immediately or if ever.

PR #1069, opened 2020-01-20
Adds decrement() to ParseObject

stevestencil:


dplewis:
No offense but  increment with a negative number has worked for the last 5 years: I feel this PR adds more logic.

stevestencil:
I'm using increment with a negative number now. I've used other API's with both increment and decrement functions and thought since this one had increment() it should also have decrement(). Considering you can call increment('fieldName') without an amount and it defaults to 1. I felt having the same functionality for decrement would be a nice to have.

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "parse",
+  "name": "@leapllc/parse",
```
acinader: doh

acinader:
My take is that this is duplicate code for no added value.  it also isn't available in any of our other apis or rest.  I appreciate the effort @stevestencil.

acinader:
If @davimacedo, @dplewis, and/or @TomWFox want to include this change, I at least think we should change the implementation to call increment under the covers.

package.json
stevestencil: doh is right... I accidentally committed this

TomWFox:
I'm not completely against the change but agree it should call increment otherwise it's just more code to maintain run tests on - even if it is simple.

Not sure about other api references / guides but the JS API Reference does not mention that a negative number can be used - it would be nice to mention if this change doesn't go ahead.

stevestencil:
I initially was going to use the increment() function under the hood but the error message that is thrown would have to change too... that's why I did it this way

davimacedo:
I am not against of having this code here as well. @TomWFox what does our governance say in the case nobody is against? :)

TomWFox:
@davimacedo Considering we can’t reach a lazy consensus we should take a vote, and if it’s a draw then I suggest we include other contributors in the vote.

If you are a PMC member react to this comment with a thumbs up (👍) to vote in favour of merging this PR and a thumbs down (👎) to vote against.

TomWFox:
Other members of the community are welcome to vote / express and opinion but at this stage only the votes of PMC members are binding.

dplewis:
I'll vote for it but some considerations.

`object.decrement('field', -2)` What should the result be?

For a feature like this is it worth it to update every SDK, API Reference and Guide? I can help update the other repos.

stevestencil:
that would be the same as object.increment('field', 2)

dplewis:
Making it positive input only removes unforeseen events like that.

I can only assume parse never had a decrement function because it is based off MongoDB.
Since Mongo only has a [$inc](https://docs.mongodb.com/manual/reference/operator/update/inc/) operator, parse followed the same. Maybe it would be different if it also had a $dec operator.

stevestencil:
I see how it can be confusing but the same could be said about object.increment('field', -1)

You make a good point about MongoDB not having the decrement option. What made me submit this PR was I needed to decrement a field and assumed since we had increment(), we would also have decrement(). I was actually surprised when I didn't see it in the docs (same with destroy() missing). I think a good argument for having this in here is we have a way to increment a field by 1 by just calling object.increment('field'), but we do not have a way to decrement a field without also having to pass in -1. This PR adds that ability.

dplewis:
@acinader @davimacedo Can you have a look at the discussion and vote?

davimacedo:
Just voted. I don't see so much upside on merging but almost no downside as well. I think it is ok to have it in the SDK.

acinader:
Davi's comment is convincing to me.  I don't think we need to update the guide or worry about the other SDKs as this is a minor variation and there are other minor variations.

I appreciate the initiative of @stevestencil

dplewis:
@stevestencil Thanks for the PR. Sorry it took so long to get in.

PR #1070, opened 2020-01-22
Metadata tags

stevestencil:


stevestencil:
This adds support for metadata and tags to be set from the client. The previous PR that was merged in only allowed for metadata and tags to be set from the beforeSaveFile hook.

I had to change saveFile to not call RESTController().ajax() directly because it didn't allow any params to be passed in. Now when calling saveFile it converts the data into base64 and then calls the saveBase64() function.

dplewis:
I'm going to do a PR to support saving and retrieving metadata on MongoDB so we can add unit tests for this feature.

stevestencil:
It would be nothing to add it to the toJSON() function but if I remember correctly isn't only the file name stored in mongo?

dplewis:
Yes but you are forgetting about fs.chunk and fs.files

https://docs.mongodb.com/manual/core/gridfs/
```
{
  "_id" : <ObjectId>,
  "length" : <num>,
  "chunkSize" : <num>,
  "uploadDate" : <timestamp>,
  "md5" : <hash>,
  "filename" : <string>,
  "contentType" : <string>,
  "aliases" : <string array>,
  "metadata" : <any>, // 
}
```
https://mongodb.github.io/node-mongodb-native/3.2/api/GridFSBucket.html#openUploadStream

In node you just add an option to bucket.openUploadStream('fileName', { metadata }); and retrieve it the same way.

stevestencil:
interesting... at first glance, does this mean files would be stored in the database and not S3?

dplewis:
I'm sorry by `saving and retrieving metadata on MongoDB` I mean saving and retrieving metadata on GridFS and GridStore the default storage adapter.

stevestencil:
So the changes I have in this PR is to get the metadata and tags from the client to the server... the reason I had to convert to base64 is because it was being sent directly through the ajax function as a Buffer of data. There was no way to add additional data to be sent with it. Do you have another suggestion to do this?

stevestencil:
@dplewis are there any changes needed for this PR?

dplewis:
@stevestencil Sorry for the late reply. I have to add a PR to the server repo in order to add test cases for this.

src/ParseFile.js
```diff
@@ -418,39 +418,45 @@ const DefaultController = {
...
-      const currentUser = await userController.currentUserAsync();
-      sessionToken =  currentUser ? currentUser.getSessionToken() : undefined;
-    }
-    if (sessionToken) {
-      headers['X-Parse-Session-Token'] = sessionToken;
-    }
-    let url = CoreManager.get('SERVER_URL');
-    if (url[url.length - 1] !== '/') {
-      url += '/';
+    if (source.base64) {
```
dplewis: Is there a case where this would happen?

src/ParseFile.js
```diff
@@ -418,39 +418,45 @@ const DefaultController = {
...
     }
-    url += 'files/' + name;
-    return CoreManager.getRESTController().ajax('POST', url, source.file, headers, options).then(res=>res.response)
+    const base64Data = await new Promise((res, rej) => {
+      // eslint-disable-next-line no-undef
+      const reader = new FileReader();
+      reader.readAsDataURL(source.file);
+      reader.onload = () => res(reader.result);
+      reader.onerror = error => rej(error);
+    });
```
dplewis: Would there be any overhead sending base64 to the server from client side? Just wanted to ask

src/ParseFile.js
stevestencil: Yes... when a new `Parse.File` is created with byte array it is converted to base64. Also when you call `save()` on a `Parse.File` that was created with a uri, it will download it to base64 before calling `saveFile()` in the controller

dplewis: In that case wouldn't `source.format` be `base64` instead of `file`

dplewis:
@stevestencil I can have a look at the unit tests if you want.

stevestencil:
@dplewis Feel free! I've been trying to get around to this all week but I have been slammed with work

PR #1071, opened 2020-01-24
Fix addIndex annotation in ParseSchema

JeffGuKang:
Fix `addIndex` comments.
But I am not sure how to use `indexing` exactly in parse server. 
I will wait any suggestion about it. :)


src/ParseSchema.js
```diff
@@ -221,9 +221,13 @@ class ParseSchema {
   /**
    * Adding an Index to Create / Update a Schema
    *
-   * @param {String} name Name of the field that will be created on Parse
-   * @param {String} type Can be a (String|Number|Boolean|Date|Parse.File|Parse.GeoPoint|Array|Object|Pointer|Parse.Relation)
+   * @param {String} name Name of the index
+   * @param {Object} index {[fieldName: string]: string | number | boolean | Date | File | GeoPoint | any[] | object | Pointer | Relation}
    * @return {Parse.Schema} Returns the schema, so you can chain this call.
```
dplewis: Oops like like a duplicate of `addField`, `index` is just an object. { field: value }

PR #1078, opened 2020-02-06
Fix cascadeSave=false bug

temaput:
On SDK 2.11 saving new objects with option cascadeSave=false returns object without id (id = undefined). 
Rationale:
_handleSaveResponse -> _migrateId demands _localId AND serverId to be defined
_localId is generated with _getId()
normally _getId() is called in object.save() method by calling unsavedChildren()
but when cascadeSave=false unsavedChildren is not called -> _localId is undefined -> migrateId does not save serverId from created object

Solution:
since _localId is needed in controller.save() method, just call _getId() there

### DIFF
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -2335,9 +2335,10 @@ const DefaultController = {
       });
 
     } else if (target instanceof ParseObject) {
-      // copying target lets Flow guarantee the pointer isn't modified elsewhere
-      target._getId();
+      // generate _localId in case if cascadeSave=false
+      target._getId(); 
       const localId = target._localId;
+      // copying target lets Flow guarantee the pointer isn't modified elsewhere
       const targetCopy = target;
       const task = function() {
         const params = targetCopy._getSaveParams();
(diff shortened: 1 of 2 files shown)


davimacedo:
@temaput thanks for the PR. Can you please also add a test?

temaput:
@davimacedo I added the test and discovered that bug only reproduces when singleInstance = false. So I had to put it in test case.

src/ParseObject.js
```diff
@@ -2335,8 +2335,10 @@ const DefaultController = {
       });
 
     } else if (target instanceof ParseObject) {
-      // copying target lets Flow guarantee the pointer isn't modified elsewhere
+      // generate _localId in case if cascadeSave=false
+      target._getId();
```
davimacedo: Maybe we should check if it is cascadeSave and singleInstance = false to avoid calling this method when not necessary? What do you think?

temaput: I think it's better call it once anyway to make sure the _localId is generated. Many things depend on it and so it seems wrong that it's only generated as a side effect of other calls (like _getStateIdentifier or unsavedChildren)

PR #1080, opened 2020-02-07
Change username signup error

JeromeDeLeon:
It was hard to determine that `name` actually refer to `username`. Can we make it specific?


dplewis:
Can you check the failing Travis build? You have a few failing test.

Issue #922, opened 2019-09-13
Performing batch transactions


JeromeDeLeon:
**Is your feature request related to a problem? Please describe.**
This was already supported in the PServer, but in JS-SDK, it limits us to only perform either save or delete but not both.

**Describe the solution you'd like**
It would be better if we could constructr a batch operation like:
```js
await Parse.Object.performTransaction(
  [
    { method: "PUT", objects: [PObject] },
    { method: "DELETE, object: PObject },
    { method: "POST", object: [PObject, PObject] }
  ],
  { transaction: true, useMasterKey: true }
);
```
and performs that sequentially.

**Describe alternatives you've considered**

**Additional context**
Also, using relations when doing batch operation is not allowed because you cannot add something to relation when it is new. I don't know if there's already a solution to this but my current implementation is to use `forkJoin([DeleteObjects(), SaveObjects()])` using RxJS.

JeromeDeLeon:
Wait. I just saw that PS already incorporated the `transactionalSession`. Is there a way for PS to send `transactionalSession` to SDK? same with `commit` and `abort`? something like:
` GET "/transaction"`, `POST "/transaction/(commit | abort)`

davimacedo:
No.. there is no way to do it. Since the Parse Server can run in multiple processes at the same time, there is no warranty that two requests will go to the same process. In the case they go to different processes, you will not be able to use the same MongoDB session.

So, the solution is actually use the `/batch` endpoint. I mean... when the JS SDK call `transaction.commit();`, we will have to put all requests together and send in a single request through the `/batch` endpoint. Got it?

JeromeDeLeon:
```js
// create unique ID or something
const transaction = Parse.Obejct.createTransaction();

// These operations will not be run until commit() is called
// maybe save it to array of operations
someObject.destroy({ useMasterKey: true, transaction });
someObject.save({ transaction });
 
try {
// since /batch already have abort operations no need to manually abort
transaction.commit();
} catch(e) {
console.log(e);
}
```

davimacedo:
Yes. That's the idea. We will probably need the abort in the client side as well to revert the object states.

mfkenson:
Currently with the JS SDK (master, 2.11.0), there is no way to pass parameter "transaction" to the Parse Server. (Correct me if I am wrong) so the transactional stuff (which have been done on Parse Server) does not work with the JS SDK.

My plan of attack is to make a PR which includes:
[...]

PR #1090, opened 2020-02-15
Pass parameter "transaction" to the Parse Server

mfkenson:
https://github.com/parse-community/Parse-SDK-JS/issues/922

Did I miss anything?


src/RESTController.js
```diff
@@ -230,6 +232,11 @@ const RESTController = {
       }
     }
 
+    const transaction = options.transaction;
+    if (transaction) {
+        payload.transaction = true;
```
JeromeDeLeon: You might want to lint your code.

mfkenson: I am quite new to this and don't know what needs to be done exactly. Can you give me some advises how to get started?

JeromeDeLeon: Just run `npm run lint` or `yarn lint` If you are using yarn

davimacedo:
Nice! Thanks for contributing. Could you please add a test case?

src/RESTController.js
dplewis: Check out the [Contributing Guide](https://github.com/parse-community/Parse-SDK-JS/blob/master/CONTRIBUTING.md) for all your testing and linting needs.

dplewis:
@davimacedo Can you take this over and add the test cases? You know more about how transactions in parse work better than anyone.

davimacedo:
Sure. I will work on this.

dplewis:
@davimacedo Any updates?

davimacedo:
Sorry for the delay. I am working on this now.

davimacedo:
@dplewis @mfkenson I started worked on the tests here but there is a problem with this implementation. In the case you have more objects to be saved/destroyed than the batch size, the SDK will split it in more than one batch and it will be generated two transactions in the server side. I think we can solve it using two different approaches:
1. Change the sdk to never split into batches when the transaction option is set to true
2. Throw an error message when the number of objects to be saved/destroyed is bigger than the batch size and the transaction option is set to true.

Thoughts?

dplewis:
@davimacedo I think that would be a great approach. I tried that [here](https://github.com/parse-community/Parse-SDK-JS/pull/1047/files) and ran into an issue. The issue is that we wouldn't be be able to find cycles in `SaveAll`. I'm not sure the backend is setup to catch cycles.

davimacedo:
I've just checked this problem and it will be very hard to be solved now. The not saved dependent objects need to go in a first batch because there is no warranty they will be executed in first place in the server side if we send them all together with the objects that depend on them. And if we have more batched sent to the server separately we cannot ensure the transaction. So, to simplify the solution and have a first version of this feature in the next release I suggest we simply do two verifications:

- Verify if there is any not saved dependent object when transaction=true and throw an error in the case they exist.
- Verify if there are more objects to be saved than the batch size and throw an error in the case there are.

Do you agree @dplewis ?

dplewis:
Thats the only solution I can come up with. We will have to document it well.

andreasanta:
@mfkenson Can you please grant me access to this PR so I can fix linting and maybe try and perform some tests?

@davimacedo Any suggestions on how to start writing tests for this? :)

davimacedo:
Hi guys. I am working on this (to include the validations and the tests) but I found out a bug in Parse Server when deleting objects in a transaction. I am working now on a first PR to Parse Server before fixing this one here.

santiagosemhan:
+1

andreasanta:
@davimacedo Thanks. Please let me know whether I can be of assistance.

vahidalizad:
Hi guys, Great work so far
I really had to use this feature so I came with the following code:
```
function getServerUrlPath() {
  let serverUrl = SERVER_URL;
  if (serverUrl[serverUrl.length - 1] !== '/') {
    serverUrl += '/';
  }
  const url = serverUrl.replace(/https?:\/\//, '');
  return url.substr(url.indexOf('/'));
}

const MyCustom = Parse.Object.extend('MyCustom');
let obj1 = new MyCustom({name: 'just a name1'});
let obj2 = new MyCustom({name: 'just a name2'});

let batch = [obj1, obj2];
let options = {};

Parse._request(
  'POST',
  'batch',
  {
    requests: batch.map(obj => {
      const params = obj._getSaveParams();
      params.path = getServerUrlPath() + params.path;
      return params;
    }),
    transaction: true,
  },
  options
)
  .then(res => console.log('success', res))
  .catch(e => console.log('failed', e));
```
But there is something I don't understand when it fails, it sends up to `REQUEST_ATTEMPT_LIMIT` retries request but when I save only one file with `await obj1.save()` it only does it once and log it once. maybe this is because the batch request error does not return `Parse.Error` and the result has a code above 500.
Is this intended and I have to change REQUEST_ATTEMPT_LIMIT somehow?

davimacedo:
Maybe. You can set it via `CoreManager.set('REQUEST_ATTEMPT_LIMIT', retry);`

vahidalizad:
`Parse.CoreManager.set('REQUEST_ATTEMPT_LIMIT', 1);` changes every request retry times to 1 and it works, but there is no way to set a retry value for a single request.
but thanks anyway it works.
Maybe you could change the `src/ParseServerRESTController.js` in the definition of `ParseServerRESTController` like bellow to return the error like the normal batch requests

```
if (path === "/batch") {
    let initialPromise = Promise.resolve();
    if (data.transaction === true) {
    initialPromise = config.database.createTransactionalSession();
    }
    return initialPromise.then(() => {
    const promises = data.requests.map((request) => {
        return handleRequest(
        request.method,
        request.path,
        request.body,
        options,
        config
        ).then(
        (response) => {
            return Promise.resolve({ success: response });
        },
        (error) => {
            return Promise.resolve({
            error: { code: error.code, error: error.message },
            });
        }
        );
    });
    return Promise.all(promises)
        .catch((error) => {
        if (data.transaction === true) {
            return config.database.abortTransactionalSession().then(() => {
            throw error;
            });
        } else {
            throw error;
        }
        })
        .then((result) => {
        if (data.transaction === true) {
            let error = result.find((t) => t.error && t.code);
            if (error) {
            return config.database.abortTransactionalSession().then(() => {
                return error;
            });
            }
            return config.database.commitTransactionalSession().then(() => {
            return result;
            });
        } else {
            return result;
        }
        });
    });
}
```

davimacedo:
That makes sense. Would you be willed to continue this PR?

vahidalizad:
> 
> 
> That makes sense. Would you be willed to continue this PR?

It's not about this PR, I have to create a pr on parse-server project, I create a new fork and when it's finished I notify you.
After I finished that I'm gonna join you in this one if you haven't finished it yet.

RahulLanjewar93:
@davimacedo Hi guys, any updates on this?

araskinwhichbnb:
Its Sep 2023 and there is an ominous comment above that says someone deleted this feature. Does that mean we are not going to get Transactional Integrity in the JS SDK?
@davimacedo

RahulLanjewar93:
@mfkenson just wanted to know if there's any other pr/issue for reference since you closed this one.

Llorx:
@mfkenson ???

Llorx:
> Hi guys. I am working on this (to include the validations and the tests) but I found out a bug in Parse Server when deleting objects in a transaction. I am working now on a first PR to Parse Server before fixing this one here.

@davimacedo Hey, have you fixed this problem? Can I create a new PR with the transaction: true flag now?

vahidalizad:
> That makes sense. Would you be willed to continue this PR?

@davimacedo Thank you for your patience! I apologize for the delay. it took me three years to get back to this 😅. I was surprised that this hadn't been completed yet, so I went ahead and created a PR to address it.

Here’s the link to the PR: [#2265](https://github.com/parse-community/Parse-SDK-JS/pull/2265).

Issue #1092, opened 2020-02-22
reactive-native build not working


davimacedo:
### Issue Description

react-native build is not working

### Steps to reproduce

```js
import React from 'react';
import { StyleSheet, Text, View } from 'react-native';
import Parse from 'parse/react-native';
import { AsyncStorage } from 'react-native'; 

Parse.setAsyncStorage(AsyncStorage);
Parse.serverURL = 'myurl';
Parse.initialize('myAppId', 'myJsKey');

export default class App extends React.PureComponent {
  constructor(props) {
    super(props);
    this.state = {
      isLoading: true,
      myObjects: undefined,
      error: undefined,
    };
  }

  async componentDidMount() {
    const isLoading = false;
    let myObjects;
    let error;
    const myQuery = new Parse.Query('MyClass');
    try {
      myObjects = await myQuery.find();
    } catch (e) {
      error = e.toString();
    }
    this.setState({
      isLoading,
      myObjects,
      error
    });
  }

  render() {
    return (
      <View style={styles.container}>
        {this.state.isLoading ? (
          <Text>Loading...</Text>
        ) : this.state.error ? (
          <Text>{error}</Text>
        ) : this.state.myObjects.map(myObject => (
          <Text key={myObject.id}>{myObject.get('myField')}</Text>
        ))}        
      </View>
    );
  }
};

const styles = StyleSheet.create({
  container: {
    flex: 1,
    backgroundColor: '#fff',
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```

### Expected Results

Show up the message I have in my class.

### Actual Outcome

```js
App.js:52 Uncaught TypeError: parse_react_native__WEBPACK_IMPORTED_MODULE_10___default.a.setAsyncStorage is not a function
    at Module.../../../../../../../../../../../Projects/AwesomeProject/App.js (App.js:52)
    at __webpack_require__ (App.js:52)
    at fn (App.js:31)
    at Module.../../../../../../../../../../../Projects/AwesomeProject/node_modules/expo/AppEntry.js (bundle.js:3463)
    at __webpack_require__ (App.js:52)
    at fn (App.js:31)
    at Object.0 (bundle.js:57905)
    at __webpack_require__ (App.js:52)
    at App.js:52
```

### Other comments

- If I change the import line to `import Parse from 'parse';` it works in my web-browser emulator.
[...]

PR #1094, opened 2020-02-22
Fix react-native build

davimacedo:
Close https://github.com/parse-community/Parse-SDK-JS/issues/1092

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,10 +49,10 @@ For server-side applications or Node.js command line tools, include `'parse/node
 const Parse = require('parse/node');
 ```
 
-For React Native applications, include `'parse/react-native'`:
+For React Native applications, include `'parse/react-native.js'`:
 ```js
 // In a React Native application
-const Parse = require('parse/react-native');
+const Parse = require('parse/react-native.js');
 
 // On React Native >= 0.50 and Parse >= 1.11.0, set the Async
 const AsyncStorage = require('react-native').AsyncStorage;
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "jest": "24.9.0",
     "jsdoc": "3.6.3",
     "jsdoc-babel": "0.5.0",
-    "metro-react-native-babel-preset": "^0.58.0",
+    "metro-react-native-babel-preset": "0.58.0",
     "parse-server": "github:parse-community/parse-server#master",
     "regenerator-runtime": "0.13.3",
     "vinyl-source-stream": "2.0.0"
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -49,10 +49,10 @@ For server-side applications or Node.js command line tools, include `'parse/node
 const Parse = require('parse/node');
 ```
 
-For React Native applications, include `'parse/react-native'`:
+For React Native applications, include `'parse/react-native.js'`:
 ```js
 // In a React Native application
-const Parse = require('parse/react-native');
+const Parse = require('parse/react-native.js');
```
davimacedo: @dplewis before you ask, it is weird, but it does not work on react-native if imported as `require('parse/react-native')`. It does import `require('parse/reactnative')` if we change the name of the file to `reactnative.js`. I don't know exactly. Probably some bug in their side for resolving files with `-`.

dplewis: How did you figure this out?

davimacedo: I used `npm link` to install the SDK in a test react native project that I built and I did some tests. In one of them, I added a `console.log` in the `react-native.js` file and it is only printing out when I import using `require('parse/react-native.js');`

dplewis: I make it a habit of adding file extensions in my projects. One of the shortcomings of node.

davimacedo:
Tests are now passing. I had to upgrade mongodb in travis from 3.2 to 3.6 (same as parse-server) since 3.2 does not support collation which is a recent requirement of parse-server to run.

davimacedo:
wait... it worked well in the browser simulator but not working in iOS. I will do some more tests and let you know.

dplewis:
Oh I didn’t realize that 3.6 is the requirement

davimacedo:
I believe that collation is now required in Parse Server since https://github.com/parse-community/parse-server/commit/fd0b535159877fa9bc74fb645169f56814c67c60

davimacedo:
The problem now is https://github.com/back4app/Parse-SDK-JS/commit/684b9e1a16a3dd0e5530d04675e4a8852177d5d0#diff-fb0c08f76d82dd2379cda467e5d0dbe6

crypto-js is not available for react native project. I'm trying to figure out what to do.

davimacedo:
@dplewis it looks good now. Can you please take a look? I tested in the device and it is working fine even using:

```
Parse.enableEncryptedUser();
Parse.secret= 'my Secrey Key';
```

package.json
```diff
@@ -69,6 +70,7 @@
     "jest": "24.9.0",
     "jsdoc": "3.6.3",
     "jsdoc-babel": "0.5.0",
+    "metro-react-native-babel-preset": "^0.58.0",
```
dplewis: Can you pin the packages?

davimacedo:
Sorry for the delay. I've just pinned the package.

PR #1112, opened 2020-03-11
Make iteration query methods respect returned promises

noahsilas:
These methods (`ParseQuery.map`, `ParseQuery.filter`, and `ParseQuery.reduce`) are documented as behaving similarly to `ParseQuery.each` (which they rely on under the hood): if the callback they are passed returns a promise, we should:
- wait for that promise to resolve before continuing iteration
- if that promise rejects, we should stop iteration and return a rejected promise

These methods weren't living up to that contract. Here's a series of adjustments to each one so that they follow those guidelines.

### DIFF
--- a/src/ParseQuery.js
+++ b/src/ParseQuery.js
@@ -1074,11 +1074,12 @@ class ParseQuery {
     const array = [];
     let index = 0;
     await this.each((object) => {
-      const flag = callback(object, index, this);
-      if (flag) {
-        array.push(object);
-      }
-      index += 1;
+      return Promise.resolve(callback(object, index, this)).then((flag) => {
+        if (flag) {
+          array.push(object);
+        }
+        index += 1;
+      });
     }, options);
     return array;
   }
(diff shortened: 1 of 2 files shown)


src/ParseQuery.js
```diff
@@ -1028,11 +1030,22 @@ class ParseQuery {
...
    */
   async reduce(callback: (accumulator: any, currentObject: ParseObject, index: number) => any, initialValue: any, options?: BatchOptions): Promise<Array<any>> {
-    const objects = [];
+    let accumulator = initialValue;
+    let index = 0;
     await this.each((object) => {
-      objects.push(object);
+      // If no initial value was given, we take the first object from the query
+      // as the initial value and don't call the callback with it.
+      if (index === 0 && initialValue === undefined) {
```
acinader: [Calling reduce() on an empty array without an initialValue will throw a TypeError.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce)

should we handle it the same way?

noahsilas: Good point! Updated to have similar behavior in 687e795. Thanks :smile:

santiagosemhan:
@TomWFox Is this documented?

dplewis:
@samuelantonioli They are documented in the API

https://parseplatform.org/Parse-SDK-JS/api/2.12.0/Parse.Query.html#map

PR #1114, opened 2020-03-11
Add `query.eachBatch()` to Parse.Query API

noahsilas:
When a query has a large enough result set that we don't want to
materialize it at once, the SDK provides `query.each()`, which yields
each matching object to a processor, one at a time. This is a handy
tool, but if the number of records to process is really so large, we
also can take advantage of batching the processing. Compare the
following operations:

```js
// Processing N items involves N calls to Parse Server
new Parse.Query('Item').each((item) => {
  item.set('foo', 'bar');
  return item.save();
})

// Processing N items involves ceil(N / batchSize) calls to Parse Server
const batchSize = 200;
new Parse.Query('Item').eachBatch((items) => {
  items.forEach(item => item.set('foo', 'bar'));
  return Parse.Object.saveAll(items, { batchSize });
}, { batchSize });
```

The `.each()` method is already written to do fetch the objects in
batches; we effectively are splitting it out into two:
- `.eachBatch()` does the work to fetch objects in batches and yield
  each batch
- `.each()` calls `.eachBatch()` and handles invoking the callback for
  every item in the batch

Aside: I considered adding the undocumented `batchSize` attribute
already accepted by `.each()`, `.filter()`, `.map()` and `.reduce()` to
the public API, but I suspect that at the time that you are performance
sensitive enough to tune that parameter you are better served by
switching to `eachBatch()`; the current implementation of `.each()` is
to construct a promise chain with a node for every value in the batch,
and my experience with very long promise chains has been a bit
frustrating.

### DIFF
--- a/src/ParseQuery.js
+++ b/src/ParseQuery.js
@@ -856,15 +856,16 @@ class ParseQuery {
   }
 
   /**
-   * Iterates over each result of a query, calling a callback for each one. If
-   * the callback returns a promise, the iteration will not continue until
+   * Iterates over objects matching a query, calling a callback for each batch.
+   * If the callback returns a promise, the iteration will not continue until
    * that promise has been fulfilled. If the callback returns a rejected
-   * promise, then iteration will stop with that error. The items are
-   * processed in an unspecified order. The query may not have any sort order,
-   * and may not use limit or skip.
+   * promise, then iteration will stop with that error. The items are processed
+   * in an unspecified order. The query may not have any sort order, and may
+   * not use limit or skip.
    * @param {Function} callback Callback that will be called with each result
    *     of the query.
    * @param {Object} options Valid options are:<ul>
+   *   <li>batchSize: How many objects to yield in each batch (default: 100)
    *   <li>useMasterKey: In Cloud Code and Node only, causes the Master Key to
    *     be used for this request.
    *   <li>sessionToken: A valid session token, used for making a request on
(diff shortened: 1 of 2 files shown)


dplewis:
There is a global batch size that you can set.

https://github.com/parse-community/Parse-SDK-JS/pull/1053/files Can you add it?

I agree with you about the promise chain.

noahsilas:
I personally expected the default to have each batch have the same number of results as a query that doesn't have a limit applied, which I think is based on a default in parse-server that is set to 100?

I guess it feels like there's two somewhat-related values: the default page size when fetching, and the default batch size when we're submitting multiple mutations. One option would be for this method to change its name to something like `eachPage` and have a `pageSize` argument, which would help make it clearer that you shouldn't expect this to respect a configured batch size, but it's also sort of nicely consistent for all these methods that are dealing with chunks of objects to use the same parameter name, so that there's never confusion about which parameter to use for a given method.

I would cast a small vote for leaving this as-is, but I'm happy to make the change if you think that's best. Let me know @dplewis !

If we do move ahead on updating this to use the global batch size configuration, do you think that we should let `ParseQuery.each()` inherit that default batch size, or should we add an override there to keep its default at 100 to preserve its current behavior?

Thanks for your feedback folks! 😄

dplewis:
Every SDK has a default batch size for example saveAll has a default 20 here but 40 in PHP. This was left over from parse.com days. We can keep your changes as is  I forget the reason why I didn’t add it .each when I implemented global batch size.

Issue #1096, opened 2020-02-25
File upload progress method was called once


JeromeDeLeon:
### Issue Description

I'm trying to track the progress of the file being uploaded to the server, but it seemed like `progress` method was only called once after the file was successfully uploaded to the server resulting the call to `1 213 213`. I tried uploading three different files, starting from 500kB then 5mB and 12mB files. What I got was single call from each file.

```js
1 243 243 // File 1
1 215 215 // File 2
1 251 251 // File 3
```

### Steps to reproduce

1. Create a sample file either Blob or File instance is fine.
2. Add that sample file to Parse File.
3. Save it but pass a callback function to `progress`

```js
const fileObj = new Parse.File('sample', file);
fileObj.save({ progress: console.log })
```

### Expected Results

The log will start from 0 then gradually increases until it reaches a 100% or 1

### Actual Outcome

Log called only once with 1 as the output

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.10.0
  - Operating System:    Linux
  - Hardware:             [FILL THIS OUT]
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Digital Ocean

- **JS SDK**
  - JS SDK version: 2.11.0
  - Application? (Browser, Node, React-Native, etc): Browser

dplewis:
Which browser are you using?

progress is handled [here](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/RESTController.js#L162)

JeromeDeLeon:
Chrome Browser latest version

JeromeDeLeon:
> Which browser are you using?
> 
> progress is handled [here](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/RESTController.js#L162)

Yes, I saw that one and even the latest changes and tests. That's why I'm wondering why was it called only once.

ladigitale:
According to [Mozzilla](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest#Monitoring_progress)  
> The upload events are fired on the XMLHttpRequest.upload object
So, parse supports only progress download yet.

JeromeDeLeon:
I don't think so. I believe this feature worked before in our previous app that's why I was wondering what happened.

JeromeDeLeon:
I'll try to investigate more.

ladigitale:
[...]

PR #1133, opened 2020-04-02
Fix file upload progress

JeromeDeLeon:
Based on the discussion from [#1096](https://github.com/parse-community/Parse-SDK-JS/issues/1096)

Thanks to @ladigitale for the advice. :+1:

### DIFF
--- a/src/RESTController.js
+++ b/src/RESTController.js
@@ -159,28 +159,27 @@ const RESTController = {
         headers[key] = customHeaders[key];
       }
 
-      function handleProgress(type, event) {
+      function handleProgress(event) {
         if (options && typeof options.progress === 'function') {
           if (event.lengthComputable) {
             options.progress(
-              type,
               event.loaded / event.total,
               event.loaded,
               event.total
             );
           } else {
-            options.progress(type, null);
+            options.progress(null);
           }
         }
       }
 
       xhr.onprogress = (event) => {
-        handleProgress('download', event);
+        handleProgress(event);
       };
 
       if (xhr.upload) {
         xhr.upload.onprogress = (event) => {
-          handleProgress('upload', event);
+          handleProgress(event);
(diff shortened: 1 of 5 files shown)


JeromeDeLeon:
I tested it without `type` and worked just fine. No duplication. Maybe the browser handles the calling of progress behind the scene if it should be calling progress for upload or progress for download.

JeromeDeLeon:
I was thinking the same thing like putting `loaded` and `total` into an object allowing the future values to be wrapped into an object or putting `type` as an object at the end of the parameter but then again, the `type` could possibly remove if we could determine about the collision of `upload progress` and `download progress`. If you could test it on your end, that would be great help :+1:

src/RESTController.js
```diff
@@ -158,15 +158,32 @@ const RESTController = {
...
+            options.progress(type, null);
           }
         }
+      }
+
+      xhr.onprogress = (event) => {
+        handleProgress('download', event);
       };
+
+      if (xhr.upload) {
```
dplewis, line 181: Oops I forgot why this was removed. The type parameter can be removed here and we should be good to go.

ladigitale:
I just found the time to test the pacth.
As expected at the end of the upload with this code you will yet have 2 consecutive callbacks of the same function with different meanings : 

1 14535341 14535341   (meaning end of the upload)
1 165 165 (meaning end of the download sent as one chunk because of the low size of the response)

Thats why aditionnal imformations / api evolution would have been usefull.

however I really thank you for that work.

ladigitale:
I hope I don't seem arrogant while writing. English is not my best part.
:)

JeromeDeLeon:
Seems like `type` is really necessary there to distinguish the difference of the two (upload and download). I'll try to push another PR and let you know about it.

PR #1140, opened 2020-04-07
Fix File Progress Type

JeromeDeLeon:
According to @ladigitale from [last PR](https://github.com/parse-community/Parse-SDK-JS/pull/1133)

> I just found the time to test the pacth.
> As expected at the end of the upload with this code you will yet have 2 consecutive callbacks of the same function with different meanings :
> 
> 1 14535341 14535341 (meaning end of the upload)
> 1 165 165 (meaning end of the download sent as one chunk because of the low size of the response)
> 
> Thats why aditionnal imformations / api evolution would have been usefull.

### DIFF
--- a/src/RESTController.js
+++ b/src/RESTController.js
@@ -163,13 +163,13 @@ const RESTController = {
         if (options && typeof options.progress === 'function') {
           if (event.lengthComputable) {
             options.progress(
-              type,
               event.loaded / event.total,
               event.loaded,
-              event.total
+              event.total,
+              { type }
             );
           } else {
-            options.progress(type, null);
+            options.progress(null, null, null, { type });
           }
         }
       }
--- a/src/RESTController.js
+++ b/src/RESTController.js
@@ -159,27 +159,28 @@ const RESTController = {
         headers[key] = customHeaders[key];
       }
 
-      function handleProgress(event) {
+      function handleProgress(type, event) {
         if (options && typeof options.progress === 'function') {
           if (event.lengthComputable) {
             options.progress(
+              type,
               event.loaded / event.total,
               event.loaded,
               event.total
             );
           } else {
-            options.progress(null);
+            options.progress(type, null);
           }
         }
       }
 
       xhr.onprogress = (event) => {
-        handleProgress(event);
+        handleProgress('download', event);
       };
 
       if (xhr.upload) {
         xhr.upload.onprogress = (event) => {
-          handleProgress(event);
+          handleProgress('upload', event);
         }
       }
 
--- a/src/__tests__/RESTController-test.js
+++ b/src/__tests__/RESTController-test.js
@@ -444,8 +444,8 @@ describe('RESTController', () => {
     jest.spyOn(options, 'progress');
 
     RESTController.ajax('POST', 'files/upload.txt', {}, {}, options).then(({ response, status }) => {
-      expect(options.progress).toHaveBeenCalledWith(0.5, 5, 10);
-      expect(options.progress).toHaveBeenCalledTimes(2);
+      expect(options.progress).toHaveBeenCalledWith('download', 0.5, 5, 10);
+      expect(options.progress).toHaveBeenCalledWith('upload', 0.5, 5, 10);
       expect(response).toEqual({ success: true });
       expect(status).toBe(200);
       done();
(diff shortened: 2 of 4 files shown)


JeromeDeLeon:
I was pushing it at the beginning, or

```js
progress(event.loaded / event.total, {
  type,
  event
});
```

ladigitale:
Is there any issue to put the type at last position, for backward compatibility ?

src/RESTController.js
```diff
@@ -159,27 +159,28 @@ const RESTController = {
         headers[key] = customHeaders[key];
       }
 
-      function handleProgress(event) {
+      function handleProgress(type, event) {
         if (options && typeof options.progress === 'function') {
           if (event.lengthComputable) {
             options.progress(
```
JeromeDeLeon, line 165: I pushed it at the beginning, or

```js
progress(event.loaded / event.total, {
  type,
  event
});
```

@dplewis , @ladigitale let me know your thoughts

JeromeDeLeon: I do think there is. I was reluctant to just add it and the end, because if somehow, why not just enclose it in object to allow future values instead of
```js
progress(progressValue, loaded, total, type, ...restArgs)
```

ladigitale: it would be perfect.
In fact when we've got the original event i think we've got the type.
The targets of each event have a different type who gives the needed imformation.

JeromeDeLeon: Is that so? Have you tried it? haven't looked into values of `event` tho.

ladigitale: I just logged the target propety one is XMLHttpRequest, and the other is XMLHttpRequestUpload

JeromeDeLeon: Feel free to make suggestion on the PR using the suggestion feature of Github. :smile:

src/RESTController.js
```diff
@@ -159,27 +159,28 @@ const RESTController = {
...
+              type,
               event.loaded / event.total,
               event.loaded,
               event.total
```
ladigitale, line 169: we could keep the 3 first parameters as is for backward compatibility, then we could add an object that can handle any future addition. we would just give the original event in it.
By testing original event.target people could infere if it's upload or download phase.
```
event.loaded / event.total,
event.loaded,
event.total,
{originalEvent:event}
```
But i'm ok with 
```
event.loaded / event.total,
event.loaded,
event.total,
type
```

and with
```
event.loaded / event.total,
event.loaded,
event.total,
event
```
for example.

JeromeDeLeon: for determining the type of progress, we can use `event.target`, right?

ladigitale: Yes, thats right.

JeromeDeLeon: Upon checking, I tried logging `event` and here's what I got:
image
I could check the instance, but it would be better to tell explicitly rather then leaving us the logic for checking whether it came from `download` or `upload`

ladigitale: yes. I agree with that.
Maybe satisfying and scallable solution would be
`event.loaded / event.total,
event.loaded,
event.total,
{type:type}`
as arguments of the progress function.

dplewis: @ladigitale I see what you are trying to do now. I didn't realize there was a difference. I agree on the backwards compatibility by adding the options to the end.

There are other places in the code that you can add progress for instance file.getData in Parse.File etc

src/__tests__/RESTController-test.js
```diff
@@ -468,7 +468,7 @@ describe('RESTController', () => {
     jest.spyOn(options, 'progress');
 
     RESTController.ajax('POST', 'files/upload.txt', {}, {}, options).then(({ response, status }) => {
-      expect(options.progress).toHaveBeenCalledWith(null);
+      expect(options.progress).toHaveBeenCalledWith(null, null, null, { type: 'upload' });
```
JeromeDeLeon, line 471: @dplewis @ladigitale  probably the reason why I'm kinda against to moving it at the end. This could be fixed when doing major upgrade tho. :smile:

dplewis: I agree with you a general rule of thumb is that more than 2 parameters can be the rest can be replaced with an object. I don't know about a major release just for something like this.

Also can you improve the documentation with an example.

https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseFile.js#L236

JeromeDeLeon: I mean when doing a major upgrade like v5, this could be included has one of the breaking change.

Yes, I'm thinking about that as well. Thanks. :smile:

PR #1144, opened 2020-04-10
Add Email Verification and Verify Password

JeromeDeLeon:
- [x] Request email verification (static method)
- [x] Verify password (both instance and static methods)

@dplewis, mind to take a look on this one? Thanks.


src/ParseUser.js
```diff
@@ -761,6 +783,69 @@ class ParseUser extends ParseObject {
...
+   * Verify whether a given password is the password of the current user.
+   *
+   * @param {String} username  A username to be used for identificaiton
+   * @param {String} password A password to be verified
+   * @param {Object} options
+   * @static
+   * @returns {Promise} A promise that is fulfilled with a user
+   *  when the password is correct.
+   */
+  static verifyPassword(username: string, password: string, options?: RequestOptions) {
```
JeromeDeLeon, line 819: Is using just `username` to verify password good enough? When I took a look on PS, I saw that it had supported both `email` and `username`.

src/ParseUser.js
```diff
@@ -506,6 +506,28 @@ class ParseUser extends ParseObject {
...
 
+  /**
+   * Verify whether a given password is the password of the current user.
+   *
+   * @param {String} password A password to be verified
+   * @param {Object} options
+   * @return {Promise} A promise that is fulfilled with a user
+   *  when the password is correct.
+   */
+  verifyPassword(password: string, options?: RequestOptions): Promise<ParseUser> {
```
JeromeDeLeon, line 517: Can you explain why is there a need for `options` on most methods? I had to it since I saw others had it too although I first tried to implement without it and works just ffine.

JeromeDeLeon:
Some of the tests are failing on my end about invalid session token.

JeromeDeLeon:
I just tried it on my local app and so far so good.

src/ParseUser.js
dplewis: Most methods have options to support ACL's, installationId etc and allow for usability in the future

src/ParseUser.js
```diff
@@ -506,6 +506,28 @@ class ParseUser extends ParseObject {
...
+
+    options = options || {};
+
+    const verificationOption = {};
+    if (options.hasOwnProperty("useMasterKey")) {
+      verificationOption.useMasterKey = options.useMasterKey;
+    }
+
+    const controller = CoreManager.getUserController();
+    return controller.verifyPassword(username, password, verificationOption);
```
dplewis, line 528: You can refactor here and use the static version for both of these.

src/ParseUser.js
dplewis: Can you write integration tests? I see you already have the unit tests.

integration/test/ParseUserTest.js
```diff
@@ -979,4 +979,18 @@ describe('Parse User', () => {
     userData = Parse.Storage.getItem(path);
     expect(JSON.parse(userData).password).toBeUndefined();
   });
+
+  it('can verify user password', async () => {
```
JeromeDeLeon, line 983: Is this not enough? @dplewis 
Let me know what I'm still missing. :smile:

JeromeDeLeon:
@dplewis Do you know why I'm still having an error of invalid session token?

dplewis:
@JeromeDeLeon I'll have a look shortly

JeromeDeLeon:
@dplewis Did you find the culprit for making the test failed? 😶😶

dplewis:
@JeromeDeLeon Can you give me access to your fork?

JeromeDeLeon:
Oops. Forgot. There you go. 😁

JeromeDeLeon:
Can you explain why moving the test up will fix the problem?

dplewis:
Sometimes clean up between tests isn't the best. The tests before directly manipulated local storage which probably caused the following test to mess up. I'll look into it more closely at a later date.

JeromeDeLeon:
Thank you so much for your time on reviewing this. :smile:

stevestencil:
I love this new function however it appears to be returning a plain object instead of an instance of Parse.User

PR #1145, opened 2020-04-10
Add Parse.CLP Object to control ClassLevelPermissions

brunocosta91:
The intent of this pull request is to provide the Parse.CLP object which could be use later to be returned on Parse.Schema.
Currently is possible create a new Parse.CLP(schema.classLevelPermissions) and check permissios of user/role/public.


src/ParseCLP.js
```diff
@@ -0,0 +1,898 @@
...
+ * of your application.</p>
+ * @alias Parse.CLP
+ */
+class ParseCLP {
+  permissionsMap: PermissionsMap;
+
+  /**
+   * @param {(Parse.User|Parse.Role|Object)} userId The user to initialize the CLP for
+   */
+  constructor(userId: ParseUser | ParseRole | PermissionsMap) {
```
JeromeDeLeon, line 51: why not `permissionId`?

JeromeDeLeon:
This is very useful IMO. We had to manually do this ourselves.

brunocosta91:
I found this problem to check permissions on CLP so i decide share my solutions, this is based on ACL.
Just to sumarize my approach (similar to basic vs advanced on parse-dashboard):
write access = update && create && delete && addField access
read access = find && count && get access

src/ParseCLP.js
JeromeDeLeon, line 51: Why not `permissionId`?

src/ParseCLP.js
```diff
@@ -0,0 +1,898 @@
...
+                if (typeof allowed !== 'boolean' && !isValidPermissionExtended) {
+                  throw new TypeError('Tried to create an CLP with an invalid permission value.');
+                }
+                this.permissionsMap[permission][user] = allowed;
+              }
+            }
+          }
+        }
+      }
+    } else if (typeof userId === 'function') {
```
JeromeDeLeon, line 99: This can be moved to the top.

davimacedo:
@brunocosta91 thanks for the PR. It is a great job! I have some questions:
- Would it be possible to also support requiresAuthentication and pointer permissions? @dplewis @acinader am I missing something else?
- Could you provide an example on how it will integrate with the other functions we have here in the SDK? I mean, how would we use the ParseCLP object and setup the CLP of a certain class?
- Could you please increase the test coverage? Only 50% of the ParseCLP.js file is currently covered.
- It would be greate to also have a PR to the JS Docs explaining how to use it.

acinader:
@davimacedo looks good to me.

src/ParseCLP.js
bcostahitachivantara: Because the Parse.CLP has the intention to give a permission to a user/group. The permissions are always: find, get, count, create, update, delete, addField

src/ParseCLP.js
bcostahitachivantara: I believe that will be call 99% with correct parameters, so I do not want waste time doing this validation every call.

src/ParseCLP.js
brunocosta91: Because the Parse.CLP has the intention to give a permission to a user/group. The permissions are always: find, get, count, create, update, delete, addField

brunocosta91: I believe that will be call 99% with correct parameters, so I do not want waste time doing this validation every call.

src/ParseCLP.js
brunocosta91: I believe that will be call 99% with correct parameters, so I do not want waste time doing this validation every call.

brunocosta91:
This weekend I will add more test and try improve PR with suggestions by @davimacedo

dplewis:
@brunocosta91 Any updates on this? It would be awesome to include this in the SDK. Let us know if you need assistance.

brunocosta91:
I will try get some time to improve this PR

dplewis:
@brunocosta91 I can finish this PR up for you. I just need write access to your fork.

dplewis:
@davimacedo @mtrezza I believe this is ready for review.

davimacedo (approved):
LGTM!

PR #1150, opened 2020-04-18
add context to Parse.Object.save

mtrezza:
Exposes the `context` object in `Parse.Object.save()` so that parameters can be made accessible in Cloud Code triggers `beforeSave`, `afterSave`.

Example:

```
const obj = new TestObject();
await obj.save(null, {context: {a: "a"}});

Parse.Cloud.beforeSave("TestObject", req => {
    console.log(req.context.a);
});

Parse.Cloud.afterSave("TestObject", req => {
    console.log(req.context.a);
});  
```
---
Don't merge until

- [ ] Concept agreed on in https://github.com/parse-community/parse-server/issues/6459
- [x] JS doc added in code for ‘context’

### DIFF
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -56,7 +56,8 @@ type SaveParams = {
 };
 
 type SaveOptions = FullOptions & {
-  cascadeSave?: boolean
+  cascadeSave?: boolean;
+  context?: AttributeMap;
 }
 
 // Mapping of class names to constructors, so we can populate objects from the
@@ -1250,6 +1251,9 @@ class ParseObject {
     }
     if (options.hasOwnProperty('installationId') && typeof options.installationId === 'string') {
       saveOptions.installationId = options.installationId;
+    }
+    if (options.hasOwnProperty('context') && typeof options.context === 'object') {
+      saveOptions.context = options.context;
     }
     const controller = CoreManager.getObjectController();
     const unsaved = options.cascadeSave !== false ? unsavedChildren(this) : null;
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -1180,6 +1180,7 @@ class ParseObject {
    *       <li>sessionToken: A valid session token, used for making a request on
    *       behalf of a specific user.
    *       <li>cascadeSave: If `false`, nested objects will not be saved (default is `true`).
+   *       <li>context: A dictionary that is accessible in Cloud Code `beforeSave` and `afterSave` triggers.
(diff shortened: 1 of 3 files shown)


mtrezza:
The context is transmitted as `_context` property in REST request.

- I did not find an existing similar use of internal properties in the REST request, but according to the [docs](https://docs.parseplatform.org/js/guide/#data-storage) a field name must start with a letter, so there cannot be a conflict with an object property:

> Key names must contain only numbers, letters, and underscore, and must start with a letter. 

So using `_context` as internal property in a REST request seems OK, it is also used for internal field names such as `_created_at` in the DB.

mtrezza:
This unidirectionally exposes the internal trigger `context` to the client.
Should this be restricted to masterKey or to within Cloud Code?

@dplewis @acinader any thoughts?

src/__tests__/ParseObject-test.js
```diff
@@ -1219,6 +1219,26 @@ describe('ParseObject', () => {
     });
   });
 
+  it('accepts context on save', async () => {
```
dplewis, line 1222: Looks good, can you add an integration test?

mtrezza: I thought about this, but this PR requires a PR in parse server and vice versa. How would I set up an integration test?

mtrezza: Btw, the PR in parse server has a test case which tests the integration.

dplewis: Oh right I didn't see that.

mtrezza: No worries, let me know if anything else to take a look at...

src/ParseObject.js
```diff
@@ -1251,6 +1252,9 @@ class ParseObject {
     if (options.hasOwnProperty('installationId') && typeof options.installationId === 'string') {
       saveOptions.installationId = options.installationId;
     }
+    if (options.hasOwnProperty('context') && typeof options.context === 'object') {
+      saveOptions.context = options.context;
```
dplewis, line 1256: Can you add documentation above for options?

mtrezza: Ah right, got that as TODO above.

mtrezza: done

mtrezza:
Thanks! Would be great if you could make a release with parse server release that will merge https://github.com/parse-community/parse-server/pull/6626.

Issue #1147, opened 2020-04-12
LiveQuerySubscription "open" event does not fire as expected


travsr:
### Issue Description
When I register an `open` event for a `LiveQuerySubscription`, the event does not fire on the intial websocket `subscribed` event as you'd expect. However, if it does fire if a reconnect happens.

My guess is the fact that you now have to wait for the subscription object to be returned via promise from `query.subscribe()` may be the cause. By the time you are able to register your event, the "open" event  will already have been emitted by `LiveQueryClient`.

My quick fix was to add a slight timeout delay before the `open` event is emitted in `LiveQueryClient.js`, but that seems like more of a hack than a real solution.

Maybe I was just doing something wrong but this seemed like something potentially worth posting here.

### Test Case
```
let query = new Parse.Query("MyClass");
let subscription = await query.subscribe();

subscription.on('open', (response) => {

    // (the stuff in here will not run)
    console.log("SUBSCRIPTION OPENED"); // <-- does not print to console
   
});
```

To accurately reproduce your issue. Add your test cases [here](https://github.com/parse-community/Parse-SDK-JS/tree/master/integration/test) and read the [Contributing Guide](https://github.com/parse-community/Parse-SDK-JS/blob/master/CONTRIBUTING.md) to run the tests. 

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : ^4.2.0
  - Operating System:     macOS 10.15.4
  - Localhost or remote server?: Localhost

- **JS SDK**
  - JS SDK version: ^2.12.0
  - Application?: Browser

dplewis:
You brought up an interesting point. The point of waiting for the subscription is to ensure the connection is successful, resulting in the open event to not fire.

We should remove it from the documentation.

@acinader @davimacedo Thoughts?

dplewis:
@travsr Can you post your quick fix? I'm curious on how you got it to work.

davimacedo:
@dplewis what do you want to remove from the docs? The open event?

dplewis:
@davimacedo Hmm we could add a slight delay [here](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/LiveQueryClient.js#L377). This will allow client time to set an Open Event.

davimacedo:
Yes. Just `setTimeout(() => subscription.emit(SUBSCRIPTION_EMMITER_TYPES.OPEN, response));` should do the magic.

[...]

PR #1151, opened 2020-04-22
Fix: Live Query Subscription Open Event

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/1147

Waits 200 milliseconds to allow for client to set open event after the subscription resolves.


davimacedo:
@dplewis I've just noticed the tests are not passing. Could you please take a look?

PR #1159, opened 2020-05-03
Pass context in destroy, saveAll, get, find hooks.

yog27ray:
This is extension of the https://github.com/parse-community/Parse-SDK-JS/pull/1150. To pass context in following calls.

1. Parse.Object.destroy
2. Parse.Object.saveAll
3. Parse.Query.get
4. Parse.Query.find
5. Parse.Query.first
6. Parse.Query.each
7. Parse.Cloud.run
8. Parse.Object.destroyAll


dplewis:
@yog27ray We can do a Release today to include your changes. Then we can start on the Server Side

yog27ray:
thanks, @dplewis  for merging the PR. I will update the server also.

PR #1179, opened 2020-05-30
Improve eachBatch method

SebC99:
Currently, what's happening with the `eachBatch` Query method is this:
- find X objects
- execute a callback on these objects
- find the next X objects
The issue is that while executing the callback, nothing is done, and we have to wait for the next Find to succeed to start again the callback.

An easy improvement is to launch the next query in parallel of the execution of the callback, so that when the callback is done, the next batch is already ready to be consumed (or more advanced if the query is longer than the callback execution)

It's done by changing the order of the promises, and waiting for both the query and the callback on the previous batch to be resolved.

### DIFF
--- a/src/ParseQuery.js
+++ b/src/ParseQuery.js
@@ -926,18 +926,20 @@ class ParseQuery {
     let previousResults = [];
     return continueWhile(() => {
       return !finished;
-    }, () => {
-      return Promise.all([
+    }, async () => {
+      const [results] = await Promise.all([
         query.find(findOptions),
-        Promise.resolve(callback(previousResults))])
-        .then(([results]) => {
-          if (results.length == 0) {
-            finished = true;
-          } else {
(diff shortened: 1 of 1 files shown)


SebC99:
I don't know why the tests are failing, I guess it has nothing to do with the modifications...

SebC99:
Ok so I had to change the way I tested for the end of `continueWhile` as in the test suite, the `find` mock doesn't take into account the `where` clause on the `objectId`. 
Instead of rewriting all the tests for `eachBatch`, `each`, `map`, `reduce`, and `filter` I've modified the last call of the promise.

SebC99:
(sorry for the force-pushed, I had wrongly pushed another commit with the "built" version I also have)

dplewis:
@SebC99 Sorry for the late reply? Can you rebase your branch?

SebC99:
Done :)

PR #1186, opened 2020-06-16
fix context when cascade-saving objects

mtrezza:
Removed deletion of context from save options to preserve it when cascade-saving child objects.

related to https://github.com/parse-community/parse-server/issues/6736


mtrezza:
@dplewis this is ready for review when you can 🙂

PR #1189, opened 2020-07-04
add idempotency error "duplicate request"

mtrezza:
Adds new Parse Error 159: `DUPLICATE_REQUEST`.

This PR completes the server side implementation of https://github.com/parse-community/parse-server/pull/6748.

### DIFF
--- a/src/ParseError.js
+++ b/src/ParseError.js
@@ -354,6 +354,15 @@ ParseError.FILE_DELETE_ERROR = 153;
 ParseError.REQUEST_LIMIT_EXCEEDED = 155;
 
 /**
+ * Error code indicating that the request was a duplicate and has been discarded due to
+ * idempotency rules.
+ * @property DUPLICATE_REQUEST
+ * @static
+ * @final
+ */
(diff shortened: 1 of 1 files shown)


SebC99:
I love the idea, especially for "long" requests. Not sure how the clients will have to be adapted to silently cope with this error, so we have to mark it as breaking change: 
Before this, there was no error, just duplicates results... and now it will result in an error for all the duplicate calls.

mtrezza:
> I love the idea, especially for "long" requests. Not sure how the clients will have to be adapted to silently cope with this error, so we have to mark it as breaking change: Before this, there was no error, just duplicates results... and now it will result in an error for all the duplicate calls.

@SebC99 This is not a breaking change. This PR merely adds an error code and occupies the currently unused error code `159` for that matter. This is an experimental feature that can be turned on / off. Even when it matures out of experimental status it should not be on by default due to the overhead it incurs, so I assume at this point that this will never be a breaking change.

SebC99:
👌

PR #1209, opened 2020-08-17
Fix User Linking tests

dplewis:
I accidentally deleted the developer account used for testing. Changed the access keys to fix Twitter failing tests.


mtrezza:
I assumed a temporary issue on FB side, glad this spotted this.

mtrezza:
@dplewis I assume we need to create a release for this SDK and point Parse Server to it, to fix the issue of failing tests for Parse Server PRs?

dplewis:
I’ll have a look today. I think it might be a separate issue.

mtrezza:
Ah, right, the Parse Server issue is related to the Facebook domain certificate, sorry for the mix-up. I've opened an [issue](https://github.com/parse-community/parse-server/issues/6867) for that.

PR #1210, opened 2020-08-19
Support Idempotency

dplewis:
Ref: https://github.com/parse-community/parse-server/issues/6744, https://github.com/parse-community/parse-server/pull/6748

Idempotency enforcement for client requests. This deduplicates requests where the client intends to send one request to Parse Server but due to network issues the server receives the request multiple times.
**Caution, this is an experimental feature that may not be appropriate for production.**

To enable use either of the following:
* `Parse.CoreManager.set('IDEMPOTENCY': true)`
* `Parse.idempotency = true`;

This PR addresses
* Add `X-Parse-Request-Id` uuid to the request.
* Improve xmlhttprequest error message handling
* Improve clearing of database after test
* Improve tests the require indexing

### DIFF
--- a/src/RESTController.js
+++ b/src/RESTController.js
@@ -95,7 +95,7 @@ const RESTController = {
       return ajaxIE9(method, url, data, headers, options);
     }
     const promise = resolvingPromise();
-    const isIdempotent = CoreManager.get('IDEMPOTENCY');
+    const isIdempotent = CoreManager.get('IDEMPOTENCY') && ['POST', 'PUT'].includes(method);
     const requestId = isIdempotent ? uuidv4() : '';
     let attempts = 0;
 
@@ -156,7 +156,7 @@ const RESTController = {
         headers['User-Agent'] = 'Parse/' + CoreManager.get('VERSION') +
           ' (NodeJS ' + process.versions.node + ')';
       }
-      if (isIdempotent && ['POST', 'PUT'].includes(method)) {
+      if (isIdempotent) {
         headers['X-Parse-Request-Id'] = requestId;
       }
       if (CoreManager.get('SERVER_AUTH_TYPE') && CoreManager.get('SERVER_AUTH_TOKEN')) {
--- a/src/RESTController.js
+++ b/src/RESTController.js
@@ -95,6 +95,8 @@ const RESTController = {
       return ajaxIE9(method, url, data, headers, options);
     }
     const promise = resolvingPromise();
+    const isIdempotent = CoreManager.get('IDEMPOTENCY');
+    const requestId = isIdempotent ? uuidv4() : '';
     let attempts = 0;
 
     const dispatch = function() {
@@ -154,8 +156,8 @@ const RESTController = {
         headers['User-Agent'] = 'Parse/' + CoreManager.get('VERSION') +
           ' (NodeJS ' + process.versions.node + ')';
       }
-      if (CoreManager.get('IDEMPOTENCY')) {
-        headers['X-Parse-Request-Id'] = uuidv4();
+      if (isIdempotent && ['POST', 'PUT'].includes(method)) {
+        headers['X-Parse-Request-Id'] = requestId;
       }
       if (CoreManager.get('SERVER_AUTH_TYPE') && CoreManager.get('SERVER_AUTH_TOKEN')) {
         headers['Authorization'] = CoreManager.get('SERVER_AUTH_TYPE') + ' ' + CoreManager.get('SERVER_AUTH_TOKEN');
--- a/src/RESTController.js
+++ b/src/RESTController.js
@@ -9,6 +9,8 @@
  * @flow
  */
 /* global XMLHttpRequest, XDomainRequest */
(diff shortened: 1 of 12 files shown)


dplewis:
@mtrezza How does this look? Is this an accurate way to test idempotency?

mtrezza:
Wow, I had this on my list for next month but you overtook me there 👏
Let me take a look at it later today.

src/__tests__/RESTController-test.js
```diff
@@ -432,6 +437,30 @@ describe('RESTController', () => {
...
+  it('sends requestId header for idempotency', async () => {
+    CoreManager.set('IDEMPOTENCY', true);
+    const requestIdHeader = (header) => 'X-Parse-Request-Id' === header[0];
+    const xhr = {
+      setRequestHeader: jest.fn(),
+      open: jest.fn(),
+      send: jest.fn()
+    };
+    RESTController._setXHR(function() { return xhr; });
+    RESTController.request('GET', 'classes/MyObject', {}, {});
```
mtrezza, line 449: Parse Server does not enforce idempotency for GET requests. Idempotency is enforced only on POST, PUT routes as GET, DELETE are already idempotent. However, the test case itself looks good.

src/__tests__/RESTController-test.js
```diff
@@ -432,6 +437,30 @@ describe('RESTController', () => {
...
+    await flushPromises();
+    expect(xhr.setRequestHeader.mock.calls.filter(requestIdHeader)).toEqual(
+      [["X-Parse-Request-Id", '1000']]
+    );
+    xhr.setRequestHeader.mockClear();
+
+    RESTController.request('GET', 'classes/MyObject', {}, {});
```
mtrezza, line 456: Also here, instead of GET we should use a POST or PUT example.

src/RESTController.js
```diff
@@ -152,6 +154,9 @@ const RESTController = {
         headers['User-Agent'] = 'Parse/' + CoreManager.get('VERSION') +
           ' (NodeJS ' + process.versions.node + ')';
       }
+      if (CoreManager.get('IDEMPOTENCY')) {
```
mtrezza, line 157: Since we only send the header for POST, PUT, I think this should be something like `if (CoreManager.get('IDEMPOTENCY') && ['POST', 'PUT'].includes(method)) {`

mtrezza (changes requested):
Looks good, just some minor comments.

Does the JS SDK have an automated request retry mechanism like the iOS and Android SDKs? If so, I think a test case makes sense that ensures that the UUID does not change on request retry. That may then also require additional logic in the REST controller to ensure that the UUID does not change.

mtrezza:
> Does the JS SDK have an automated request retry mechanism like the iOS and Android SDKs? If so, I think a test case makes sense that ensures that the UUID does not change on request retry. That may then also require additional logic in the REST controller to ensure that the UUID does not change.

I've looked into the Parse JS SDK and it seems that it does not have an auto-retry mechanism on failed requests, nor does XHR. For future reference: If such a mechanism ever gets implemented, similar to how it exists on iOS and Android, this test case will likely need to be adapted to properly test that the UUID does not change.

mtrezza (commented):
Can we add a test where we ensure that the request ID header is not sent for GET request?

dplewis:
@mtrezza I've added additional test. There is an auto-retry strategy in the SDK (in RESTController.js) I added a fix for that also. This is ready for review

mtrezza:
Great, I will take a look later today.

dplewis:
@mtrezza @davimacedo How does this look?

mtrezza (commented):
It's already merged but I reviewed it anyway just to be sure and it looks good. Thanks for this PR!

dplewis:
@mtrezza I didn’t want to open a new issue, I’m writing documentation for this. How would you catch a duplicate request error?

```
try {
  await Parse.Cloud.run('duplicate')
  // code here would run for the first request
} catch(e) {
 // duplicate request here
}
```

mtrezza:
Yes, in addition the catch block would check for Parse.Error Duplicate_Request 159, see https://github.com/parse-community/Parse-SDK-JS/pull/1189/files.

dplewis:
Can you think of a way to replicate a duplicate request in a test environment?

```
try {
  const value = await Parse.Cloud.run('duplicate');
  // do something with value
} catch(e) {
  if (e.code === Parse.Error.DUPLICATE_REQUEST) {
   // duplicate request here
  }
  // handle other errors
}
```

mtrezza:
Sure, just send an identical request ID in the header and the request should fail as duplicate request. To send an identical request ID you could spyOn the request ID and fake it to make it the same in every request.

dplewis:
The tests cases are good, I was referring to a real environment. Website, press a button and send a duplicate request.
I just copied the RESTController, set a static requestId and CoreManager.setRESTController.

mtrezza:
>  I was referring to a real environment. Website, press a button and send a duplicate request.

You mean as an example description in the docs or as a real coded test?

PR #1221, opened 2020-09-12
fix(test): Flaky Tests

dplewis:
I don't know the proper way to fix this but I receive this error randomly when I run the test suite. Sometimes it happens once, sometimes 10 times in a row. Has anyone else encountered this?

P.S. I've had this error for a long long time.

Screen Shot 2020-09-12 at 9 57 19 AM


davimacedo (approved):
LGTM

Issue #1196, opened 2020-07-18
Queries fail to parse results when asking for subfield of a null pointer or null array of pointers


henrikperrochon:
### Issue Description

When using select to specify fields and subfields, a query may fail to parse results if one of the field is null.

### Steps to reproduce

1 - From the dashboard, select a collection (let say Sample) with a field typed as array (let say arrayfield).
2 - Set the field to be null.
3 - Call the following query :
`new Parse.Query('Sample').select(['arrayfield.subfield']).find().then(console.log, console.error);`

### Expected Results

An error occurs: 
`Uncaught (in promise)
code: undefined
text: "Cannot read property 'subfield' of null"
`

### Actual Outcome

The networking tools from Chrome shows that the server manage to answer, but the Parse JS SDK fails to parse the results. The following method fails when reading the null array. ` if (_obj !== undefined) {` does not prevent null objects.
      
      function handleSelectResult(data, select) {
        var serverDataMask = {};
        (0, _forEach.default)(select).call(select, function (field) {
          var hasSubObjectSelect = (0, _indexOf.default)(field).call(field, ".") !== -1;
      
          if (!hasSubObjectSelect && !data.hasOwnProperty(field)) {
            // this field was selected, but is missing from the retrieved data
            data[field] = undefined;
          } else if (hasSubObjectSelect) {
            // this field references a sub-object,
            // so we need to walk down the path components
            var pathComponents = field.split(".");
            var _obj = data;
            var serverMask = serverDataMask;
            (0, _forEach.default)(pathComponents).call(pathComponents, function (component, index, arr) {
              // add keys if the expected data is missing
              if (_obj && !_obj.hasOwnProperty(component)) {
                _obj[component] = undefined;
              }
      
              if (_obj !== undefined) { <----------------------------------------------- the issue is triggered after this test
                _obj = _obj[component];
              } //add this path component to the server mask so we can fill it in later if needed
      
      
              if (index < arr.length - 1) {
                if (!serverMask[component]) {
                  serverMask[component] = {};
                }
      
                serverMask = serverMask[component];
              }
            });
          }
        });
      
        if ((0, _keys2.default)(serverDataMask).length > 0) {
          // When selecting from sub-objects, we don't want to blow away the missing
          // information that we may have retrieved before. We've already added any
          // missing selected keys to sub-objects, but we still need to add in the
          // data for any previously retrieved sub-objects that were not selected.
[...]

PR #1223, opened 2020-09-12
fix(ParseQuery): Select on null fields

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/1196

### DIFF
--- a/src/ParseQuery.js
+++ b/src/ParseQuery.js
@@ -97,7 +97,7 @@ function handleSelectResult(data: any, select: Array<string>){
         if (obj && !obj.hasOwnProperty(component)) {
           obj[component] = undefined;
         }
-        if (obj != undefined) {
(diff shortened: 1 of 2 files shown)


src/ParseQuery.js
```diff
@@ -97,7 +97,7 @@ function handleSelectResult(data: any, select: Array<string>){
         if (obj && !obj.hasOwnProperty(component)) {
           obj[component] = undefined;
         }
-        if (obj !== undefined) {
+        if (obj != undefined) {
```
davimacedo, line 100: Maybe `obj && typeof obj === 'object'`?

davimacedo (commented):
It is good for me. Just sending a suggestion.

src/ParseQuery.js
dplewis: That looks right. I'll try it.

PR #1226, opened 2020-09-15
fix(docs): Matches Modifier Documentation

mirzams2000:
Add missing modifier docs for matches query.


dplewis (approved):
Thanks for the contribution!

PR #1229, opened 2020-09-22
allow optional HTTP POST method override for logIn user method

kolohelios:


src/ParseUser.js
```diff
@@ -439,7 +439,7 @@ class ParseUser extends ParseObject {
    * @return {Promise} A promise that is fulfilled with the user when
    *     the login is complete.
    */
-  logIn(options?: FullOptions): Promise<ParseUser> {
+  logIn(options?: FullOptions, usePost?: boolean): Promise<ParseUser> {
```
dplewis, line 442: Can you move `usePost` to options? Also add it to documentation.

src/__tests__/ParseUser-test.js
```diff
@@ -229,6 +229,39 @@ describe('ParseUser', () => {
     });
   });
 
+  it('can log in as a user with POST method', (done) => {
```
dplewis, line 232: Can you add an integration test?

dplewis (changes requested):
Thanks for the PR! I was just thinking about this. I don't this will work without server update.

I added a fix last week [here](https://github.com/parse-community/parse-server/pull/6903). If you write an integration test it might fail.

src/ParseUser.js
kolohelios: Done. Thank you so much for the quick review!

src/__tests__/ParseUser-test.js
kolohelios: I did, and it passed. I believe that the server is reading the credentials from the body as the short-circuit for query params is falsy though I could be wrong about that.

dplewis (approved):
LGTM!

PR #1233, opened 2020-10-18
Add a findAll method to ParseQuery

codeKonami:
To get around the 100 objects maximum retrieval on the Parse Server side I used to do `Parse.Query().limit(9999).find()`.

And after looking at the code I found out about the `eachBatch` method which is really neat and I wonder if you'd consider adding a `findAll` method the the Parse.Query class based on it.

### DIFF
--- a/integration/test/ParseQueryTest.js
+++ b/integration/test/ParseQueryTest.js
@@ -1223,6 +1223,18 @@ describe('Parse Query', () => {
     });
   });
 
+  it('can return all objects with findAll', async () => {
+    const objs = [...Array(101)].map(() => new Parse.Object('Container'));
+
+    await Parse.Object.saveAll(objs);
+
+    const query = new Parse.Query('Container');
+
+    const result = await query.findAll();
+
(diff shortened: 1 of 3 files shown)


davimacedo (approved):
It looks good to me! @dplewis any thoughts?

integration/test/ParseQueryTest.js
```diff
@@ -1223,6 +1223,18 @@ describe('Parse Query', () => {
     });
   });
 
+  it('can return all objects with findAll', async () => {
```
dplewis, line 1226: Looks good! Can you add unit tests to increase coverage?

codeKonami: I've added some unit tests. I basically took the ones from .eachBatch but using the findAll method. I hope it's enough.

dplewis (approved):

integration/test/ParseQueryTest.js
dplewis: Looks good!

Issue #491, opened 2017-10-18
Setting Parse.serverURL not working


msebgf:
Hi, I created a react native project.

My dependencies are:

`"dependencies": {
    "events": "^1.1.1",
    "history": "^4.7.2",
    "parse": "^1.10.0",
    "react": "16.0.0",
    "react-native": "^0.48.4",
    "react-native-splash-screen": "^3.0.1",
    "react-redux": "^5.0.6",
    "react-router-native": "^4.2.0",
    "react-router-redux": "^5.0.0-alpha.6",
    "redux": "^3.7.2",
    "redux-persist": "^4.10.1",
    "redux-thunk": "^2.2.0",
    "timers": "^0.1.1"
  }`

So basically when I tried to configure the Parse connections I did it on App.js on componentWillMount(), like so:

`componentWillMount(){
   Parse.initialize("AppId","JSKey");
   Parse.serverURL = 'http://parseServer:1337/parse' 
}
`

Everything works fine until Parse.intialize, but when trying to set the serverURL I get this error:

`https://api.parse.com/1 is not a function (evaluating '_reactNative3.default.serverURL('http://parseServer:1337/parse')')
componentWillMount`

PR #1237, opened 2020-10-22
Improve initialization setters/getters

dplewis:
Addresses: https://github.com/parse-community/Parse-SDK-JS/issues/491

I've randomly seen users post about not being able to set the `serverURL`. It's usually typescript / web pack / babel issue with `defineProperty`.

https://parseopensource.slack.com/archives/C011ZF86LR2/p1600690460002100

Thoughts?


davimacedo (approved):
I've never seen this problem but I believe the new way you are creating the accessor methods will be easier for typescript, etc "understand". LGTM!

PR #1239, opened 2020-10-23
Remove backbone callback documentation

dplewis:
Old backbone callbacks (options.success, options.error) are only used in linkWith authentication and object.set validation now.

i.e
```
object.set("currentPlayer", player2, {
  error: function(objectAgain, error) {
    // The set failed validation.
  }
});
```


mtrezza (approved):
lgtm 🚀

PR #1241, opened 2020-10-29
Improve Code Coverage

dplewis:
Giving the SDK some much needed love

Fixes a few bugs

* Set default emitter error on LiveQueryClient to prevent crashing
* Query subscriptions will return error on socket error
* WeChat socket sets handlers first before connecting
* Fix Parse Installation error message
* ParseACL remove unreachable code


davimacedo (approved):
Very good job @dplewis LGTM!

PR #1247, opened 2020-11-09
fix(fetchAll): Properly handle passed in values

dplewis:
Prevents `TypeError: Cannot read property 'id' of null`

Handles input the same way saveAll and destroyAll.

Also removed unnecessary LDS call on destroy all for non ParseObject / Array<ParseObject>

Let me know if I need to change anything.


davimacedo (approved):
LGTM!

PR #1255, opened 2020-11-12
fix: Internal Referencing for Increment Dot Notation

dplewis:
This PR is building upon https://github.com/parse-community/Parse-SDK-JS/pull/1219, together they solve the following issues.

There was a internal referencing issue that resulted in the increment to happen every time you get the attributes
Lucky this `.get` returns the proper value so clients aren't affected. (Fixed in this PR).
```
const obj = new TestObject();
obj.set('objectField', { number: 5 });
console.log(obj.attributes); // Internally number is 20
console.log(obj.attributes); // Internally number is 35
console.log(obj.attributes); // Internally number is 50
```

When you save with increment there is an extra attribute returned from _handleSaveResponse (Fixed in #1219)
```
{
  objectField: { number: 20 },
  createdAt: 2020-11-11T17:35:48.317Z,
  updatedAt: 2020-11-11T17:35:48.337Z,
  'objectField.number': 15 // Has been removed
}
```

When saving after using increment nested, the full object isn't returned. (Fixed in #1219)
```
{
  objectField: { number: 20, letter: 'a' }, // Didn't return 'a' before when using increment nested
  createdAt: 2020-11-11T17:35:48.317Z,
  updatedAt: 2020-11-11T17:35:48.337Z,
}
```

* Also adds error listener to test server. Useful for catching 'Address in Use` errors.

When merged with the server this will work in beforeSave / afterSaves hooks.

### DIFF
--- a/integration/test/ParseObjectTest.js
+++ b/integration/test/ParseObjectTest.js
@@ -12,6 +12,7 @@ describe('Parse Object', () => {
   beforeEach((done) => {
     Parse.initialize('integration', null, 'notsosecret');
     Parse.CoreManager.set('SERVER_URL', 'http://localhost:1337/parse');
+    Parse.CoreManager.set('REQUEST_ATTEMPT_LIMIT', 1);
     Parse.Storage._clear();
     clear().then(() => {
       done();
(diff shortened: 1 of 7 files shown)


davimacedo (commented):
It looks good but I believe you should revert the changes in the package-lock, right?

PR #1257, opened 2020-11-18
Change file destroy error message

mtrezza:
Improving the error message when deleting an unnamed file.

The current error message mentions an "unsaved" file, while checking for a file name.
To be more specific, the error message should mention an "unnamed" file.

As discussed in https://github.com/parse-community/docs/pull/791#discussion_r526480463.


davimacedo (approved):

davimacedo:
There is a failing test you also need to change the message in the test case.

dplewis:
@mtrezza can you fix the test and update from master for the CI?

mtrezza:
Fails and I cannot see any output, is that a known issue?

image

davimacedo:
Could you check again? I can now see. The Github Actions logs are little bit buggy. Sometimes I had to refresh the page to see the data.

dplewis:
@mtrezza you can run the test locally and find the failing test.

mtrezza:
Thanks, I can see the log output in the CI here now too.

The error was that the error message was different in the test vs. in the method. That confirms that it could be beneficial to [consolidate the Parse Error codes and messages into their own repo](https://github.com/parse-community/parse-server/pull/6977#issuecomment-738635737) at some point in the future.

Back to the test here, it throws just an `Error`, there is no dedicated `ParseError` for this:
https://github.com/parse-community/Parse-SDK-JS/blob/f50b94ed4a42d76bb4f87bd4b2f2e721b0e470f8/src/ParseFile.js#L323

On the other hand in the [docs](https://docs.parseplatform.org/js/guide/#file-related-errors) we describe a ParseError that is not defined in the SDK and the error code and message are not referenced anywhere in Parse JS SDK or Parse Server:

`FileDeleteError | 131 | File could not be deleted.`

What do you think about adding the error code definition to the JS SDK and throwing this Parse Error instead of the current error?

dplewis:
I would prefer some consistency. Now that you mention there are a few TypeError being thrown instead of Parse.Error.

mtrezza:
> I would prefer some consistency. Now that you mention there are a few TypeError being thrown instead of Parse.Error.

By consistency you mean we should throw a proper Parse.Error and use the `131` error?

mtrezza:
I found the missing Parse Error with error number 153 instead of 131 as documented:

https://github.com/parse-community/Parse-SDK-JS/blob/f24bd991514414f7a3e8ac13a30f1548e2052e0b/src/ParseError.js#L338-L344

And it's thrown in Parse Sever when deleting a file fails for an unspecified reason, as in this test case:

https://github.com/parse-community/Parse-SDK-JS/blob/7cf413eed95d7096a3b8f0106d74a7e59d4d04d9/integration/test/ParseFileTest.js#L107-L117

Since the Parse Error codes don't follow any number ranges, we can just correct the docs by moving the error to the [file related errors](https://docs.parseplatform.org/js/guide/#file-related-errors).

And since the error is thrown for unspecified errors, it makes sense to introduce a new Parse Error for this PR.

mtrezza:
Incorrect error number fixed in https://github.com/parse-community/docs/issues/803

mtrezza:
> I would prefer some consistency. Now that you mention there are a few TypeError being thrown instead of Parse.Error.

I have opened https://github.com/parse-community/Parse-SDK-JS/issues/1269 for this.

mtrezza:
Ready for review, added new Parse Error for when trying to delete an un-named file.

davimacedo (approved):
LGTM!

dplewis (approved):

PR #1267, opened 2020-12-03
Github Actions

dplewis:


dplewis:
@davimacedo How does this look?

.github/workflows/pages.yml
```diff
@@ -0,0 +1,47 @@
+name: docs
+on:
```
davimacedo, line 2: I am not sure because it is new for me, but I'd try with:
on:
  release:
    types: [published]

davimacedo (commented):
@dplewis great job! Just left a comment.

.github/workflows/pages.yml
dplewis: I can change it when we do a release. I know it works but still want to test it on push

davimacedo: Makes sense

davimacedo (approved):
LGTM!

davimacedo:
I will try to remove Travis from being required

dplewis:
Thanks, I didn’t know the best way to remove it

davimacedo:
Found it. We should be good now.

davimacedo:
For reference: Settings -> Branches -> Branches Protection Rules > Edit Master

There you will find a list of checkboxes with all integrations for you to check which ones are required.

dplewis:
Thanks, the docs work

https://github.com/parse-community/Parse-SDK-JS/tree/gh-pages

Issue #1277, opened 2020-12-11
Do not support fieldName 'length' on Mongo 4.X


Moumouls:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-JS/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse JS SDK](https://github.com/parse-community/Parse-SDK-JS/releases). 

### Issue Description

Issue name

### Steps to reproduce

Run Parse SDK JS integration test with a mongo version 4.X
### Actual Outcome
Cannot save object with a field "length"

### Expected Outcome

Can save the object

###  Environment

Server
- Parse Server version: `master`
- Operating system: `macOS`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `local`

Database
- System (MongoDB or Postgres): `Mongo`
- Database version: `4.X`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `local`

Client
- Parse JS SDK version: `master`

### Logs

```
Failures:
1) Parse LocalDatastore Parse Query Pinning (Default) Default supports objects with length
  Message:
    Error: Invalid field name: length.
  Stack:
    error properties: Object({ code: 105 })
    Error: Invalid field name: length.
        at handleError (/Users/antoinecormouls/Desktop/OpenSource/Parse-SDK-JS/lib/node/RESTController.js:422:17)
        at <Jasmine>
        at processTicksAndRejections (internal/process/task_queues.js:97:5)

2) Parse LocalDatastore Parse Query Pinning (React-Native) React-Native supports objects with length
  Message:
    Error: Invalid field name: length.
  Stack:
    error properties: Object({ code: 105 })
    Error: Invalid field name: length.
        at handleError (/Users/antoinecormouls/Desktop/OpenSource/Parse-SDK-JS/lib/node/RESTController.js:422:17)
        at <Jasmine>
        at processTicksAndRejections (internal/process/task_queues.js:97:5)

3) Parse Query supports objects with length
  Message:
[...]

PR #1278, opened 2020-12-11
Update tests

dplewis:
* Addresses https://github.com/parse-community/Parse-SDK-JS/issues/1277
* Fix flaky Tests

Bump parse-server to [#43d9af8
](https://github.com/parse-community/parse-server/commit/43d9af847bdd66c2cc09a2f293c34534a7c042bd)

### DIFF
--- a/integration/test/ParseLocalDatastoreTest.js
+++ b/integration/test/ParseLocalDatastoreTest.js
@@ -2375,22 +2375,6 @@ function runTest(controller) {
       assert(['Billy', 'Tom'].includes(results[1].get('name')));
     });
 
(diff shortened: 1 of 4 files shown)


davimacedo (commented):
Why are we removing the tests with length?

davimacedo:
Nevermind. I've just seen the other related PRs.

davimacedo (approved):

mtrezza:
LGTM, although I am starting to wonder whether we have the best approach when it comes to reserved field names. Why would a developer have to care about internally used field names. Maybe we should just encode the field names in a different way, so that they are separate from any JS property names. But that's a whole other PR.

PR #1279, opened 2020-12-13
Fix prettier

dplewis:
Adding single quotes around the glob allows it to find subfolders in directories

### DIFF
--- a/src/__tests__/test_helpers/asyncHelper.js
+++ b/src/__tests__/test_helpers/asyncHelper.js
@@ -10,17 +10,17 @@
 // We need this until Jest finishes upgrading to Jasmine 2.0
 module.exports = function asyncHelper(fn) {
   let finished = false;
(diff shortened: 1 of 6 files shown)


src/__tests__/test_helpers/asyncHelper.js
```diff
@@ -10,17 +10,17 @@
 // We need this until Jest finishes upgrading to Jasmine 2.0
 module.exports = function asyncHelper(fn) {
   let finished = false;
-  const done = function() {
+  const done = function () {
```
mtrezza, line 13: I think we had that Lint discussion already, but the space between `function` and brackets is what we want?

dplewis: Yeah, technically you need a space before the name of the function. The name here is empty so the space should remain.

PR #1282, opened 2021-01-07
fix(LiveQuery): Allow connect with null fields

dplewis:
Fixes issue where optional field if null won't connect to the LiveQueryServer.

Fixes: https://github.com/parse-community/parse-server/pull/7099

See [LiveQuery Protocol Specification](https://github.com/parse-community/parse-server/wiki/Parse-LiveQuery-Protocol-Specification)


davimacedo (approved):
LGTM!

PR #1288, opened 2021-01-28
docs: update equals method in ParseACL

JeffGuKang:
Original description didn't have type expression for a argument. 

before

```
@param other The other object to compare to
```

after

```
@param {ParseACL} other The other object's ACL to compare to
```


dplewis (approved):
LGTM!

Issue #1230, opened 2020-09-24
Feature saveEventually for offline use


KAIOAVILA:
I have difficulties in implementing an offline-first application in react native, due to the lack of a very essential functionality to work well, which in the sdk of Android and iOS has to be saveEventually.  Because if not, I have to use another database to keep track of it.  Firebase has this functionality implemented in their lib.

implement something like what you have in the Android and iOS sdk.  I only talked about save, but there is also delete.  use some solution to store in storage to store these objects and when you resume the network synchronize with Parse.  https://docs.parseplatform.org/android/guide/#saving-objects-offline

dplewis:
I tried working on something similar after I added the LDS. You can view the conversation [here](https://github.com/parse-community/Parse-SDK-JS/pull/734#issuecomment-463011648).

I'd be more that happy to discuss a strategy with you based off the existing implementation in iOS and Android.

meepeek:
Seems there was an attempt for JS SDK saveEventually on Titanium. It may be useful for the implementation. Link as follow

https://gist.github.com/cr0ybot/001c5aafcfd35251304961670287ee21

I'm not very deep into the code. Hope this help. Also react and react native webapp quite popular nowadays, I wish this would have been implemented.

dplewis:
@meepeek Thanks that’s really helpful. The only problem I see it how to tell if the network connection is offline / online. I would like to avoid using the RestController since its only goal should be to send requests.

meepeek:
@dplewis 
should we make it part of livequery ? this way the client must connect ws then we can check on ws status.

dplewis:
Not user had a Live Query server set up. But now that you mention it. We could add an emitter to the RestController to check if there is connection / deconnection

PR #1291, opened 2021-02-03
Add EventuallyQueue API

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/1230

This is a similar feature to the [iOS SDK EventuallyQueue](https://github.com/parse-community/Parse-SDK-iOS-OSX/blob/master/Parse/Parse/Internal/PFEventuallyQueue.m) to allow a user to cache objects to be saved later. Inspired by @francimedia [repo](https://github.com/francimedia/parse-js-local-storage).

This has a lot of room for improvement, any feedback is appreciated.

- [ ] Save internal unique hash on object to the DB to ensure operations won't be duplicated.
- [ ] If the operation happens at a later time update object internal state
- [ ] Change Polling setInterval to setTimeout?

### Parse.Object

Saves / Deletes this object to the server at some unspecified time in the future, even if Parse is currently inaccessible.

 Use this when you may not have a solid network connection, and don't need to know when the save completes.
 If there is some problem with the object such that it can't be saved, it will be silently discarded.

 Objects saved / deleted with this method will be stored locally in an on-disk cache until they can be delivered to Parse.
 They will be sent immediately if possible. Otherwise, they will be sent the next time a network connection is
 available. Objects saved this way will persist even after the app is closed, in which case they will be sent the
 next time the app is opened / SDK is initialized.

```
await object.saveEventually(options);
await object.destroyEventually(options);
```

#### Add Object to Queue
Adding unique hashes to your objects is highly recommended

```javascript
const object = new Parse.Object('TestObject');
object.save({ hash: 'unique' });

await Parse.EventuallyQueue.save(object, options);
await Parse.EventuallyQueue.destroy(object, options);
```

#### Send Queue to Parse Server
```javascript
await Parse.EventuallyQueue.sendQueue();
```
You can send the queue either right after adding an object or on connect (or other events).

#### Clear Queue
```javascript
await Parse.EventuallyQueue.clear();
```

#### Polling
This will call sendQueue when a connection to the server is established. Uses the `serverURL` used to initialize sdk
```javascript
Parse.EventuallyQueue.poll();
```


src/EventuallyQueue.js
```diff
@@ -0,0 +1,188 @@
...
+        if (e.message !== 'XMLHttpRequest failed: "Unable to connect to the Parse API"') {
+          await this.remove(queueObject.queueId);
+        }
+      }
+      break;
+    }
+  },
+
+  poll() {
+    if (this.polling) {
```
mtrezza, line 145: Could this cause race conditions issues?

dplewis: Good question. I'll have to think on it.

dplewis: By raise condition do you mean that multiple polling could happen at the same time?

mtrezza: Yes, as in calling `poll` multiple times from different threads, then `this.polling` may not be set thread safely.

dplewis: I was thinking about changing the setInterval to setTimeout implementation.

mtrezza: I think that makes sense since we're looking at a network op. For example, if the network op times out only after 60 seconds (or more), and the setInterval also executes every 60 seconds (or less), there may be an overlap. That issue wouldn't exists using a nested setTimeout.

mtrezza: I think that makes sense since we're looking at a network op. For example, if the network op times out only after 60 seconds, and the setInterval also executes every 61 seconds, then the time between network retries would be only 1 second. And maybe in the previous 60 seconds, the host system also has some retry mechanism before it times out, so it would be constantly trying to reach the server. That issue wouldn't exists using a nested setTimeout.

Technically, polling could hardly ever overlap in JS (except for edge cases with multi-worker maybe), so the if clause there should be fine whether setTimeout or setInterval is used, both are based on the same timer mechanism.

dplewis: I fixed the raise condition issue, just to be sure.

mtrezza:
Looks good, should this be flagged as experimental in the changelog like we do with Parse Server sometimes?

dplewis:
Since we can’t turn it off it’s not experimental.

integration/test/ParseEventuallyQueueTest.js
```diff
@@ -0,0 +1,238 @@
...
+    const results = await query.find();
+    assert.strictEqual(results.length, 3);
+    assert.strictEqual(results[0].get('foo'), 'bar');
+    assert.strictEqual(results[1].get('foo'), 'baz');
+    assert.strictEqual(results[2].get('foo'), 'bag');
+
+    length = await Parse.EventuallyQueue.length();
+    assert.strictEqual(length, 0);
+
+    // TODO: Properly handle SingleInstance
```
dplewis, line 66: I’m using JSON.stringify to save the object in local storage. When I retrieve it’s no longer a class instance or same reference as the stored object. Is their a way to fix this? In a singleInstance environment like the browser. Objects with similar id are the same. I can’t pass obj1 into destroy since it didn’t get an objectId from save.

src/EventuallyQueue.js
```diff
@@ -0,0 +1,373 @@
...
+    action: string,
+    object: ParseObject,
+    serverOptions: SaveOptions | RequestOptions
+  ): Promise {
+    const queueData = await this.getQueue();
+    const queueId = this.generateQueueId(action, object);
+
+    let index = this.queueItemExists(queueData, queueId);
+    if (index > -1) {
+      // Add cached values to new object if they don't exist
```
dplewis, line 107: If you try to add an object to the queue that already exists what should happen? Should it override the existing object?

dplewis (commented):
I have a few questions that I would like feedback on. @mtrezza @davimacedo

integration/test/ParseEventuallyQueueTest.js
mtrezza: Not sure I understood this correctly, but `Object.create` takes  the prototype to recreate the instance. Depending on what exactly goes missing, `JSON.stringify` only stringifies enumerable, own properties, to sometime you may want to write a custom method to serialize/encode and then deserialize/decode an object.

dplewis: Stringify calls toJSON() internally. I ran into this issue with the LocalDatastore as well. I don’t really have time to figure this out. Might just post a bounty or something.

src/EventuallyQueue.js
mtrezza: Not sure about overriding. Couldn't that contradict the queue's FIFO? Unless the two operations to be merged are also adjacent to each other, then that operation could be optimized. However, this would be a custom flavor because a separate create and update operation are not necessarily the same as a merged create op.

dplewis: It sends to the server as a queue. It currently merges for optimization. I’m not sure if the way I’m merging is correct. I do have a test for this.

dplewis: https://github.com/parse-community/Parse-SDK-JS/pull/1291/files#diff-75a47fd2940c491cf43ac004fdfa07d1324e16064dab5a241132941fa949ef5cR141

What should the output be?

integration/test/ParseEventuallyQueueTest.js
```diff
@@ -0,0 +1,238 @@
...
+    await Parse.EventuallyQueue.save(object);
+
+    object.set('foo', undefined);
+    await Parse.EventuallyQueue.save(object);
+
+    const length = await Parse.EventuallyQueue.length();
+    assert.strictEqual(length, 1);
+
+    const queue = await Parse.EventuallyQueue.getQueue();
+    assert.strictEqual(queue[0].object.foo, 'bar');
```
mtrezza, line 154: This should actually be `undefined`, according to this test:

```js
const o = new Parse.Object('Tmp');
o.set('key', 'override1');
o.set('key', undefined);
await o.save(null, {useMasterKey: true});
```
Creates an object with `key: undefined`.

src/EventuallyQueue.js
mtrezza: Sorry for this criss-cross comment... https://github.com/parse-community/Parse-SDK-JS/pull/1291#discussion_r571507883

integration/test/ParseEventuallyQueueTest.js
dplewis: Ah makes sense

mtrezza: Another optimization could be save-delete:
1. Add save new obj operation to queue.
2. Add delete new object operation to queue.
3. Optimize: delete overrides save and therefore removes both entries from queue.

Although, still I am not sure these optimizations would be intuitive or wanted by everyone? Saving an object can trigger hooks in cloud code, deleting an object as well. So someone may want to save, then delete an object and expect triggers to be executed in between. If the two ops are optimized in the queue, they never reach the server. Therefore these optimizations maybe should be optional? Not sure.

mtrezza: Maybe also a test case that ensures that no optimization takes places if two operations are not adjacent to each other?

mtrezza: @dplewis I think your test case is right after all:

a)
```js
const o = new Parse.Object("Temp");
o.set('key', 'override');
await o.save(null, {useMasterKey: true});
o.set('key', undefined);
await o.save(null, {useMasterKey: true}); // results in object with `key: override`
```

b)
```js
const o = new Parse.Object("Temp");
o.set('key', 'override');
o.set('key', undefined);
await o.save(null, {useMasterKey: true}); // results in object with `key: undefined`
```

It makes sense considering that `undefined` is currently interpreted as "no value set" and therefore ignored when saving an object.

dplewis: Thanks for looking into this. I think this is ready to merge for now

mtrezza (approved):
Maybe add a chapter about this in the README so people become aware and try this this out? Otherwise LGTM!

PR #1292, opened 2021-02-03
Improve integration testing

dplewis:
Allows for users to change the testing server configurations during a test suite. Also allows for use to access current ParseServer instance. I referenced the [helper.js](https://github.com/parse-community/parse-server/blob/master/spec/helper.js#L122) from the Server repo.

* Ensure only one beforeAll exists (in helper.js)
* Add missing awaits to ParseACLTest.js and MasterKeyTest.js
* Rename tests with duplicate names ArrayOperationTest.js
* Fix flaky test in LiveQueryClientTest
* Add reconfigureServer test cases
* Add `use strict` to prevent global side effects

I'll run the CI a few times to make sure everything runs properly.

Needed for https://github.com/parse-community/Parse-SDK-JS/pull/1291


dplewis:
@mtrezza I turn on `{ random: true }` in `jasmine.json`. I did this to find tests that depend on other tests, poorly written tests or test that don't teardown / clean up when they are done.

mtrezza:
That is a good idea, we could also try to do that for Parse Server, if it doesn't go crazy.

mtrezza (approved):
LGTM! 🚀

PR #1293, opened 2021-02-10
fromJSON: Return date if value is a date 

dblythy:
Not sure how often this error would occur, but if you try to create a `ParseObject` with `fromJSON` with a Date, the date field is returned as `{}`

[Related](https://github.com/parse-community/parse-server/issues/7082)

Evidently, this bug also affects the localDataStore tests as:

descending date order should be larger (newer) to smaller (older)

Obj1: createdAt: new Date('2018-08-10T00:00:00.000Z'),
Obj2: createdAt: new Date('2018-08-11T00:00:00.000Z'),
Obj3: createdAt: new Date('2018-08-12T00:00:00.000Z'),

So, this test:

```
    q = new ParseQuery('Item');
    q.descending('_created_at');
    q.fromLocalDatastore();
    results = await q.find();
```

Should return Obj3, Obj2, Obj1. Instead, it returns Obj1, Obj2, then Obj3.

If this doesn't make sense, if you change the order that the objects are loaded [here](https://github.com/parse-community/Parse-SDK-JS/blob/d40f1505a29ae0080d6094f5557f4f372f14a428/src/__tests__/ParseQuery-test.js#L3254) (e.g to Obj2, Obj3, Obj1), the tests will fail on the current master.

### DIFF
--- a/src/__tests__/ParseObject-test.js
+++ b/src/__tests__/ParseObject-test.js
@@ -1243,7 +1243,6 @@ describe('ParseObject', () => {
     expect(o.op('count')).toBe(undefined);
   });
 
-
   it('can revert a specific field in unsaved ops', () => {
     const o = ParseObject.fromJSON({
       className: 'Item',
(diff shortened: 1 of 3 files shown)


dplewis:
Looks good, but the coverage decreased.

dblythy:
Should be good now @dplewis

dplewis (approved):

PR #1294, opened 2021-02-11
feat: Add option to return raw json from queries

dplewis:
Mimics [options](https://github.com/parse-community/parse-php-sdk/pull/423) in PHP SDK.

Works for get, find and first queries.


davimacedo (approved):
LGTM!

Issue #482, opened 2017-09-19
Parse.Object.fromJSON creates empty record in DB


andrekosak:
SDK package version: **1.10.0**

When I use Parse.Object.fromJSON on NodeJS server, record are being created containing only `objectId`, `createdAt`, `updatedAt`. All other properties i have passed in my JSON are being ignored.

In the example below I am trying to create an object with properties `a` and `b`, but they are not being passed to server. At first I create an object and then getting this object by its ID to see if it has my properties.

Here is my code:

```
const testObject = {
    a: 1,
    b: 2,
    className: 'TestOnly1'
}
const test = Parse.Object.fromJSON(testObject);
test.save()
    .then((newObject) => {
        console.log(JSON.stringify(newObject, null, 3));

        const query = new Parse.Query('TestOnly1')
        query.get(newObject.id)
            .then((queryResponse) => {
                console.log(JSON.stringify(queryResponse, null, 3));
            })
    });
```
console output from first console.log():

```
{
   "a": 1,
   "b": 2,
   "createdAt": "2017-09-19T09:48:58.174Z",
   "updatedAt": "2017-09-19T09:48:58.174Z",
   "objectId": "fqJV15MOLK"
}
```

output from second console.log():

```
{
   "createdAt": "2017-09-19T09:48:58.174Z",
   "updatedAt": "2017-09-19T09:48:58.174Z",
   "objectId": "fqJV15MOLK"
}
```

Is it a bug or i am doing something wrong?

dplewis:
I saw this as well. Look like `fromJSON` wasn't designed to save data.

* `save` only works on dirty fields i.e `obj.set()` that have yet to be committed and not in the cache.
* `fromJSON` saves fields in the cache.
* Your example ends up with no dirty fields to be saved since all the data is cache.
* `save.then(newObject)` just returning what was in the cache

@flovilmart Is this by design or for offline use?

flovilmart:
I’m not sure, i’ll Have to investigate more, starting with when it was added :)

francojohnc:
I check the code and no set called from parse object


```
static fromJSON(json
  /*: any*/
  , override
  /*:: ?: boolean*/
  ) {
    if (!json.className) {
      throw new Error('Cannot create an object without a className');
    }

    const constructor = classMap[json.className];
    const o = constructor ? new constructor() : new ParseObject(json.className);
    const otherAttributes = {};

    for (const attr in json) {
[...]

PR #1295, opened 2021-02-11
Allow fromJSON to set keys to dirty

### LINKED ISSUE #161: unable to save Object.fromJSON() because it's not dirty
Hello!

I'm using Parse library within angular.js framework.
To bind parse object to Angular model, I'm using Parse.Object.toJSON() method.
Then, I need to update data on parse. To do this, I create new parse object using method Parse.Object.fromJSON({json}) and then call save() method.
I found that it sends to server only _updateAt field.
If I manually call "set" method on this object fields, parse starts working as expected.
I think, there is an issue with "dirty" fields.

dblythy:
Currently, there is no way to save fields to a Parse.Object using `fromJSON`, as `dirtyKeys` doesn't get set.

This can be confusing as at first glance, you would expect the keys to be set as they are defined. However, after the object is fetched, the keys will be null.

```
it('from json save data', async (done) => {
    const json = {
      className : 'TestObject',
      date: new Date(),
      array: [],
      object: {},
      string: ''
    }
    const obj = Parse.Object.fromJSON(json);
    expect(obj.get('date')).toBeDefined()
    expect(obj.get('date')).toBeInstanceOf(Date);
    expect(obj.get('array')).toBeDefined()
    expect(obj.get('array')).toBeInstanceOf(Array);
    expect(obj.get('object')).toBeDefined()
    expect(obj.get('object')).toBeInstanceOf(Object);
    expect(obj.get('string')).toBeDefined()
    expect(obj.get('string')).toBeInstanceOf(String);
    await obj.save();
    // all keys are defined


    await obj.fetch();
    // only created at and updated at are defined
    expect(obj.get('date')).toBeDefined()
    expect(obj.get('date')).toBeInstanceOf(Date);
    expect(obj.get('array')).toBeDefined()
    expect(obj.get('array')).toBeInstanceOf(Array);
    expect(obj.get('object')).toBeDefined()
    expect(obj.get('object')).toBeInstanceOf(Object);
    expect(obj.get('string')).toBeDefined()
    expect(obj.get('string')).toBeInstanceOf(String);
    done();
  });
```

This PR allows a 3rd parameter to `fromJSON`, where you can specify whether the keys should be set to dirty.

```js
    const json = {
      className : 'TestObject',
      date: new Date(),
      array: [],
      object: {},
      string: ''
    }
    const obj = Parse.Object.fromJSON(json, false, true);
    // now keys will be dirty
```

Properly closes #161 and #482

### DIFF
--- a/src/__tests__/ParseObject-test.js
+++ b/src/__tests__/ParseObject-test.js
@@ -1243,7 +1243,6 @@ describe('ParseObject', () => {
     expect(o.op('count')).toBe(undefined);
   });
 
-
   it('can revert a specific field in unsaved ops', () => {
     const o = ParseObject.fromJSON({
       className: 'Item',
--- a/src/__tests__/ParseObject-test.js
+++ b/src/__tests__/ParseObject-test.js
@@ -255,11 +255,13 @@ describe('ParseObject', () => {
   });
 
   it('can be inflated from server JSON', () => {
+    const date = new Date();
     const json = {
       className: 'Item',
       createdAt: '2013-12-14T04:51:19Z',
       objectId: 'I1',
       size: 'medium',
+      date: date
     };
     const o = ParseObject.fromJSON(json);
     expect(o.className).toBe('Item');
@@ -268,8 +270,10 @@ describe('ParseObject', () => {
       size: 'medium',
       createdAt: new Date(Date.UTC(2013, 11, 14, 4, 51, 19)),
       updatedAt: new Date(Date.UTC(2013, 11, 14, 4, 51, 19)),
+      date
     });
     expect(o.dirty()).toBe(false);
+    expect(o.get('date')).toBeInstanceOf(Date);
   });
 
   it('can override old data when inflating from the server', () => {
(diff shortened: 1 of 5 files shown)


src/__tests__/ParseObject-test.js
```diff
@@ -277,6 +277,28 @@ describe('ParseObject', () => {
     expect(o.get('date')).toBeInstanceOf(Date);
   });
 
+  it('can be dirty with fromJSON', () => {
```
dplewis, line 280: Can you add an integration test?

dplewis (approved):
Nice! I remember this being an issue a while ago.

ChinaeduPascal:
Why is it I cannot find the changes in the latest version `3.4.0`. The method `fromJSON` still only takes these parameters:
`fromJSON(json: any, override?: boolean): T;`.

mtrezza:
@ChinaeduPascal This may be just a TypeScript issue. I suggest you look into the SDK code and whether the feature has been removed again. The tests in `integration/test/ParseObjectTest.js` should demo the feature. If not, please open a new issue.

ChinaeduPascal:
@mtrezza You are right. The @types/parse definitions are not up-to-date. That's why I won't see the changes.

Issue #1270, opened 2020-12-06
Support web workers with alternative to localStorage (IndexedDB ?)


FunMiles:
I have a Progressive Web App (PWA) that maintains a local copy of a user's data in an IndexedDB database. This app uses Firebase and I am trying to switch to Parse as it is not only open source, but would allow easier queries directly from the database than what Firebase supports. 
The synchronization between the local copy and the cloud database is done in a web worker as to not impact the UI performance. When implementing the Parse code in the web worker, I discovered an annoying stumbling block: Parse-SDK-JS keeps a record of the logged in user, session and other data using `localStorage`. However `localStorage` is not available in web workers. The documentation mentions some mechanisms used with react-native that relies on `Parse.setAsyncStorage(StorageController);`. However trying to implement my own StorageController, not from react-native, I discovered that this is not the right controller that I needed to do the first Parse operation. I had to use (and didn't find the doc about it, I had to read the code to find out the details of the call sequence that still used `localStorage`) `Parse.CoreManager.setStorageController(StorageController);`

It would be good to have an easier mechanism to handle storing the data in a way that can cover all cases. 

**The solution I'd like**
At minimum, the documentation needs to explain how to cover the web-worker case. However, though I can't speak for the react-native situation, IndexedDB is supported by both the browser side and the web-worker side. So substituting indexedDB to localStorage would be a solution covering both sides without any more consideration for a first-time user.

** Alternatives I've considered**
There are two alternatives I can see that could address the issue:

- Provide an alternative `StorageController` in the library for those who want to use Parse from web workers and provide the documentation on how to activate it correctly.
- There is a `setLocalDatastoreController` method, but it is rather obscure if it could help here. The comments for it mention the use of `setAsyncStorage` instead of this method. Shouldn't all this be unified so that a single customization point of local storage could cover all the uses of a local storage mechanism?

**Additional context**
The documentation of the various storage objects, their use, and inter-relations should be improved.

dplewis:
This is a good feature. Is this something you would like to work on?

dplewis:
@davimacedo What do you think? I'm not too familiar with IndexedDB or the database creation.

FunMiles:
[...]

PR #1297, opened 2021-02-13
feat: Add IndexedDB Storage Controller

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/1270

```js
Parse.CoreManager.setStorageController(Parse.IndexedDB);
```

The goal of this PR is to provide a default indexedDB implementation. Users are more than welcome to create their own storage controllers if they need transactions, indexes or schema migration etc...

Uses the lightweight, [idb-keyval](https://github.com/jakearchibald/idb-keyval) package instead of the full [idb](https://www.npmjs.com/package/idb)

@FunMiles Can you give it a try?

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,12 @@ const Parse = require('parse');
 import Parse from 'parse/dist/parse.min.js';
 ```
 
+For web worker or browser applications, indexedDB storage is available:
+
+```js
+Parse.CoreManager.setStorageController(Parse.IndexedDB);
+```
+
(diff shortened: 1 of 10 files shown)


davimacedo:
@dplewis do you believe there is a way to run the tests in ci for the scenario of using this controller (and not the default one)?

dplewis:
@davimacedo I can't think of anything off top of my head. Might have to look into jasmine

davimacedo (approved):
We can add the additional test scenario later

Issue #1296, opened 2021-02-12
linkWith() on React Native using currentUser()


eberperetto:
### New Issue Checklist
- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-JS/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse JS SDK](https://github.com/parse-community/Parse-SDK-JS/releases). 

### Issue Description
When using `linkWith()` on React Native using Google as a provider, it calls `currentUser()` on the background, which causes the following error:
```
Error: Cannot call currentUser() when using a platform with an async storage system. Call currentUserAsync() instead.
```

### Steps to reproduce
Call `linkWith()` on a new `Parse.User` object and pass Google as the provider along with the correct `authData` from Google Signin. I've set this project [here](https://github.com/eberperetto/ParseSDKJSlinkWithIssue) with the complete setup.

### Actual Outcome
Error message:
```
Error: Cannot call currentUser() when using a platform with an async storage system. Call currentUserAsync() instead.
```

### Expected Outcome
The linked `Parse.User` object.

###  Environment
React Native v0.63.4

Server
- Parse Server version: `4.4.0`
- Operating system: `Mac OS Catalina v10.15.7`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `local`

Database
- System (MongoDB or Postgres): `MongoDB`
- Database version: `4.4.1`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `local`

Client
- Parse JS SDK version: `2.19,0`

### Logs
```
Error: Cannot call currentUser() when using a platform with an async storage system. Call currentUserAsync() instead.
currentUser@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:111793:24
value@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:111502:38
value@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:111280:40
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:111393:31
[...]

PR #1298, opened 2021-02-13
fix: User management on React-Native

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/1296

This addresses issues using AsyncStorage.

* Introduces new `user.isCurrentAsync` function

### DIFF
--- a/src/ParseUser.js
+++ b/src/ParseUser.js
@@ -309,6 +309,16 @@ class ParseUser extends ParseObject {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


dplewis:
@davimacedo I just realized that this new `isCurrentAsync` could be used in other places. Should I add them to this PR or another?

davimacedo (approved):
LGTM!

PR #1302, opened 2021-02-22
Support PushStatus on Parse.Push

dplewis:
Similar to how jobStatus is handled.

* Parse.Push.send now returns the pushStatusId
* Add useMasterKey to Parse.Push.send
* Add new Parse.Push.getPushStatus
* Add new Push Integration tests.

```
const pushStatusId = await Parse.Push.send(payload);
const pushStatus = await Parse.Push.getPushStatus(pushStatusId);
```
This is currently a breaking change. Parse.Push.send returned `{ result: true }` before.

### DIFF
--- a/integration/test/ParsePushTest.js
+++ b/integration/test/ParsePushTest.js
@@ -1,0 +1,15 @@
+'use strict';
+
+const Parse = require('../../node');
+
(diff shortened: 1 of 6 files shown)


davimacedo (approved):
Great job! LGTM!

PR #1303, opened 2021-02-23
Allow nested increment on undefined fields

dplewis:
Calling increment on a field that doesn't exists (serverData stored in object state) gets ignored.
```
const obj = new TestObject();
obj.increment('a.b.c.d'); // ignored
await obj.save();
```
I believe this was ignored because the server doesn't support it and throws the following error
```
Unhandled promise rejection: TypeError: Cannot read property 'split' of undefined
```

### DIFF
--- a/integration/test/ParseObjectTest.js
+++ b/integration/test/ParseObjectTest.js
@@ -414,16 +414,6 @@ describe('Parse Object', () => {
     assert.equal(result.get('objectField').unknown, 20);
   });
 
-  it('ignore set nested fields on new object', async () => {
-    const obj = new TestObject();
-    obj.set('objectField.number', 5);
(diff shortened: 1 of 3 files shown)


mtrezza:
What is the change? Is a new field with this PR incremented when saving?

dplewis:
The change is that we are allowing this request to go to the server.

mtrezza:
Do we have enough integration test in this repo to ensure letting this go to the server has no detrimental effects?

dplewis:
Technically the other SDK's don't prevent you from doing this. Dot notation didn't exist in this SDK until I added it.
https://github.com/parse-community/Parse-SDK-JS/pull/729

Since there is that unhandled promise rejection. We can't really test much.

mtrezza (approved):
LGTM!

PR #1305, opened 2021-02-23
fix(EventuallyQueue): Poll against /health endpoint

dplewis:
The `/health` endpoint has been supported on the server since 2.2.5.
This fixes an issue where a `403` would let us know that server was online.

### DIFF
--- a/src/EventuallyQueue.js
+++ b/src/EventuallyQueue.js
@@ -308,12 +308,14 @@ const EventuallyQueue = {
     }
     polling = setInterval(() => {
       const RESTController = CoreManager.getRESTController();
-      RESTController.ajax('GET', CoreManager.get('SERVER_URL')).catch(error => {
-        if (error !== 'Unable to connect to the Parse API') {
+      RESTController.request('GET', 'health/')
+        .then(() => {
           this.stopPoll();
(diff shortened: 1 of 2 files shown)


dplewis:
What do you think about adding a health check in the SDK like the PHP SDK

https://github.com/parse-community/parse-php-sdk/pull/366

src/EventuallyQueue.js
```diff
@@ -308,12 +308,14 @@ const EventuallyQueue = {
...
-        if (error !== 'Unable to connect to the Parse API') {
+      RESTController.request('GET', 'health/')
+        .then(() => {
           this.stopPoll();
           return this.sendQueue();
-        }
-      });
+        })
+        .catch(() => {
+          // 'Unable to connect to the Parse API' keep polling
```
mtrezza, line 317: I don't mean to be pedantic, but would you mind catching the error and returning it `.catch(e => e);`, then checking for the status code to be 200? Empty `catch` blocks are just so hard to look at 😅

dplewis: Done! I like that `e => e` trick

mtrezza (approved):
LGTM!

Issue #1304, opened 2021-02-23
ParseObject.startsWith & endsWith cannot be use in a case-insensitive way


sadortun:
**Is your feature request related to a problem? Please describe.**
Our MongoDB use a collation/strength that allows for case-insensitive queries/indexes

Unfortunatly `ParseObject.startsWith()` and `ParseObject.endssWith()` use a `$regex` that is case-sensitive.

**Describe the solution you'd like**

Considering that `matches()` already allow `modifiers`,  would you consider adding a 2nd parameter to `startsWith/endsWith` to allow case-insensitive regex ?
```js
matches(key, regex , modifiers) 
   //....
    if (regex.ignoreCase) {
      modifiers += 'i';
    }
```

**Describe alternatives you've considered**

```js
// Currently using instead
q.matches('name', '^' +value, 'i')

//But it would be nice to ...
q.startsWith('name', value, 'i')
```

**Additional context**

LMK and i can submit a PR later this week

PR #1306, opened 2021-02-23
Add modifiers to startsWith / endsWith

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/1304

### DIFF
--- a/integration/test/ParseQueryTest.js
+++ b/integration/test/ParseQueryTest.js
@@ -1158,7 +1158,7 @@ describe('Parse Query', () => {
       });
   });
 
-  it('can test case insensitive regex', done => {
+  it('can test case insensitive matches', done => {
     const obj = new TestObject();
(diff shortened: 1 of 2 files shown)


mtrezza (approved):
Nice!

mtrezza:
I'm a bit surprised we have a separate `startsWith` end `endsWith`. I would deprecate these methods to keep the codebase small, because they just seem to add a single regex char to `matches`.

dplewis:
Yeah but nobody likes regex expressions and new users can just use the functions. They are also used in the dashboard as well.

mtrezza:
> Yeah but nobody likes regex expressions

@TomWFox when are we launching a Parse Platform merch shop and begin printing these things on t-shirts?

TomWFox:
http://teemill.com

PR #1307, opened 2021-02-23
Add Server Health Check

dplewis:
Adds `Parse.getServerHealth` as a quick way to check on the server health without having to make an arbitrary request to a functional endpoint, and without needing keys.

Similar to the PHP SDK
https://github.com/parse-community/parse-php-sdk/pull/366


mtrezza (approved):
LGTM!

Issue #1097, opened 2020-02-25
Allow saving with custom objectId


JeromeDeLeon:
**Is your feature request related to a problem? Please describe.**
This feature was already merged on Parse Server but I just tried it with this SDK and got the result of `Object not found` and upon looking up, [ParseObject#L313](https://github.com/parse-community/Parse-SDK-JS/blob/05b097c6256ca058f10a0c2551b01d1f321b123d/src/ParseObject.js#L313) was the cause of it because if it detects an objectId being present, it will simply update the object hence the `PUT` method. Correct me if I'm wrong. 

**Describe the solution you'd like**
Ability to set `objectId` without making an update to the non-existent record.

**Describe alternatives you've considered**
For now, we could just use REST API to manually create an object.

**Additional context**
I thought of a solution that we could check if it is new or not but looking up to `isNew` method, it checks whether `this.id` is present or not and unfortunately, hitting save will automatically replace `this.id` the `objectId` we assigned to it.

JeromeDeLeon:
At first, I thought of adding `_existed` property to `ParseObject` but seems like too cumbersome and not a good solution IMO, so I thought of another one. We could just replace `_localId` of the `objectId` that we set instead of `uuid`?

JeromeDeLeon:
We also need to support reverting `id` back to `undefined` in case it failed to save.

dplewis:
@JeromeDeLeon Thanks for your suggestions.

Have global config `CoreManager.get('AllowCustomId')` this will allow us to make sure that an objectId is required (or we could reuse _localId) before saving.

The next problem is with `PUT` vs `POST`. I think we can use the fact that a saved / existing object doesn't have a _localId.

@parse-community/core-maintainers Thoughts?

davimacedo:
It looks a good approach.

PR #1309, opened 2021-02-23
feat: Allow saving with custom objectId

dplewis:
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/1097
Ref: https://github.com/parse-community/parse-server/pull/6177

@JeromeDeLeon

### DIFF
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -324,10 +324,19 @@ class ParseObject {
   }
 
   _getSaveParams(): SaveParams {
-    const method = this.id ? 'PUT' : 'POST';
+    let method = this.id ? 'PUT' : 'POST';
     const body = this._getSaveJSON();
     let path = 'classes/' + this.className;
-    if (this.id) {
+    if (CoreManager.get('ALLOW_CUSTOM_OBJECT_ID')) {
+      if (!this.createdAt) {
+        method = 'POST';
+        path += '';
+        body.objectId = this.id;
+      } else {
+        method = 'PUT';
+        path += '/' + this.id;
+      }
+    } else if (this.id) {
       path += '/' + this.id;
     } else if (this.className === '_User') {
       path = 'users';
@@ -2362,6 +2371,7 @@ const DefaultController = {
 
     const RESTController = CoreManager.getRESTController();
     const stateController = CoreManager.getObjectStateController();
+    const allowCustomObjectId = CoreManager.get('ALLOW_CUSTOM_OBJECT_ID');
 
     options = options || {};
     options.returnStatus = options.returnStatus || true;
@@ -2384,6 +2394,12 @@ const DefaultController = {
         if (el instanceof ParseFile) {
           filesSaved.push(el.save(options));
         } else if (el instanceof ParseObject) {
+          if (allowCustomObjectId && !el.id) {
+            throw new ParseError(
+              ParseError.MISSING_OBJECT_ID,
+              'objectId must not be empty, null or undefined'
+            );
+          }
           pending.push(el);
         }
       });
@@ -2477,6 +2493,12 @@ const DefaultController = {
         });
       });
     } else if (target instanceof ParseObject) {
+      if (allowCustomObjectId && !target.id) {
+        throw new ParseError(
+          ParseError.MISSING_OBJECT_ID,
+          'objectId must not be empty, null or undefined'
+        );
+      }
       // generate _localId in case if cascadeSave=false
       target._getId();
       const localId = target._localId;
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -330,7 +330,6 @@ class ParseObject {
     if (CoreManager.get('ALLOW_CUSTOM_OBJECT_ID')) {
       if (!this.createdAt) {
         method = 'POST';
-        path += '';
         body.objectId = this.id;
       } else {
         method = 'PUT';
--- a/integration/test/ParseObjectTest.js
+++ b/integration/test/ParseObjectTest.js
@@ -2055,4 +2055,60 @@ describe('Parse Object', () => {
     expect(obj.get('string')).toBeDefined();
     expect(obj.get('string')).toBeInstanceOf(String);
   });
+
+  it('allowCustomObjectId', async () => {
+    await reconfigureServer({ allowCustomObjectId: true });
+    Parse.allowCustomObjectId = true;
+    const customId = `${Date.now()}`;
+    const object = new Parse.Object('TestObject');
+    try {
+      await object.save();
+      fail();
+    } catch (error) {
+      expect(error.message).toBe('objectId must not be empty, null or undefined');
+    }
+    object.id = customId;
+    object.set('foo', 'bar');
+    await object.save();
+    expect(object.id).toBe(customId);
+
+    const query = new Parse.Query('TestObject');
+    const result = await query.get(customId);
+    expect(result.get('foo')).toBe('bar');
+    expect(result.id).toBe(customId);
+
+    result.set('foo', 'baz');
+    await result.save();
+
+    const afterSave = await query.get(customId);
+    expect(afterSave.get('foo')).toBe('baz');
+    Parse.allowCustomObjectId = false;
+  });
+
+  it('allowCustomObjectId saveAll', async () => {
+    await reconfigureServer({ allowCustomObjectId: true });
+    Parse.allowCustomObjectId = true;
+    const customId1 = `${Date.now()}`;
+    const customId2 = `${Date.now()}`;
+    const obj1 = new TestObject({ foo: 'bar' });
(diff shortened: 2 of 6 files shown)


mtrezza (approved):
LGTM!

src/ParseObject.js
```diff
@@ -324,10 +324,19 @@ class ParseObject {
...
   _getSaveParams(): SaveParams {
-    const method = this.id ? 'PUT' : 'POST';
+    let method = this.id ? 'PUT' : 'POST';
     const body = this._getSaveJSON();
     let path = 'classes/' + this.className;
-    if (this.id) {
+    if (CoreManager.get('ALLOW_CUSTOM_OBJECT_ID')) {
+      if (!this.createdAt) {
+        method = 'POST';
+        path += '';
```
JeromeDeLeon, line 333: Is there any purpose for this?

JeromeDeLeon (commented):
LGTM! Just a question, if we were to merge this and update the Parse-Server, would it be allowed to use with save trigger?

src/ParseObject.js
dplewis: Nope

JeromeDeLeon (approved):

dplewis:
@JeromeDeLeon The purpose of this PR is to allow the customId to get sent to the server. I wrote a quick test and the beforeSave trigger gets called but the behavior looks wrong. Might need a server update.

src/ParseObject.js
```diff
@@ -2375,6 +2384,12 @@ const DefaultController = {
         if (el instanceof ParseFile) {
           filesSaved.push(el.save(options));
         } else if (el instanceof ParseObject) {
+          if (allowCustomObjectId && !el.id) {
```
mstniy, line 2387: This forces all objects to have custom ids, if custom ids are allowed.
Is this intentional? This is not how the server treats allowCustomObjectId.

breitman:
Hi, I am trying to set my own Id with this new feature. In my parse options I added the allowCustomObjectId to be true. Is there another thing i need to do with Parse JS SDK?

mstniy:
Currently, one also needs to set allowCustomObjectId on the client side, which I find to be weird. It is used to decide between doing a POST and a PUT, in conjunction with updatedAt, which I find to be even weirder. Perhaps a later PR can modify the js sdk to have a create()/update() seperation instead, like the Dart sdk?

mtrezza:
> Currently, one also needs to set allowCustomObjectId on the client side, which I find to be weird. It is used to decide between doing a POST and a PUT, in conjunction with updatedAt, which I find to be even weirder.

That indeed needs a closer look. Can you give more details / examples for that behavior?

breitman:
> Currently, one also needs to set allowCustomObjectId on the client side, which I find to be weird. It is used to decide between doing a POST and a PUT, in conjunction with updatedAt, which I find to be even weirder. Perhaps a later PR can modify the js sdk to have a create()/update() seperation instead, like the Dart sdk?

So what is the notation to allow for custom IDs on the client side? Parse-server automatically creates the js sdk so i am confused as where i declare it on client side

mstniy:
`Parse.allowCustomObjectId = true;` , as you can also see in the unit tests added by this PR. It can go anywhere, but it must be set by the time you call .save()
Another quirk is that this flag forces all objects to have a custom id, so it is really a `forceCustomObjectId` . I would love for this to change. In the meanwhile, do not forget to set it back to false after the operation involving custom object ids.

I am not sure what you mean by the server automatically creating the sdk.

breitman:
> `Parse.allowCustomObjectId = true;` , as you can also see in the unit tests added by this PR. It can go anywhere, but it must be set by the time you call .save()
> Another quirk is that this flag forces all objects to have a custom id, so it is really a `forceCustomObjectId` . I would love for this to change. In the meanwhile, do not forget to set it back to false after the operation involving custom object ids.
> 
> I am not sure what you mean by the server automatically creating the sdk.

That isn't working. I added that line to the top of my cloud function, and i am still getting an object not found error, due to the custom id....

mstniy:
Weird. Unless you are using an outdated version of the server, you might want to open an issue.
I can see the latest version on npm is 4.5.0, which was released before this PR was merged.

mtrezza:
> In the meanwhile, do not forget to set it back to false after the operation involving custom object ids.

Are you intending to switch between custom and non-custom IDs during runtime? I'm not sure that should even be possible.

mstniy:
> > Currently, one also needs to set allowCustomObjectId on the client side, which I find to be weird. It is used to decide between doing a POST and a PUT, in conjunction with updatedAt, which I find to be even weirder.
> 
> That indeed needs a closer look. Can you give more details / examples for what you described?
> 
> The difference between POST and PUT should whether the object already exists or is newly created, not how the object ID is set.

Please see [the code](https://github.com/parse-community/Parse-SDK-JS/blob/7512f533026431f7a085fb3f88d4b5c30430877c/src/ParseObject.js#L330-L337). Because the server doesn't support upsert (see parse-community/parse-server#7139), the SDK ends up having to guess whether to use POST (which fails if the object already exists) or PUT (which fails if the object does not already exist). The way it dos that is by checking if `createdAt` is set.

To see the way the Flutter SDK does it, see [its source](https://github.com/parse-community/Parse-SDK-Flutter/blob/aa6bed5df425ed1564261f7213177135bedb041a/packages/dart/lib/src/objects/parse_object.dart). It has a `.create()` that always does a `POST`, an`.update()` that always does a `PUT`, as well as a `.save()` that tries to guess.

mstniy:
> > In the meanwhile, do not forget to set it back to false after the operation involving custom object ids.
> 
> Are you intending to switch between custom and non-custom IDs during runtime? I'm not sure such a fundamental change should be possible during run-time, or even at all for a deployment.

We want to _allow_ custom ids, but the changes in this PR forces us to either always use custom object ids, or to disallow them altogether. The way we currently work around this is by toggling the flag.

lsmilek1:
In my application I found also a use case for the mixed environment of server generated and custom objectId ([ParseSwift SDK](https://github.com/parse-community/Parse-Swift/issues/220), [forum](https://community.parseplatform.org/t/both-custom-and-generated-objectid-in-one-project/1812)) and would like to implement *allow* instead of *force*.

From what I tested with forked ParseSwift SDK where I removed the guards that are checking that `objectId` is always set when `allowCustomObjectId = true` on both the server and the client, it works as intended:
- creating object with server generated `objectId` when `ParseObject.objectId = nil` and `ParseObject.createdAt = nil`
- creating object with custom `objectId` when `ParseObject.objectId = customId` and `ParseObject.createdAt = nil`
- updating object with any `objectId` when `ParseObject.objectId = any id` and `ParseObject.createdAt = any date`
- it throws error "Duplicate value" when trying to create object with custom id that already exists

So from this point of view there is no urgent need for separate `.create()` and `.update()` as long as it would be clearly documented that the `createdAt` is needed to decide whether the object should be updated or created - at least from client SDK, I am not 100% sure thing could break on server side JS SDK. But again, as it is already working and together with the documented need of `createdAt`, is there something blocking this to be implemented? In my use case if would safe few fields and indexes in the database together with extra fetches in the cloud code triggers.

cbaker6:
> This forces all objects to have custom ids, if custom ids are allowed.
Is this intentional? This is not how the server treats allowCustomObjectId.

Though what @mstniy stated above is true. I don't think removing the checks that @dplewis added should be done as they make sure the proper items are needed when using a custom objectId. Instead, I recommend using a approach similar to the Swift SDK https://github.com/parse-community/Parse-Swift/pull/222#issue-984092514 in which a flag, `` is added and the developer wants to ignore the check to use a mixed environment, they can. The reasons for this is in the warnings added to the documentation:

lsmilek1:
I see, that could be also a way to go. I am just still missing what is being protected exactly or at least it feels redundant, because from what I tested in ParseSwift SDK, it behaves correctly (assuming the `allowCustomObjectId = true` has to be set on the server). To summarise the behaviour I have in mind:

1) `allowCustomObjectId = false`
- `objectId = nil` --> save new object
- `objectId = generatedId` --> update object with `objectId == generatedId` or throw "does not exist"

2) `allowCustomObjectId = true`
- `objectId = nil`, `createdAt = nil` --> save new object with generated id
- `objectId = customId`, `createdAt = some date` --> save new object with custom id
- `objectId = someId`, `createdAt = some date` --> update object with `objectId == someId` or throw "does not exist"
- `objectId = someId`, `createdAt = nil` --> throw "duplicate value"

At least with ParseSwift SDK it work that way when I turned of the guards in forked SDK

src/ParseObject.js
bdevore17: I second this. Just turned this on, and this does not work as expected.

cbaker6: I recommend looking at https://github.com/parse-community/Parse-SDK-JS/pull/1309#issuecomment-909436697

PR #1317, opened 2021-03-11
Enable integration tests to be randomized

dplewis:
Allowing the test suite to run in a random order will help find test passes only based on a previously ran test.

* Adds new `Parse.Object.unregisterSubclass` feature
* Properly closes test created servers
* Generate unique username and password for user tests
* Properly handle unsorted queries (results returned in random order)
* Improve tests with background operations
* Remove setTimeout when possible


dplewis:
@davimacedo @mtrezza I ran the entire suite 20 times locally without any issues.

mtrezza (approved):
LGTM!

dplewis:
I hope so there are 2800 test in the server repo.

PR #1324, opened 2021-03-19
Remove deletion of Anonymous User on logout

dplewis:
Based on conversation [here](https://community.parseplatform.org/t/anonymous-user-destroyed-on-logout/1425/6)

* Remove deletion on logout
* Remove deletion on setCurrentUser
* Exposes registered on AnonymousUtils


davimacedo (approved):
It looks good to me and makes more sense now. Should we classify it as a breaking change?

PR #1341, opened 2021-04-12
Added  relative date support to OfflineQuery class

zanderisrael:
Since $relativeTime is supported by parse-server backend, and are possible to construct using the parse-sdk-js lib, iv added the support to the OfflineQuery class so this type of query is available when querying local pinned objects.

construction of a $relativeTime query: 
``` const date = new Date('1995-12-17T03:24:00');
const obj = new ParseObject('Item');
obj.set('field', date);
const q = new ParseQuery('Item');
q.lessThanOrEqualTo('field', { $relativeTime: 'in 0 day' }); 
```


dplewis:
Can you add an integration test https://github.com/parse-community/Parse-SDK-JS/blob/master/integration/test/ParseLiveQueryTest.js

zanderisrael:
> Can you add an integration test?
> 
> https://github.com/parse-community/Parse-SDK-JS/blob/master/integration/test/ParseLocalDatastoreTest.js

sure

zanderisrael:
> Can you add an integration test?
> 
> https://github.com/parse-community/Parse-SDK-JS/blob/master/integration/test/ParseLocalDatastoreTest.js

done

dplewis:
@mtrezza The build is failing because of the S3 change. Can you look into it?

package-lock.json
```diff
@@ -1903,67 +1903,6 @@
         }
       }
     },
-    "@parse/s3-files-adapter": {
```
dplewis, line 1906: @zanderisrael Can you revert the package-lock.json?

dplewis (changes requested):

mtrezza:
@dplewis From a quick look on mobile I think any dependency on the S3 adapter should be removed the JS SDK.

The solution would be to check whether there is any dependency in package.json and rebuild package-lock.json. If there are tests that need an adapter, then a duplicate of the mock adapter in Parse Server should be used.

package-lock.json
zanderisrael: I just deleted the package-lock.json file, Im not really sure why this is being tracked in the first place... 
It should be gitignored since the package.json file is what we need, the lock file is generated per environment when npm installing...

mtrezza:
@zanderisrael You would need to delete `package-lock`, run `npm i` to generate a new file, then commit.

zanderisrael:
> @zanderisrael You would need to delete `package-lock`, run `npm i` to generate a new file, then commit.

ok np.

package-lock.json
mtrezza: @zanderisrael, it is common practice to commit package-lock. It prevents unexpected behavior because it fixates dependencies to a specific version. Without committing it, every installation of Parse Server could vary in its dependencies.

mtrezza:
Has there been a misunderstanding or has this PR been abandoned?

zanderisrael:
> Has there been a misunderstanding or has this PR been abandoned?

I created a new one. I have little experience with OSS projects, sorry if I'm making a mess...

mtrezza:
I see, actually all that was necessary was to delete the package-lock file (as you already did), then run `npm i` which will create a new package-lock file. That new file would make the tests pass.

zanderisrael:
> I see, actually all that was necessary was to delete the package-lock file (as you already did), then run `npm i` which will create a new package-lock file. That new file would make the tests pass.

actually did that on new PR, but still tests don't pass :(

mtrezza:
Can you upload the PR?

zanderisrael:
> Can you upload the PR so we can have a look at the errors.

https://github.com/parse-community/Parse-SDK-JS/pull/1344

zanderisrael:
Hey about this, id like to support start of day, and end of day. In current implementation you can't set a range that is relative to today (not now)

PR #1343, opened 2021-04-13
[bug] Fix increment bug for nested array object field

KaustubhDamania:
Fixes https://github.com/parse-community/parse-server/issues/6687 

When I ran the below script, 

```js
const Parse = require('parse');

var jsondata = {
  _id: 'someId',
  items: [ { value: 'a', count: 5 }, { value: 'b', count: 1 } ],
  className: 'bug'
}

var object = Parse.Object.fromJSON(jsondata)
object.increment('items.0.count')
console.log(object.get('items'))
```
I got this output: 
```js
{ '0': { value: 'a', count: 6 }, '1': { value: 'b', count: 1 } }
```

The array was converted into a JSON object. I fixed this behaviour so that the resulting output is as expected:
```js
[ { value: 'a', count: 6 }, { value: 'b', count: 1 } ]
```
I have modified the `estimateAttribute` function logic in `src/ObjectStateMutations.js` and modified relevant tests accordingly.

### DIFF
--- a/integration/test/ParseObjectTest.js
+++ b/integration/test/ParseObjectTest.js
@@ -304,10 +304,7 @@ describe('Parse Object', () => {
     });
 
     obj.increment('objectField.number', 15);
-    assert.deepStrictEqual(obj.attributes.objectField, {
-      number: 20,
-      letter: 'a',
-    });
+    await obj.save();
     assert.deepStrictEqual(obj.attributes.objectField, {
       number: 20,
       letter: 'a',
@@ -381,7 +378,7 @@ describe('Parse Object', () => {
     } catch (error) {
       assert.equal(
         error.message,
-        'schema mismatch for TestObject.hello; expected String but got Object'
+        `Cannot create property 'dot' on string 'world'`
       );
     }
   });
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,44 +5,58 @@
   "requires": true,
   "dependencies": {
     "@apollo/client": {
-      "version": "3.3.7",
-      "resolved": "https://registry.npmjs.org/@apollo/client/-/client-3.3.7.tgz",
-      "integrity": "sha512-Cb0OqqvlehlRHtHIXRIS/Pe5WYU4hHl1FznXTRSxBAN42WmBUM3zy/Unvw183RdWMyV6Kc2pFKOEuaG1K7JTAQ==",
+      "version": "3.3.14",
+      "resolved": "https://registry.npmjs.org/@apollo/client/-/client-3.3.14.tgz",
+      "integrity": "sha512-z7YwMvF9grmpWUG+26e3gPcBAOA/r/Ci5gwK7JVm3bGYG9kKqG8MF6sMXEbuwTsFseE4duSp0icJ6tdzxJhhlA==",
       "dev": true,
       "requires": {
         "@graphql-typed-document-node/core": "^3.0.0",
         "@types/zen-observable": "^0.8.0",
-        "@wry/context": "^0.5.2",
-        "@wry/equality": "^0.3.0",
+        "@wry/context": "^0.6.0",
+        "@wry/equality": "^0.4.0",
         "fast-json-stable-stringify": "^2.0.0",
-        "graphql-tag": "^2.11.0",
+        "graphql-tag": "^2.12.0",
         "hoist-non-react-statics": "^3.3.2",
-        "optimism": "^0.14.0",
+        "optimism": "^0.15.0",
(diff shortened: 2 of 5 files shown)


package-lock.json
```diff
@@ -85,13 +85,40 @@
         "apollo-env": "^0.6.5"
       }
     },
-    "@apollographql/graphql-playground-html": {
-      "version": "1.6.26",
-      "resolved": "https://registry.npmjs.org/@apollographql/graphql-playground-html/-/graphql-playground-html-1.6.26.tgz",
-      "integrity": "sha512-XAwXOIab51QyhBxnxySdK3nuMEUohhDsHQ5Rbco/V1vjlP75zZ0ZLHD9dTpXTN8uxKxopb2lUvJTq+M4g2Q0HQ==",
+    "@apollographql/graphql-upload-8-fork": {
```
dplewis, line 88: Can you revert the this file?

integration/test/ParseObjectTest.js
```diff
@@ -381,7 +378,7 @@ describe('Parse Object', () => {
     } catch (error) {
       assert.equal(
         error.message,
-        'schema mismatch for TestObject.hello; expected String but got Object'
+        `Cannot create property 'dot' on string 'world'`
```
dplewis, line 381: Can you add a separate test to ensure that you can increment on nested array field

dplewis (changes requested):

package-lock.json
KaustubhDamania: Yeah I did it, maybe that's why the CI build was failing because it couldn't find `graphql` module

integration/test/ParseObjectTest.js
KaustubhDamania: Done :+1:

dplewis: I mean create a separate integration test. The unit tests looks fine but we would want to see real objects being saved and queried.

integration/test/ParseObjectTest.js
```diff
@@ -304,10 +304,7 @@ describe('Parse Object', () => {
     });
 
     obj.increment('objectField.number', 15);
```
dplewis, line 306: The following test fails. Can you add it to the PR?

```
it('can increment array nested fields', async () => {
    const obj = new TestObject();
    obj.set('items', [ { value: 'a', count: 5 }, { value: 'b', count: 1 } ]);
    await obj.save();

    obj.increment('items.0.count', 15);
    await obj.save();

    const result = await query.get(obj.id);
    assert.equal(result.get('items')[0].count, 20);
  });
```

dplewis (changes requested):

integration/test/ParseObjectTest.js
mtrezza: How would we differentiate between:
```
{
  "a": {
    "b": [
      { "c": 0 }
    ]
  }
}
```
and
```
{
  "a": {
    "b": {
      "0": {
        "c": 0
      }
    }
  }
}
```

mtrezza:
Does this PR also address calling other atonic ops on nested keys, such as `add` or `addUnique`? I would assume it touches on a common central logic?

src/ObjectStateMutations.js
```diff
@@ -135,7 +135,8 @@ export function estimateAttributes(
           const fields = attr.split('.');
           const first = fields[0];
           const last = fields[fields.length - 1];
-          data[first] = { ...serverData[first] };
+          // data[first] = { ...serverData[first] };
+          data[first] = (serverData[first] === undefined) ? {} : serverData[first];
```
mstniy, line 139: This line is buggy. It should be doing a deep-copy. Note that it is buggy even on master, except the master does a deep-copy on the first level, thanks to the spread operator (see [the Mozilla docs on the spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) ) and this patch does a complete shallow copy.
The shallow copy causes estimateAttributes() to actually change serverData as a side effect, though only for nested fields. The result is that a mere call to get() changes the object attributes, for nested fields with pending increments.
I will be creating a new issue for this.

mstniy (changes requested):

integration/test/ParseObjectTest.js
dplewis: @mtrezza https://github.com/parse-community/Parse-SDK-JS/pull/1451 fixes handles applying logic internally. The only problem is the Server. The server treats all dot notation as type Object so using dot notation an Array will result in a schema mismatch.

PR #1344, opened 2021-04-13
Added date support to OfflineQuery class

zanderisrael:
Hopefully this fixes the issues iv had with package-lock.json file. 

Added date support to OfflineQuery class

this enables you to use $relativeTime in local data store mode the same way you can use it via normal parseQuery. 
````
    const date = new Date('1995-12-17T03:24:00');
    const obj = new ParseObject('Item');
    obj.set('field', date);
    const q = new ParseQuery('Item');
    q.lessThanOrEqualTo('field', { $relativeTime: 'in 0 day' });
````


zanderisrael:
I'm sorry but im not sure what im doing wrong...

zanderisrael:
npm version issue maybe? used 7 downgrading to 6 now, maybe that will help...

mtrezza:
@dplewis Maybe we want to NPM version CI check in the JS SDK as well?

zanderisrael:
> @dplewis Maybe we want the NPM version CI check in the JS SDK as well?

sure, actually don't know how to do that tho...

mtrezza:
No worried, we'll add it in another PR so that others can identify the issue easier when they are building with npm 7.

dplewis (approved):
LGTM! If you want you could include this in LiveQuery

https://github.com/parse-community/parse-server/blob/master/src/LiveQuery/QueryTools.js

PR #1349, opened 2021-04-15
doc improvement

TravisDev:
added an example in the Parse.Object docs to show usage of saving a single attribute on a parse object. Users may use this in typescript class extensions of Parse.Object as a class attribute mutator.


davimacedo (approved):
Thanks for the PR! LGTM!

Issue #1348, opened 2021-04-15
Opinion - Makes the API more simple and readable like other platform


lamualfa:
**First of all, Thank you very much for this good library.**

There really isn't any problem that i get with the current feature and format. But I think, the Parse makes using the API is more complicated for ordinary people when first using it (_opinion_).

# Examples

## to Perform a Query

**Parse**
```js
const query = new Parse.Query(Parse.Object.extend("Blog"));
query.equalTo("title", "Hello world");

const data = await query.find();
```

**Supabase** (_The Open Source Firebase Alternative_)
```js
const { data, error } = await supabase.from('blog').select().eq('title', 'Hello world')
```

**Firebase**
```js
const data = await db.collection("blogs").where("title", "==", "Hello world").get()
```

## to Insert a New Data

**Parse**
```js
const Blog = Parse.Object.extend("Blog");
const blog = new Blog();

blog.set('title', 'Hello world')
await blog.save()
```

**Supabase**
```js
await supabase.from('blog').insert({ title: 'Hello world' })
```

**Firebase**
```js
await db.collection('blogs').add({ title: 'Hello world' });
```

## to Perform a Relational Query

**Parse**
```js
const User = Parse.Object.extend('user')
const Blog = Parse.Object.extend("Blog");

const author = new User();
author.id = 'abcdef'

const blog = new Blog();
blog.equalTo('author', author)
// I think just pass the author.id as a string in .equalTo will also work.
// But i haven't tried it. I just follow the documentation.
// https://docs.parseplatform.org/js/guide/#relational-queries

const result = await blog.find()
```

**Supabase**
```js
const { data } = await supabase.from('blog').select().eq('author', 'abcdef')
```

**Firebase**
```js
const result = await db.collection("blogs").where("author", "==", "abcdef").get()
```

<hr>

Do you have any plans to adapt the API ease of these examples? or are you going to stick with complex APIs now?

I don't think you need to delete the old API. You can enter the new method into the main core of the current library so that users have several alternatives for accessing the API.

I apologize. there is not much I can help you with. this is just a suggestion. You can close this issue at any time if it is not useful.

Thank you very much to all contributors

davimacedo:
I'm not sure if there is something to improve here as, if you prefer, you can also write the following codes using Parse SDK:

```js
// Find
const data = await new Parse.Query('Blog').equalTo('title', 'Hello world').find();

// Save
[...]

PR #1354, opened 2021-04-28
feat(Query): Add Parse.Query.from

dplewis:
ref: https://github.com/parse-community/Parse-SDK-JS/issues/1348

The goal of this PR is to make the Query API more readable and similar to use. This PR introduces `Parse.Query.from` which returns a Query instance without requiring the use of `new`.

Before
```
const data = await new Parse.Query('Blog').find();
```

After
```
const { Query } = require('parse');
const data = await Query.from('Blog').find();
```

### DIFF
--- a/src/ParseQuery.js
+++ b/src/ParseQuery.js
@@ -580,15 +580,10 @@ class ParseQuery {
    * Static method to create Parse.Query instance
    *
    * @param {string} className
-   * @param {object} eq key / value object for query.equalTo()
    * @returns {Parse.Query} new created query
    */
-  static from(className: string, eq: { [key: string]: any }): ParseQuery {
-    const query = new ParseQuery(className);
-    if (eq) {
-      query.equalTo(eq);
-    }
-    return query;
+  static from(className: string): ParseQuery {
+    return new ParseQuery(className);
   }
 
   /**
--- a/src/ParseQuery.js
+++ b/src/ParseQuery.js
@@ -577,6 +577,16 @@ class ParseQuery {
   }
 
   /**
+   * Static method to create Parse.Query instance
+   *
+   * @param {string} className
+   * @returns {Parse.Query} new created query
+   */
+  static from(className: string): ParseQuery {
+    return new ParseQuery(className);
+  }
+
+  /**
    * Static method to restore Parse.Query by json representation
    * Internally calling Parse.Query.withJSON
    *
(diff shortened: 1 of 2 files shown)


mtrezza:
It seems that `Query.from` only allows for `$eq` ops. How would we overload that method or extend it in the future to accept other ops?

davimacedo:
I was looking at a solution for passing the whole query (and not only the equal to) and I remembered we actually have Parse.Query.fromJSON (https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseQuery.js#L587), which is essentially the same thing, right?

mtrezza:
I don't see the advantage of before vs. after, rather the disadvantage that `.from` and `.fromJson` are not syntax checking. So for special use I see a purpose for `fromJson`, but for regular use, as in writing queries by hand, I don't.

dplewis:
I changed the OP. This PR just includes the static constructor.

mtrezza:
What is the purpose of this PR?

I think adding `from` to instantiate a query makes the code less readable / less intuitive, because:
- `from(Parse.Object)` is semantically inconsistent with the meaning of `fromJson(query)`. It's more like `for(Parse.Object)`.
- it requires more or the same amount of chars, so there doesn't seem to be any benefit:
  ```
  const q = new Parse.Query(o);
  const q = Parse.Query.from(o);
  const q = Parse.Query.for(o);
  ```

The only shortening I can think of is adding a static instructor:

```
const q = Parse.Query(o);
```

But then: why? It seems unintuitive to not require the `new` keyword when instantiating an object, it's just the expected programming syntax.

dplewis:
> You are unlikely to call new Parse.Query(o).find(); just like you're unlikely to call new Parse.Object(o).save();.

This is true but I do see users doing the following

```
const data = await (new Parse.Query('Test')).find()
```

I'm sure when we all learned how Promises worked we did this too.

mtrezza:
Hm, I still don't understand what the purpose of this PR is? Compensate for lack of programming knowledge?

dplewis:
Can you comment on the issue that was posted and provide feedback?

mtrezza:
Just [did](https://github.com/parse-community/Parse-SDK-JS/issues/1348#issuecomment-829144604), in a rather verbose way.

dplewis:
@davimacedo, @mtrezza Has some good points. What do you think? Should I close this?

davimacedo:
Yes. I think so.

PR #1356, opened 2021-05-06
[bug] Fix weapp uuid error.

onerain88:
In weapp, crypto is not provided.
So, I fake it with random string for installation id.


davimacedo:
It makes sense to me. What do you think @dplewis ?

dplewis:
The uuid module is used in other places besides the InstallationController. I think a separate `uuid.js` file would suffice.

onerain88:
I wrapped the uuid.js instead of uuid/v4

Issue #1372, opened 2021-06-08
Can not combine equalTo clause with any other clause


sadakchap:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-JS/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse JS SDK](https://github.com/parse-community/Parse-SDK-JS/releases). 

### Issue Description

Combining `equalTo` clause with any other clause seem to overwriting each other. 

### Steps to reproduce

```js
let q = new Parse.Query('MyClass');
q.equalTo('age', null);
q.exists('age');
// or these
q.equalTo('age', 10);
q.greaterThan('age', 0);
```

### Actual Outcome

getting a query like this `where={age: {$exists: true}}` or error in 2nd case(using greaterThan)

### Expected Outcome

`where={age: {$exists: true, $eq: null }}` or `where={age: {$gt: 0, $eq: 10 }}`

###  Environment

Server
- Parse Server version: `4.5.0`
- Operating system: `Ubuntu`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `local`

Database
- System (MongoDB or Postgres): `MongoDB`
- Database version: `4.4`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `local`

Client
- Parse JS SDK version: `3.2.0`

sadakchap:
There's the https://github.com/parse-community/Parse-SDK-JS/pull/1373 with a failing test case.

sadakchap:
After looking into [code](https://github.com/parse-community/Parse-SDK-JS/blob/005321fa9d820b7f5d5275c6c77b842c449b44eb/src/ParseQuery.js#L1225) for `equalTo`, wherever we had queries like this
```js
q.greaterThan(age, 0);
q.equalTo(age, 10); // this will override any existing query on that key
// will result in where={age: 10}
```
It seems like `equalTo` was always overriding any query on that key instead of adding new clause.


And when we try to apply other clauses to equalTo query,
```js
q.equalTo(age, 10);
[...]

PR #1373, opened 2021-06-08
Combine equalTo clause with other clauses

sadakchap:
https://github.com/parse-community/Parse-SDK-JS/issues/1372

have added only a failing test case.

### DIFF
--- a/src/__tests__/ParseQuery-test.js
+++ b/src/__tests__/ParseQuery-test.js
@@ -807,6 +807,21 @@ describe('ParseQuery', () => {
     });
   });
 
+  it('can combine equalTo clause with any other clause', () => {
+    const q = new ParseQuery('Item');
+    q.equalTo('inStock', null);
+    q.exists('inStock');
+
+    expect(q.toJSON()).toEqual({
+      where: {
+        inStock: {
+          $eq: null,
+          $exists: true,
+        },
+      },
+    });
+  });
+
   it('can specify ordering', () => {
     const q = new ParseQuery('Item');
     q.greaterThan('inStock', 0).ascending('createdAt');


davimacedo (changes requested):
@sadakchap do you mind to look at the failing tests?

Overall I like the idea. @dplewis what do you think?

mtrezza:
> have added only a failing test case.

@sadakchap it seems you have actually also added the fix in this PR, correct? If so, is this ready for review? I would like to add the related dashboard feature with the null filter.

sadakchap:
> If so, is this ready for review? I would like to add the related dashboard feature with the null filter.

Thanks @mtrezza , I think it still has some failing tests. I will try to fix those during this weekend and will let you know :slightly_smiling_face:

sadakchap:
@mtrezza Thanks for following up & waiting.

About failing test cases. Only [Live Query](https://github.com/parse-community/Parse-SDK-JS/blob/master/integration/test/ParseLiveQueryTest.js) test are failing as they depend on parse-server to fire events. We will need to update parse-server to send events for `$eq` queries in order to fix all failing test case.

src/ParseQuery.js
```diff
@@ -1222,7 +1222,8 @@ class ParseQuery {
       return this.doesNotExist(key);
     }
 
-    this._where[key] = encode(value, false, true);
+    // this._where[key] = encode(value, false, true);
```
mtrezza, line 1225: This comment seems to be a leftover?

src/__tests__/Cloud-test.js
```diff
@@ -301,7 +301,9 @@ describe('CloudController', () => {
     expect(data).toEqual({
       limit: 1,
       where: {
-        objectId: 'jobId1234',
```
mtrezza, line 304: Should all these tests be modified? That means the original syntax would not be tested anymore?

mtrezza: Could that be a breaking change for existing deployments?

PR #1375, opened 2021-06-20
Fix: explain should return as an array of results

cbaker6:


mtrezza (changes requested):
Mergeability depends on discussion https://github.com/parse-community/parse-server/pull/7440#discussion_r779542926.

dplewis:
@mtrezza Any news on this? I'd recommend we document either explain doesn't support first / get queries or that if explain is used an array is return here.

mtrezza:
Whatever causes the least breaking changes while providing the most functional support.

PR #1351, opened 2021-04-19
fix EventEmitter undefined on React Native 0.64

bduyng:
fixes #1335

PR #1381, opened 2021-07-06
Fix react native build

davimacedo:
Improve https://github.com/parse-community/Parse-SDK-JS/pull/1351

### DIFF
--- a/src/EventEmitter.js
+++ b/src/EventEmitter.js
@@ -10,7 +10,10 @@
  */
 
 if (process.env.PARSE_BUILD === 'react-native') {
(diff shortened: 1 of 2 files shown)


dplewis (commented):
LGTM

dplewis (approved):

davimacedo:
@dplewis @mtrezza I'd like to release this fix. Do you think there is any break change here that we need to wait for Parse Server release before releasing the SDK?

mtrezza:
If there was a breaking change, wouldn't the tests fail with 4.5.0?

PR #1384, opened 2021-07-12
Fix doc build

davimacedo:


src/OfflineQuery.js
```diff
@@ -83,11 +83,11 @@ function equalObjectsGeneric(obj, compareTo, eqlFn) {
 /**
  * Converts human readable relative date string
  * ie: 'in 10 days' to a date relative to now
+ * copied from parse-server/src/Adapters/Storage/Mongo/MongoTransform.js
  *
  * @param {*} text
  * @param {*} now
- * @returns {status,info,result}
- * copied from parse-server/src/Adapters/Storage/Mongo/MongoTransform.js
+ * @returns {{status:string,info:string,result:Date}}
```
mtrezza, line 90: The docs only say return type is `Object`, without any more details:

image

Is that what you intended? Or should we add a @typedef to document that output object in detail?

mtrezza: I added a @typedef to complete the docs, is that ok?

image

davimacedo:
Yes. It looks good to me. It's important to merge this one because the docs failed to build on last release.

mtrezza (approved):

mtrezza:
Yes, sorry for the long wait time, I noticed earlier that I only subscribed to releases of this repo, so I never received any notifications for review.

Issue #799, opened 2019-04-22
Parse.Object.saveAll error handling


foobar8675:
**Is your feature request related to a problem? Please describe.**
When doing `await Parse.Object.saveAll(objectsToUpdate, { useMasterKey: true });` if it fails I don’t know which `objectToUpdate` failed.

**Describe the solution you'd like**
An error telling me which object failed. That would allow me to write some code to properly handle the error.

dplewis:
Can you post logs? Running server with VERBOSE=1

foobar8675:
thanks for writing back @dplewis . I've always been a bit challenged when it came to the logs. I have the VERBOSE=1 set in my .env but never have seen verbose logs written out to the console.

i have a snippet here to recreate the issue
```
  try {
    ideas[1].id = "abc"
    const updatedIdeas = await Parse.Object.saveAll(ideas, { useMasterKey: true });
    return { updatedIdeas };
  } catch (e) {
    throw e;
  }
```
where the exception looks like

```
[[StableObjectId]]:8
code:101
message:"Object not found."
stack:"Error: Object not found.\n    at batchReturned.then (/Users/mattc/Documents/DecorMatters/dm_server/Server/node_modules/parse-server/node_modules/parse/lib/node/ParseObject.js:2299:35)\n    at process._tickCallback (internal/process/next_tick.js:68:7)"
__proto__:Error {constructor: , toString: }

```

is it possible to let me know the `1` index is the one that failed?

dplewis:
@foobar8675 Look at the [test suite](https://github.com/parse-community/Parse-SDK-JS/blob/master/integration/server.js#L26) for logging example

foobar8675:
got it. i re-ran and here is what i see in the logs. was thinking it would be nice if we knew which object was causing the error.

```
{"functionName":"deleteIdeas2","error":{"message":"Object not found.","code":101},"params":{"objectIds":["u9mm18znMo","VlxBdBCfPN"]},"user":"v5qg9E6sM3","level":"error","message":"Failed running cloud function deleteIdeas2 for user v5qg9E6sM3 with:\n  Input: {\"objectIds\":[\"u9mm18znMo\",\"VlxBdBCfPN\"]}\n  Error: {\"message\":\"Object not found.\",\"code\":101}","timestamp":"2019-04-25T18:02:33.606Z"}
[...]

PR #1432, opened 2021-11-22
feat: add Parse.Object.saveAllSettled to support more robust saving of multiple Parse Objects

simonaberry:

### Issue Description

Related issue:  #494 #799 #1433 and [this StackOverflow issue](https://stackoverflow.com/questions/34951768/parseobject-saveall-sometimes-results-in-duplicate-objects-being-created)

`Parse.Object.saveAll` can create data consistency issues. Because `saveAll` (which uses the `/batch` endpoint) will reject if *any* of the save items in the list fails, you can end up with 19 saved objects and 1 unsaved object. If the cause of the failure was a network delay, then the `saveAll` will automatically retry, with *all* 20 objects again - which leads to duplicated objects on the server.

 

### Approach

`Parse.Object.saveAllSettled` fixes this problem, by using the [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) method and individual, staggered `object. save()` calls - which waits for the saving of all the objects in the list to resolve or reject, then only tries to resave any objects that didn't successfully save the first time.

### TODOs before merging

- [ ] Add tests
- [x] Add entry to changelog
- [ ] ?support passing of context?

### DIFF
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -1800,6 +1800,94 @@ class ParseObject {
   }
 
   /**
+   * Saves the given list of Parse.Object.
+   * Will keep trying to save all items lists, even if ecountering errors, up to options.retyMax (default 3).
+   *
+   * <pre>
+   * Parse.Object.saveAllSettled([object1, object2, ...], options)
+   * .then((list) => {
+   * // All the objects were saved.
+   * }, (error) => {
+   * // An error occurred while saving one of the objects.
+   * });
+   * </pre>
+   *
+   * @param {Array} list A list of <code>Parse.Object</code>.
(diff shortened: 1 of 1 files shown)


simonaberry:
I tried to do some test, but could not figure out how the `xhr.open.mock.calls` bit works ... bit out of my depth there

mtrezza:
Could you please create an issue or reference an existing issue for this PR.

mtrezza:
Closing as the issue has been identified as merely a docs issue and transferred to docs repo as https://github.com/parse-community/docs/issues/859.

Issue #1347, opened 2021-04-14
Stale data with Cloud functions and deleted fields


aforemny:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-JS/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse JS SDK](https://github.com/parse-community/Parse-SDK-JS/releases). 

### Issue Description
Objects returned by Cloud functions may cause the client to return stale data when used with deleted fields.

### Steps to reproduce

- Suppose there is a collection `Item` with optional String-field `label`
- Suppose that collections initially contains one Item with `label: 'foobar'`
- Suppose there is a cloud function `listItems` that returns the list of items, ie.
```js
Parse.Cloud.define('listItems', async req => await new Parse.Query(Parse.Object.extend('Item')).find())
```
- Call that cloud function and list the item's labels, ie.
```js
(await Parse.Cloud.run('listItems')).map(item => item.get('label')) ==> [ 'foobar' ]
```
- In Parse dashboard, delete the label of that one Item so that its value becomes undefined
- Call that cloud code function again, without invalidating the client cache (without reloading the browser page):
```js
(await Parse.Cloud.run('listItems')).map(item => item.get('label')) ==> [ 'foobar' ]
```

### Actual Outcome
```js
(await Parse.Cloud.run('listItems')).map(item => item.get('label')) ==> [ 'foobar' ]
```

### Expected Outcome
```js
(await Parse.Cloud.run('listItems')).map(item => item.get('label')) ==> [ undefined ]
```

###  Environment

Server
- Parse Server version: master / 4.5.0
- Operating system: Linux
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): local

Database
- System (MongoDB or Postgres): MongoDB
- Database version: 4.2.8
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): local

Client
- Parse JS SDK version: 3.2.0

### Logs

It is observable that the second response does not include `"label": "foobar"` in its response.

aforemny:
[...]

PR #1442, opened 2022-01-19
fix: user object may contain already deleted fields

rdhelms:

### Issue Description
As discussed in #1347, the return from `Parse.User.logIn` can include fields that no longer exist on the user, even if the network response from parse-server does not include the removed fields. There is also an issue with the return from `Parse.Cloud.run`.

Related issue: #1347

### Approach
This PR is a work in progress, initially containing only a failing test for the `Parse.User.logIn` case. If the failing test is confirmed to represent the issue well, a failing test for `Parse.Cloud.run` can also be added.

No solution for the `Parse.User.logIn` case has yet been proposed.
The [proposed solution](https://github.com/parse-community/Parse-SDK-JS/issues/1347#issuecomment-819539080) for the `Parse.Cloud.run` case is to update one of the cases in `decode.js` to call `ParseObject.fromJSON(value, true)` instead of `ParseObject.fromJSON(value)`.

### TODOs before merging

- [x] Add failing test for `Parse.Cloud.run` case
- [x] Add failing test for `Parse.User.logIn` case
- [ ] Fix `Parse.Cloud.run` case
- [ ] Fix `Parse.User.logIn` case
- [ ] Add entry to changelog

### DIFF
--- a/src/Cloud.js
+++ b/src/Cloud.js
@@ -121,6 +121,7 @@ const DefaultController = {
       if (typeof res === 'object' && Object.keys(res).length > 0 && !res.hasOwnProperty('result')) {
         throw new ParseError(ParseError.INVALID_JSON, 'The server returned an invalid response.');
       }
+      ParseObject._clearAllState();
       const decoded = decode(res);
       if (decoded && decoded.hasOwnProperty('result')) {
         return Promise.resolve(decoded.result);
--- a/src/Cloud.js
+++ b/src/Cloud.js
@@ -121,7 +121,6 @@ const DefaultController = {
       if (typeof res === 'object' && Object.keys(res).length > 0 && !res.hasOwnProperty('result')) {
         throw new ParseError(ParseError.INVALID_JSON, 'The server returned an invalid response.');
       }
-      ParseObject._clearAllState();
       const decoded = decode(res);
       if (decoded && decoded.hasOwnProperty('result')) {
         return Promise.resolve(decoded.result);
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -347,9 +347,12 @@ class ParseObject {
     };
   }
 
-  _finishFetch(serverData: AttributeMap) {
+  _finishFetch(serverData: AttributeMap, override?: boolean) {
     if (!this.id && serverData.objectId) {
       this.id = serverData.objectId;
+    }
+    if (override) {
+      this._clearServerData();
     }
     const stateController = CoreManager.getObjectStateController();
     stateController.initializeState(this._getStateIdentifier());
--- a/src/ParseUser.js
+++ b/src/ParseUser.js
@@ -1088,6 +1088,7 @@ const DefaultController = {
         stateController.setPendingOp(user._getStateIdentifier(), 'username', undefined);
         stateController.setPendingOp(user._getStateIdentifier(), 'password', undefined);
         response.password = undefined;
+        user._clearServerData();
         user._finishFetch(response);
         if (!canUseCurrentUser) {
           // We can't set the current user, so just return the one we logged in
--- a/src/ParseUser.js
+++ b/src/ParseUser.js
@@ -1088,8 +1088,7 @@ const DefaultController = {
         stateController.setPendingOp(user._getStateIdentifier(), 'username', undefined);
         stateController.setPendingOp(user._getStateIdentifier(), 'password', undefined);
         response.password = undefined;
-        user._clearServerData();
-        user._finishFetch(response);
+        user._finishFetch(response, true);
         if (!canUseCurrentUser) {
           // We can't set the current user, so just return the one we logged in
           return Promise.resolve(user);
--- a/src/__tests__/Cloud-test.js
+++ b/src/__tests__/Cloud-test.js
@@ -15,15 +15,22 @@ jest.dontMock('../ParseError');
 jest.dontMock('../ParseObject');
 jest.dontMock('../ParseQuery');
 jest.dontMock('../Push');
+jest.dontMock('../parseDate');
+jest.dontMock('../ObjectStateMutations');
+jest.dontMock('../SingleInstanceStateController');
+jest.dontMock('../UniqueInstanceStateController');
 
 const Cloud = require('../Cloud');
 const CoreManager = require('../CoreManager');
 const Push = require('../Push');
+const ParseObject = require('../ParseObject').default;
 
 const defaultController = CoreManager.getCloudController();
 
 describe('Cloud', () => {
   beforeEach(() => {
+    ParseObject.enableSingleInstance();
+
     const run = jest.fn();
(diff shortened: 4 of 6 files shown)


mtrezza (commented):
Thanks for the test! Do you think you could rewrite this in async/await? We are slowly getting rid of promise chains in tests.

rdhelms:
@mtrezza Rewrote using async/await 👍 Happy to make any other adjustments. I also started trying to write a failing test in `Cloud-test.js`, but for some reason the `attributes` of my mock response aren't getting processed as I expected...any suggestions here?

```ts
    CoreManager.setRESTController({
      request (method, path) {
        expect(path).toBe('functions/myfunction')

        return Promise.resolve(
          {
            result: {
              objectId: 'abc123',
              className: 'Item',
              __type: 'Object',
              createdAt: '2015-01-01T00:00:00.000Z',
              updatedAt: '2015-01-01T00:00:00.000Z',
              label: 'foobar',
            },
          },
          200
        );
      },
      ajax () {},
    });

    Cloud.run('myfunction').then(response => {
      expect(response.get('label')).toBe('foobar')
    });
```

Giving this error during `npm test`:


And I'm seeing this in the debugger for `response`:

rdhelms:
@mtrezza I finished creating the failing test for the `Cloud.run` case that @aforemny identified in #1347, so both of the identified cases now have failing tests. Is this enough info for you guys to start working on a solution?

A couple notes:
- The issue with my previous attempt to write the `Cloud.run` failing test was that the test was running through too much mocked logic, so I had to add more `jest.dontMock` statements to the top of the file. This made me wonder if more of the test files should have those `jest.dontMock` statements (or maybe adjusting jest's `automock` behavior) so that it's easier for contributors to add new tests (and so that more of Parse's actual logic is tested)
- The [comment](https://github.com/parse-community/Parse-SDK-JS/issues/1347#issuecomment-819539080) from @aforemny about the `Cloud.run` issue only happening with `SingleInstanceController` can be verified by removing `ParseObject.enableSingleInstance()` from `Cloud-test.js`. The test case only fails when the single instance controller is enabled.

Here are the failing tests:

**Parse.User.logIn case**
```
FAIL src/__tests__/ParseUser-test.js (13.[23](https://github.com/parse-community/Parse-SDK-JS/runs/5055265717?check_suite_focus=true#step:9:23)6s)
  ● ParseUser › user who logs out, deletes attribute, then logs back in has correct attributes

    expect(received).toBe(expected) // Object.is equality

    Expected: undefined
    Received: "This field is returned in the first login but not the second"

      993 |     expect(user2.id).toBe('uid2');
      994 |     expect(user2.get('username')).toBe('username');
    > 995 |     expect(user2.get('fieldToBeDeleted')).toBe(undefined); // Failing test PR #1442
          |                                           ^
      996 |   });
      997 |
      998 |   it('can retreive a user with sessionToken (me)', async () => {

      at Object.<anonymous> (src/__tests__/ParseUser-test.js:995:43)
```

**Parse.Cloud.run case**
```
FAIL src/__tests__/Cloud-test.js
  ● CloudController › run same function twice with different responses

    expect(received).toBe(expected) // Object.is equality

    Expected: undefined
    Received: "foobar"

      263 |     const response2 = await Cloud.run('myfunction');
      264 |     expect(response2.get('label2')).toBe('control to confirm correct mock usage');
    > 265 |     expect(response2.get('label')).toBe(undefined); // Failing test PR #1442
          |                                    ^
      266 |   });
      267 |
      268 |   it('startJob passes encoded requests', () => {

      at Object.<anonymous> (src/__tests__/Cloud-test.js:265:36)
```

mtrezza:
Thanks! While writing the tests, did you also identify possible solutions for the fix?

rdhelms:
@mtrezza The most straightforward "solution" that comes to mind is simply disabling the Single Instance Controller. It seems to me that there may be a fundamental flaw in the Single Instance Controller's assumption that its `objectState` is reliable. It's completely possible that data could have been changed in Mongo in such a way that the Single Instance Controller's state is stale. And if that's the case, then I'm not sure how we can ever use the Single Instance Controller reliably.

But I don't know the context behind the creation and intended use case of the Single Instance Controller, and I don't know how significant the potential benefits are. If it has demonstrable value, then perhaps we can figure out a way to be smarter about clearing the state in the particular cases that we're dealing with here (`Parse.Cloud.run` and `Parse.User.logIn`).

If I don't enable the Single Instance Controller in the `Cloud` and `ParseUser` tests, the only thing that fails is the test about maintaining the session token when refetched. If that behavior is essential, then we'd need to find a way to still use the Single Instance Controller when it's needed, while resetting the state whenever there's a chance the controller's state is stale.

```
 FAIL  src/__tests__/ParseUser-test.js
  ● ParseUser › maintains the session token when refetched

    expect(received).toBe(expected) // Object.is equality

    Expected: "abc141"
    Received: undefined

      1286 |       );
      1287 |       expect(u.getSessionToken()).toBe('abc141');
    > 1288 |       expect(u2.getSessionToken()).toBe('abc141');
           |                                    ^
      1289 |       expect(u.get('number')).toBe(undefined);
      1290 |       expect(u2.get('number')).toBe(undefined);
      1291 |       done();

      at src/__tests__/ParseUser-test.js:1288:36
```

mtrezza:
> It's completely possible that data could have been changed in Mongo in such a way that the Single Instance Controller's state is stale. 

What could that be for example?

rdhelms:
@mtrezza In our case, we noticed that someone using our web app and mobile app at the same time could end up in a bad state. The web app was fetching the correct data from the server, but the web client was still using stale data that had been removed via the mobile app.

But, in general, any direct interaction with the data in MongoDB will result in the Single Instance Controller's state being out of date.

rdhelms:
@mtrezza As an example of what I'm talking about, I just pushed a commit with all the unit tests passing where I call `ParseObject._clearAllState()` in `Cloud.run`, and `user._clearServerData()` in `User.logIn`. But I'm not sure if that defeats the general intent behind the Single Instance controller.

rdhelms:
@mtrezza It looks like clearing the data as I mentioned causes a couple twitter/facebook flows to fail in the integration tests. If you or anyone else has any suggestions for ways to update the Single Instance controller to be more reliable, I'm happy to make more updates here. But in the short term on our end, we're probably going to disable the Single Instance controller.

rdhelms:
@mtrezza It turns out that disabling single instance mode has some pretty terrible side effects when it comes to how Parse updates the current user in local storage...so I think we're going to need to continue trying a solution here within the SDK. I can continue investigating when I have the time, but any assistance would be greatly appreciated.

It would also be incredibly helpful if the impact of enabling or disabling single instance mode could be better documented.

mtrezza:
Maybe you could take a look at how single instance is used in code and what effects it has to suggest an amendment to the docs.

PR #1467, opened 2022-03-29
fix: creating a Parse.File with base64 string fails for some file types

### LINKED ISSUE #1466: Wrong processing of a base64 string in accordance to RFC 2397 when calling new Parse.File
### New Issue Checklist

### Issue Description

It is common practice to specify other parameters in the date URI such as `;filename="parse.pdf"`.
I discovered this issue after generating a pdf in the browser and tried to create a new file using `Parse.File(...)`.

I created the following test to point out the error:
```js
it('can extract data type from base64 with a filename parameter', () => {
  const file = new ParseFile('parse.pdf', {
    base64: 'data:application/pdf;filename=parse.pdf;base64,ParseA==',
  });
  expect(file._source.base64).toBe('ParseA==');
  expect(file._source.type).toBe('application/pdf');
});
```

With the following results:
image

The main problem is that the currently used regex doesn't take into account that there may be other parameters (such as `filename`). It takes into account only `charset`.
```js
const dataUriRegexp = /^data:([a-zA-Z]+\/[-a-zA-Z0-9+.]+)(;charset=[a-zA-Z0-9\-\/]*)?;base64,/;
```

### Steps to reproduce

Create a new Parse.File with data starting with `data:application/pdf;filename=test.pdf;base64,`

### Actual Outcome

`new Parse.File(...)` throws an error at line 150 in ParseFile.js.

### Expected Outcome

`new Parse.File(...)` should return a new object even when there are other parameters specified.

###  Environment

Client
- Parse JS SDK version: `3.4.1`

### Logs

- n/a

RazvanCristian:

### Issue Description

Related issue: [#1466](https://github.com/parse-community/Parse-SDK-JS/issues/1466)

### Approach

Reworked logic for base64 formatting in ParseFile.js.

### TODOs before merging

- [x] Add tests
- [x] Add entry to changelog

### DIFF
--- a/changelogs/CHANGELOG_release.md
+++ b/changelogs/CHANGELOG_release.md
@@ -54,6 +54,8 @@ for (const className of classNames) {
 ### Bug Fixes
 - Fixes build for WeChat WeApp, to reduce package size, see [issue/#1331](https://github.com/parse-community/Parse-SDK-JS/issues/1331)
 
+- Fixes base64 formatting, see [issue/#1467](https://github.com/parse-community/Parse-SDK-JS/pull/1467)
+
 # [3.1.0](https://github.com/parse-community/Parse-SDK-JS/compare/3.0.0...3.1.0)
 
 ### Breaking Changes
--- a/changelogs/CHANGELOG_release.md
+++ b/changelogs/CHANGELOG_release.md
@@ -54,8 +54,6 @@ for (const className of classNames) {
 ### Bug Fixes
 - Fixes build for WeChat WeApp, to reduce package size, see [issue/#1331](https://github.com/parse-community/Parse-SDK-JS/issues/1331)
 
-- Fixes base64 formatting, see [issue/#1467](https://github.com/parse-community/Parse-SDK-JS/pull/1467)
-
 # [3.1.0](https://github.com/parse-community/Parse-SDK-JS/compare/3.0.0...3.1.0)
 
 ### Breaking Changes
--- a/src/ParseFile.js
+++ b/src/ParseFile.js
@@ -158,7 +158,7 @@ class ParseFile {
         // If <mediatype> is omitted, it defaults to text/plain;charset=US-ASCII.
         // As a shorthand, "text/plain" can be omitted but the charset parameter supplied.
         if (dataUriRegex.test(data.base64)) {
-          type ||= 'text/plain';
+          type = type || 'text/plain';
         }
 
         this._source = {
--- a/src/ParseFile.js
+++ b/src/ParseFile.js
@@ -148,7 +148,9 @@ class ParseFile {
         // Check if data URI or base64 string is valid
         const validationRegex = new RegExp(base64Regex.source + '|' + dataUriRegex.source, 'i');
         if (!validationRegex.test(data.base64)) {
-          throw new Error('Files passed in must have valid data URIs or base64 encoded data.');
+          throw new Error(
+            'Cannot create a Parse.File without valid data URIs or base64 encoded data.'
+          );
         }
 
         const base64 = data.base64.split(',').slice(-1)[0];
--- a/src/ParseFile.js
+++ b/src/ParseFile.js
@@ -42,6 +42,16 @@ export type FileSource =
       type: string,
     };
 
+const base64Regex = new RegExp(
+  '([0-9a-zA-Z+/]{4})*(([0-9a-zA-Z+/]{2}==)|([0-9a-zA-Z+/]{3}=))',
+  'i'
+);
+
+const dataUriRegex = new RegExp(
+  `^\\s*data:([a-zA-Z]+\\/[-a-zA-Z0-9+.]+(;[a-z-]+=[a-zA-Z0-9+.-]+)?)?(;base64)?,(${base64Regex.source})*\\s*$`,
+  'i'
+);
+
 function b64Digit(number: number): string {
   if (number < 26) {
     return String.fromCharCode(65 + number);
@@ -135,6 +145,12 @@ class ParseFile {
           type: specifiedType,
         };
       } else if (data && typeof data.base64 === 'string') {
+        // Check if data URI or base64 string is valid
+        const validationRegex = new RegExp(base64Regex.source + '|' + dataUriRegex.source, 'i');
+        if (!validationRegex.test(data.base64)) {
+          throw new Error('Files passed in must have valid data URIs or base64 encoded data.');
+        }
+
         const base64 = data.base64.split(',').at(-1);
         const type = specifiedType || data.base64.split(';').at(0).split(':').at(1) || '';
 
--- a/src/ParseFile.js
+++ b/src/ParseFile.js
@@ -42,8 +42,6 @@ export type FileSource =
       type: string,
     };
 
-const dataUriRegexp = /^data:([a-zA-Z]+\/[-a-zA-Z0-9+.]+)(;charset=[a-zA-Z0-9\-\/]*)?;base64,/;
-
 function b64Digit(number: number): string {
   if (number < 26) {
     return String.fromCharCode(65 + number);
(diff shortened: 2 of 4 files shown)


changelogs/CHANGELOG_release.md
```diff
@@ -54,6 +54,8 @@ for (const className of classNames) {
 ### Bug Fixes
 - Fixes build for WeChat WeApp, to reduce package size, see [issue/#1331](https://github.com/parse-community/Parse-SDK-JS/issues/1331)
 
+- Fixes base64 formatting, see [issue/#1467](https://github.com/parse-community/Parse-SDK-JS/pull/1467)
```
mtrezza: Changelog entries are not required anymore, please revert any changes to the file.

src/__tests__/ParseFile-test.js
```diff
@@ -67,11 +67,17 @@ describe('ParseFile', () => {
...
     });
-    expect(file._source.base64).toBe('ParseA==');
-    expect(file._source.type).toBe('image/png');
+    expect(file1._source.base64).toBe('ParseA==');
+    expect(file1._source.type).toBe('image/png');
+
+    const file2 = new ParseFile('parse.txt', {
+      base64: 'data:application/pdf;filename=test.pdf;base64,ParseA==',
+    });
+    expect(file2._source.base64).toBe('ParseA==');
```
mtrezza: Does this throw an error without the fix? Can you please revert changes to the existing test and add a new test with a description that reflects what this PR fixes? For example, why does it throw an error without the fix and what is it specifically that we’re testing? The PR title should also reflect that.

mtrezza (changes requested):

src/__tests__/ParseFile-test.js
RazvanCristian: Hi, the description was made in a hurry, I updated the issue to better describe the problem. (see #1466)  
I also updated the tests:
image

src/ParseFile.js
```diff
@@ -137,26 +135,14 @@ class ParseFile {
...
-        } else {
-          this._data = base64;
-          this._source = {
-            format: 'base64',
-            base64: base64,
-            type: specifiedType,
-          };
-        }
+        const base64 = data.base64.split(',').at(-1);
+        const type = specifiedType || data.base64.split(';').at(0).split(':').at(1) || '';
```
mtrezza: This is a less strict string parsing than the previous regex. What are the benefits of removing the original regex and could a less strict parsing create potential vulnerabilities?

src/__tests__/ParseFile-test.js
```diff
@@ -67,13 +67,21 @@ describe('ParseFile', () => {
   });
 
   it('can extract data type from base64', () => {
-    const file = new ParseFile('parse.txt', {
+    const file = new ParseFile('parse.png', {
```
mtrezza, line 78: Not sure whether this was deliberately a `txt` extension to demonstrate that the image type is derived from the data, not from the extension?

RazvanCristian: It doesn't matter that much. The type is never taken from the extension. In other tests the extension corresponds to the specified type.

src/ParseFile.js
RazvanCristian: I added validation to check if the string follows the scheme of a URI or if it is base64 encoded data.

src/ParseFile.js
```diff
@@ -137,26 +145,20 @@ class ParseFile {
...
-            type: specifiedType,
-          };
+        // Check if data URI or base64 string is valid
+        const validationRegex = new RegExp(base64Regex.source + '|' + dataUriRegex.source, 'i');
+        if (!validationRegex.test(data.base64)) {
+          throw new Error('Files passed in must have valid data URIs or base64 encoded data.');
         }
+
+        const base64 = data.base64.split(',').at(-1);
+        const type = specifiedType || data.base64.split(';').at(0).split(':').at(1) || '';
```
rubenatorio: According to RFC 2397 if there is no type shouldn't it default to `text/plain`?

```
If <mediatype> is omitted, it
   defaults to text/plain;charset=US-ASCII.  As a shorthand,
   "text/plain" can be omitted but the charset parameter supplied.
```

RazvanCristian: You're right, I've set the default type to `text/plain`.

mtrezza (changes requested):
Tests failing, please request another review from `@parse-community/js-sdk-review` when this is ready.

RazvanCristian:
> Tests failing, please request another review from `@parse-community/js-sdk-review` when this is ready.

Tests are failing because you're using node `14.19.1`, while the current LTS version is node `16.14.2`. 
image
`||=` operator was added in `15.0.0`. [Logical OR assignment (||=)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR_assignment)

mtrezza:
Can you adapt the code to pass? We won't change the node engine version at this point. Node 12 and 14 are still under LTS.

RazvanCristian:
Done.

mtrezza (changes requested):
Tests still fail, could you take a look? And please let me know if the new PR title is correct.

RazvanCristian:
> Tests still fail, could you take a look? And please let me know if the new PR title is correct.

Sorry for taking so long, I had some work to do. 
I updated the PR and checked that it passes the tests with version `v14.19.1`. It passes all the tests and there should be no other problems.

mtrezza:
I changed the PR title because the previous one was difficult to read. It seems you changed it back. What's incorrect or missing for the title:

> fix: creating a Parse.File with base64 string fails for some file types

src/ParseFile.js
```diff
@@ -137,26 +145,28 @@ class ParseFile {
...
-          this._data = base64;
-          this._source = {
-            format: 'base64',
-            base64: base64,
-            type: specifiedType,
-          };
+        // Check if data URI or base64 string is valid
+        const validationRegex = new RegExp(base64Regex.source + '|' + dataUriRegex.source, 'i');
+        if (!validationRegex.test(data.base64)) {
+          throw new Error('Files passed in must have valid data URIs or base64 encoded data.');
```
mtrezza: Could you rephrase the error message (if you think it makes sense) to:

> Cannot create a Parse.File without valid data URIs or base64 encoded data.

It also seems that this new error is missing a test case, could you add one?

mtrezza (changes requested):

mtrezza (approved):
Looks good to me! Thanks for this fix, let's just wait for the CI to pass.

mtrezza:
@rubenatorio Do you want to do a final review before we merge?

parseplatformorg:
🎉 This change has been released in version [3.4.3-alpha.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/3.4.3-alpha.1)

parseplatformorg:
🎉 This change has been released in version [3.4.3-beta.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/3.4.3-beta.1)

parseplatformorg:
🎉 This change has been released in version [3.4.3](https://github.com/parse-community/Parse-SDK-JS/releases/tag/3.4.3)

Saquib764:
It seems there is still issue with `Parse.File` while creating base64 file.

I am generating data URL from canvas-
```
let url = canvas.toDataURL("image/png");

const file = new Parse.File(`shading_texture.png`, { base64: url.split('base64,')[1] });
await file.save()
```

Upon running this script, the frontend freezes for a couple of minutes and then fails. 

The exact same script was working with version `3.4.2`

I also tried the following script to extract base64 string. No success.
```
let base64 = url.replace('data:', '').replace(/^.+,/, '')
```

mtrezza:
Could you open a new issue?

samateja:
this fix is not working when i upload a video and convert the video to base64. The file size is around 17mb, do you suggest any other solution?

FransGH:
See #1532

PR #1484, opened 2022-05-24
feat: Add two-way binding of Parse object properties with `Parse.Object.bind.key = value`

### LINKED ISSUE #1313: Add dot notation to get/set `Parse.Object` attributes
**Is your feature request related to a problem? Please describe.**
VueJS is pretty heavy on binding data using dot notation. This makes things pretty difficult when using Parse, as to get properties you need to use `.get(key)` and `.set(key)`.

This means that you have to constantly convert your data to and from JSON and Parse Objects.

**Describe the solution you'd like**
Be able to use dot notation for `.get(key)` and `.set(key)`. This would mean, for example, you could bind an input to obj.toJSON().name `<input v-model="obj.toJSON().name" placeholder="edit me">`, without having to have a function to pass it to another JSON object. 

**Describe alternatives you've considered**
Subclassing using Proxy, but doesn't work with vueJS

**Additional context**
-

dblythy:

### Issue Description

When using VueJS, it becomes difficult to manage Parse Objects via their native `.get` and `.set` syntax. Also, for newer developers who are use to standard JSON dot notation, the custom Parse Object getters and setters can be convoluted.

Related issue: #1313

### Approach

Adds `object.bind` option that allows for getting and setting of Parse Object properties via dot notation.

### TODOs before merging

- [x] Add tests
- [x] Add entry to changelog
- [x] Add changes to documentation (guides, repository pages, in-code descriptions)


dblythy:
For those unfamiliar with VueJS, you can bind object properties and they are automatically updated by their respective component.

```
<template>
  <div id="app">
    <div>
      <input type="text" placeholder="Name" v-model="object.name"/>
      <br>
      Name: {{object.name}}
      <br>
      <button @click="save()">Save</button>
    </div>
  </div>
</template>
<script>
import { Parse } from 'parse';
export default {
  name: 'App',
  data() {
    return {
      object: new Parse.Object('Monster'),
    }
  },
  methods: {
    async save() {
      await this.object.save();
    },
  },
}
</script>
```
https://user-images.githubusercontent.com/4974769/170022415-01188336-ff6e-473e-9820-c7d9fd84c1da.mov

dblythy:
As this feature matures we could perhaps ship it as a default with a new major version of Parse Server. I think it is a welcome improvement of usability to Parse. What do you think @mtrezza @Moumouls

mtrezza:
How matures if not released?

mtrezza:
This seems like a convenient feature, but I'm not sure this would work. How would you prevent collisions between a native JS object property and a Parse.Object property of the same name?

Related to this is https://github.com/parse-community/parse-server/issues/7130 where the discussion goes into the contrary direction to allow *any* field name, because a developer shouldn't have to worry about or be limited by collisions with Parse Server internal or JS native field names.

> It might be best to roll this out as experimental as it might have side effects with other Parse Objects.

We abandoned the concept of "experimental" features. If a feature is merged, it has the alpha / beta pre-release periods to mature, effectively 2 months minimum. If that's not enough because the feature is still too unstable or we haven't fully explored potential side effects, it won't be merged into the codebase for general stability and security reasons.

dblythy:
> This seems like a convenient feature

I think convenience and ease of use are important - especially for newer developers. There is a learning curve with Parse and this feature makes Parse objects behave intuitively instead of the convoluted `.get().get().get().get()` syntax.

> How would you prevent collisions between a native JS object property and a Parse.Object property of the same name

Can you give an example? I'm sure it would be easy enough to check if `key === Object.getOwnPropertyNames()` or something like that. The set/get behaviour of a `Proxy` can be completely customized.

Well, if it's not suitable for default usage, it would be good to have it as an option so Parse becomes a lot more usable with Vue. The only way to be assured that dot notation works for every single use case, would be to re-write every test with the additional dot notation syntax. I think for a new feature it's expected that it matures over time.

mtrezza:
> if it's not suitable for default usage, it would be good to have it as an option

As I said, we don't have the concept of experimental codebase changes at this time, even if part of the feature is a new Parse Server option. The feature would need to already have a certain level of maturity to enter the alpha branch, to a degree where it's reasonable to believe that it will become production ready within 1 month, before it moves to beta.

dblythy:
Well any initial release of software is expected to become more stable over time. All I’m saying is that there may be errors thrown specifically relating to the dot notation that might need refining / future PRs, which is expected of new features in an open source environment. 

I’m still not convinced there’s enough reason _not_ to consider this **optional** feature, especially as it dramatically increases usability with VueJS. If it’s a feature you wouldn’t use in your Parse environment, you don’t have to, but I don’t see why other developers should be restricted.

mtrezza:
Sure, if there are no open concerns this PR will be merged like any other. What I'm saying is that at this time we don't merge a PR if we do not believe in best conscience that it is production ready. I'm not saying this PR is or is not production ready, that's a different question.

I've opened https://github.com/parse-community/parse-server/issues/8016 to hopefully give you more insight into the challenges of "experimental" features.

dblythy:
Ahh, ok. Sorry we got a bit carried away. I'll continue to test this feature on my servers / sdks and let you know when I feel it's ready.

mtrezza:
Always good to have a healthy debate. I think if this is technically feasible, it would be a nice feature to have, not just for Vue developers. I will pin the issue to hopefully give it more exposure and have more eyes on it.

Just thinking, if you enabled the feature in a branch and did a regex replace for all tests from `.get(...)` to dot notation, it may give you an indication whether there are any obvious issues.

The reason I think this is a PR we have to look at very carefully is because it removes the containment of `Parse.Object` attributes. Currently they are stored in `Parse.Object.attributes` and `.get()` is merely a shortcut. This helps to clearly distinguish between attributes and any other JS object property. Once this is short-circuited, it creates ambiguity.

dblythy:
> if you enabled the feature in a branch and did a regex replace for all tests from .get(...) to dot notation, it may give you an indication whether there are any obvious issues.

That's a good idea, thanks.

> The reason I think this is a PR we have to look at very carefully is because it removes the containment of Parse.Object attributes.

Well I think this could be a documented side-effect of this feature. The same functionally happens in the Swift SDK without any troubles. Proxy objects merely intercept gets and sets too, no data is stored on the respective keys.

For example, console.log on `Object.keys(obj)` only returns `['id', '_localId', '_objCount', 'className']`, even when `object.name = 'foo'` is called (I just added a test for this).

mtrezza:
> Well I think this could be a documented side-effect of this feature.

What do you mean by "side-effect"? How should a developer interpret this? Are there any circumstances under which this change could break something?

mtrezza:
Here is an example that would break:

```js
const obj = new Parse.Object("Example");
await obj.save(null, { useMasterKey: true});
console.log(JSON.stringify(obj)); // "{"createdAt":"2022-06-02T21:35:24.430Z","updatedAt":"2022-06-02T21:35:24.430Z","objectId":"rnBNREYBUN"}"
obj.x = "x";
console.log(JSON.stringify(obj)); // "{"createdAt":"2022-06-02T21:35:24.430Z","updatedAt":"2022-06-02T21:35:24.430Z","objectId":"rnBNREYBUN"}"
console.log(obj.x); // "x"
```

I'm not sure whether the docs currently discourage from setting custom properties of a `Parse.Object`. The feature doc would need to highlight that.

dblythy:
> Here is an example that would break

If you were using that code _without_ dot notation, `obj.get('x')` would be undefined, but with the feature, it will be saved. So yes if existing code assigns properties directly, they will be set to the Parse Object.

With dot notation:

```js
const obj = new Parse.Object("Example");
await obj.save(null, { useMasterKey: true});
console.log(JSON.stringify(obj)); // {"createdAt":"2022-06-02T21:35:24.430Z","updatedAt":"2022-06-02T21:35:24.430Z","objectId":"rnBNREYBUN"}
obj.x = "x";
console.log(JSON.stringify(obj)); // {"createdAt":"2022-06-02T21:35:24.430Z","updatedAt":"2022-06-02T21:35:24.430Z","x":"x","objectId":"rnBNREYBUN"}
console.log(obj.x); // "x"
```

> The feature doc would need to highlight that.

Yes, that's a good point, and a valid reason why defaulting it to true is a bad idea for now.

> What do you mean by "side-effect"? How should a developer interpret this? Are there any circumstances under which this change could break something?

Well consider:

```js
class Example extends Parse.Object {
   constructor() {
      super();
      this.isSaved = false;
   }
}

```

If you set any values with a subclass using this[key], the proxy object will assume you want to save those keys. By default, the proxy checks if a function exists for the key (such as `obj.save`, or if the key is objectId, id, className, attributes, createdAt, updatedAt, or then, and if not, call `obj.get(key)`. It is also possible that another Parse internal subclass utilizes other internal keys (for example, Role might set this.name) that haven't been caught by the tests.

In the case of setting `this.isSaved`, it is recommended _not_ to use the proxy feature. A solution could be to allow the option to extend `internalKeys` on subclassed proxy object.

mtrezza:
It seems we are getting closer to the potential issues now. To merge this, we'd need to be sure enabling it doesn't break anything inside Parse Sever.

Also, whether native JS object properties are accessible. For example the `prototype` property. That's probably another limitation that needs to be documented?

The fact that enabling the option may break custom code when setting a [custom property](https://github.com/parse-community/Parse-SDK-JS/pull/1484#issuecomment-1145368395) would be considered a characteristic of the feature (a trade-off for the convenience of dot notation) that just needs to be well documented but wouldn't hinder merging.

dblythy:
> To merge this, we'd need to be sure enabling it doesn't break anything inside Parse Server.

I actually manually deployed the complied lib folder to my Parse Server, and there we a few changes I had to make for it to be compatible (I had to add "then" to the internal keys).

> Also, whether native JS object properties are accessible.

Yes, that's a good point. Most native JS properties are functions so they are returned anyway.

One thing i'm thinking about is prototype pollution. Do you think this feature is a risk if an external dependency calls `Parse.Object.__proto__.foo` for example? I think because the proxy only runs `.set` when the `obj[]` is called, it's probably safe.

I guess if a bad actor can pollute a Parse Object they can just as easily pollute `attributes`.

 (sorry not overly familiar with prototype pollution)

mtrezza:
> I had to add "then" to the internal keys

Is there a global list we can use instead of this local internal keys list? Local lists need to be maintained separately and increase code path complexity, that all makes it difficult to track down bugs.

Or even better, is there any way to avoid an internal keys list? If I read the current logic correctly, you are checking for excluding conditions:

```
if (typeof value === 'function' || key.toString().charAt(0) === '_' ||  _internalFields.includes(key.toString())
```

Could you instead check for including conditions:

```
if (Object.keys(this.attributes).includes(key))
```

So you only apply the new logic if the requested property is actually an existing attribute.

I also think this is missing collision handling. It may be possible to somehow set an object property with the same name as an attribute, so `set` and `get` should throw an error if there is an attribute *and* an obj property. This way the developer can know when there is a conflict, because the dot notation logic doesn't know which of the two to return. 

> One thing i'm thinking about is prototype pollution.

That's one of the major JS attack vectors, which is why I'd like to examine this PR closely. I'm not sure whether this changes the risk at all, but it may since it involves reflection.

dblythy:
> So you only apply the new logic if the requested property is actually an existing attribute.

I tried this, but as getting attributes calls `.attributes`, and then `.className`, `.id` etc internally (which triggers the proxy again), it ends up being an infinite loop.

> I also think this is missing collision handling.

The only thing with the collision handling is it's impossible to know whether a developer _or_ the internal SDK is setting those keys. Throwing on attempting to set these fields will break existing, internal SDK functionality. An example is with subclassing, you may very well want to override `destroy` for example. Having the conflict check prevents this from happening.

> That's one of the major JS attack vectors

I actually don't think it's a concern, as no values are set directly on the object. Logging `Object.keys(obj)` with dot notation on logs the same as if it was off.

mtrezza:
It seems for both points that the logic is missing something. The object internally should be able to distinguish a proxied property from an actual property. To the outside it can be opaque, but for the object itself this should be transparent.

dblythy:
Also noting that if we get `if (Object.keys(this.attributes).includes(key))` to work, setting a value to a new key will never work, e.g:

```
const obj = new Parse.Object('Proxy');
obj.name = 'name' // this won't work because 'name' isn't part of attributes yet
```

mtrezza:
@dblythy have you been using this feature since you developed it? do you have any experiences you can share?

dblythy:
I have primarily used this code for frontend use via subclassing Parse Object. Have used minimally with cloud code. Only issue I have found is that setting nested objects don’t work, as the proxy only works on the top layer (i.e `obj.field.key = true` won’t mutate `obj.field`)

SebC99:
Very naive question: for Vue.js why don’t you simply use `object.attributes.name`? 
It is a simple getter without the `get("name")` and even if not as easy as `object.name` it is very similar.

dblythy:
`attributes` is fine for `getter` but has no option for `set` (which is the utility this PR provides)

SebC99:
True, the object is frozen.
Thanks for the explanation :)

mtrezza:
> Only issue I have found is that setting nested objects don’t work, as the proxy only works on the top layer (i.e obj.field.key = true won’t mutate obj.field)

Is that easy to address, or a complex obstacle?

@SebC99 If you are trying this feature out, it would be great to hear your feedback.

dblythy:
I think it would be complex as it would have to be a proxy tree that ultimately call .set on the root. 

There are some [other](https://www.digitalocean.com/community/tutorials/js-proxy-traps) proxy traps we can use as well, I think a `delete` trap could be good, so `delete obj.key` calls `obj.unset("key")`

dblythy:
I have added a `delete` trap so when `delete object.key` is called it will call `object.unset("key")`

dblythy:
Latest commit adds deep proxies, so that `obj.field.key = true` _will_ mutate `obj.field`. Also adds support for `revert`

dblythy:
- `dirtyKeys.bind` helps vue/other reactive frameworks know when `dirtyKeys` have been changed.
- `_proxy_op` is used to re-trigger the `GET` trap, as the object's values have changed. This happens when `revert` or `fetch()` is called

This helps vue and other reactive frameworks keep the UI up to the Parse Object's state when using Dot Notation.

Consider the saved Parse Object with the data: `{name: 'name', price: 22, eft: true}`.

With this reactivity you can:

```
<div>
  <input v-model="object.name"/>
  <input v-model="object.price"/>
  <toggle v-model="object.eft"/>
  <button label="Save" @click="object.save()"/>
  <button v-if="object.dirtyKeys().length" label="Revert" @click="object.revert()"/>
</div>
```

With dot notation, this all functions without needing any additional code, efficiently updating the Parse Object as expected:

https://user-images.githubusercontent.com/4974769/195522977-3245a364-195a-47e6-9b4c-5c45cb1a1a3d.mov

dblythy:
Here is another example. This Parse Object has an array of `permissions`, which can only be edited via a cloud function, which adds another user onto the array.

```
<div v-for="user of obejct.permissions" :key="user.email">
  <div> {{ user.firstName }} </div>
  <div> {{ user.lastName }} </div>
  <div> {{ user.permissions }} </div>
</div>
```

Cloud function to create new user:

```
await Parse.Cloud.run('createUser', {firstName: '', lastName: ''});
await object.fetch(); // new user automatically is shown on UI thanks to reactivity
```

https://user-images.githubusercontent.com/4974769/195524170-6dd0cc93-03f1-4a66-83e9-83e9462e51a4.mov

dblythy:
Hope that helps explain why I'm so excited for this feature as a Vue developer 😃

mtrezza:
Exciting feature! 

As a requirement to release this we need two conditions to be fulfilled:
- The feature needs to be completely optional and the code related to that feature needs to be separated from other code to reduce the risk of any impact; i.e. the code changes of existing execution flows should be as minimal as possible. So that if the feature is disabled, the existing code paths are followed. That is specific to this feature, since it touches on fundamentals  and deals with proxies and reflection.
- The feature needs to be mature; we don't currently have "experimental features", every released feature needs to be fully functional and tested. If there is doubt about the stability of the feature, the impact on existing functionality, open questions or the feature is still under active development, these things needs to be addressed before a release. It would be nice to hear some feedback from people using this feature, maybe we can post instructions on how to test this branch out and make a Twitter post about it?

dblythy:
I had only suggested this as experimental as the only way to know it works across _everything_ is to refactor all the tests to use dot notation. I’ve been using this feature in production for about 3 months and it’s been great. I haven’t used it in cloud code though as Parse Server is shipped with its own version of the JS SDK

mtrezza:
>  the only way to know it works across everything is to refactor all the tests to use for notation

Sounds like a lot of work; and duplicating many tests probably will make them hard to maintain; maybe we find a hack to run the tests with dot notation by overriding the non-dot notation to use the dot notation instead (kind of inverse-proxying of the dot notation), or we only add a critical subset of tests and if someone reports a bug, we'll add tests as we go?

We could think about adding experimental features, but we need to develop a policy for that, because of how experimental features were used in the past.

dblythy:
> we only add a critical subset of tests and if someone reports a bug, we'll add tests as we go

that’s what I was hoping, that bugs that people discover with it will be collaborated and evolved.

So far I have used this across 3 production projects with 4-8 Parse Classes. Uses signup, save, find, first, fetch, pointers, etc. Works great with the only problems coming down to reactivity, which isn’t a major issue. Works as expected with cloud code too, payloads are the same with and without the feature.

mtrezza:
There will always be bugs, we just need to have a high confidence that there aren't any bugs in the first place. We currently believe of every feature that we merge that it's production ready to the best of our knowledge.

You've been doing a great bit of testing already, I would just like to hear some feedback from others as well, because there is usually a high probability that they will discover something specific to their use case. Should we do a tweet and invite people to test it?

dblythy:
The proxied objects still function as Parse Objects, the bugs that i would assume might be issues with reactivity around operations for example. I don’t expect the fundamentals of the feature (getting and setting) to have any issues. 

> Should we do a tweet and invite people to test it

that sounds good!

SebC99:
Sorry I’m not more involved in Parse development, I can’t find the time :(
As far as I’m concerned we do use a lot of cloud code so I guess I could try to use it in cloud code to see if it works well. 
But I suspect it might not: what about “temporary” properties? In both iOS and cloudCode we sometimes use property to save a state property or something similar (like a score) and it has no impact on Parse for now as we add it as simple properties to the object (and not to the attributes one). So these properties are not saved in DB, and are not part of the schema.
Would that be still possible? It might be a little hacky but I guess we’re not the only one doing so…
Just my 2 cents 😅

dblythy:
@SebC99 in your case you would not use the feature - it's going to be optional and default to `false`. Personally I don't think it has much value in cloud code (other than minor readability), it's strength more or less comes in when working with Vue.

mtrezza:
> and it has no impact on Parse for now as we add it as simple properties to the object (and not to the attributes one). So these properties are not saved in DB, and are not part of the schema.

What if we add a dedicated container property that can be excluded from this dot notation mechanism?

For example:
```obj.ephemeralAttributes.myTempVar = 1;```

Whatever is set in the `ephemeralAttributes` object can explicitly be ignored by Parse.

That may also be helpful for developers because it minimizes the risk of something breaking. If a developer today uses `Parse.Object.myCustomProperty` and the Parse SDK in the future added a feature that uses the same property it would create a conflict. So we give developers a container which we promise not to touch.

We already have `Parse.Object.attributes`, I assume that will keep existing to minimize breaking changes for users who enable the dot notation feature but are already manipulating  `Parse.Object.attributes` in some way?

**Comment by dblythy** [comment]
That’s a good idea - the other option is developers can simply prefix the key with `_` and it won’t be saved, so `object._myTempVar = 1` would not save to the server

**Comment by mtrezza** [comment]
That sounds even easier, but I'm afraid it may be too ambiguous for some. Underscore is often used to denote "private variables", since JS doesn't have a concept for that. If we use underscore for something else, we'd need to document that for developers and we'd probably have a hard time ensuring that all reviewers enforce that.

**Comment by mtrezza** [comment]
> the other option is developers can simply prefix the key with _ and it won’t be saved, so object._myTempVar = 1 would not save to the server

Is that an existing feature of this dot notation PR?

**Comment by mtrezza** [comment]
@dblythy Could you resolve the conflicts? It would be great if we could merge this, but in order for others to test this out and provide feedback we'd need to make it up-to-date. I think we could also write a Twitter post to vue-developers about this PR so we can get more feedback.

**Comment by dblythy** [comment]
I have created a sample project for examples for how to use this feature:

https://github.com/dblythy/Parse-Vue-Dot-Notation

I'm pretty confident with this feature as I've been using it in production for a while now, but any additional feedback / testing is welcome.

**Comment by dblythy** [comment]
> Is that an existing feature of this dot notation PR?

Yes, it wasn't initially intended for this, but the proxy won't apply if the key is prefixed by `_`, as a lot of Parse's internal keys are prefixes with `_` (`_localId`, `__type`, etc)

**Comment by mtrezza** [comment]
Great to hear that you've been using and testing this already thoroughly. 

- Currently we have attributes containerized, meaning one can use `Parse.Object.attributes` to access the properties and `Parse.Object.set / .get` to explicitly set/get them.
- Could adding properties on the root level of `Parse.Object` limit / complicate future development of the SDK due to naming conflicts? For example:
  - What happens if the SDK adds a property to `Parse.Object` in the future and the developer assigns a custom object property of the same name?
  - What happens if the developer adds a field `attributes` to the object class, would that interfere with the already existing `Parse.Object.attributes`?
- What happens if a developer adds a custom property `save` to an object class? Could there be any issues with the existing method `Parse.Object.save()`? At least it looks confusing to have both on the root level. 
- Could the existing `Parse.Object.attributes` be adapted in a way that it can be used the same way as the root notation in this feature?
- Why do these properties need to be on the root level of `Parse.Object`, could they be also containerized in like `Parse.Object.dotAttributes` to avoid possible naming conflicts?
- We should have a good reason to introduce rules and mechanisms with not-so-obvious background effects on properties with underscore. They can be difficult to consider for developers, lead to obscure bugs and make it difficult for reviewers of PRs. It sounds problematic to instruct developers to use an underscore prefix if the SDK uses the same syntax for internal properties. I think the instructions should actually be to not set any custom property on a `Parse.Object` because it may conflict with future properties and methods the SDK introduces. I think the advice for @SebC99 would be to not set any properties but instead use a parent object that hold the Parse object and any other custom properties. And we should probably add that advice to the docs.

**Comment by dblythy** [comment]
I think your suggestions highlight why this feature should be opt in, so developers can be aware of any caveats. 

If an existing internal `Parse Object` method is named via dot notation (such as `obj.save = true`), nothing will happen. It will need to be a written caveat that keys assigned via get/set using dot notation cannot clash with the existing `Parse Object` methods. 

If new internal methods are added in the future, they will take precedent over any user defined dot notation fields.

The `attributes` suggestion could work fine, but would lead to additional code when using nested pointers. Rather than being able to assign `obj.foo.bar` you would have to do `obj.attributes.foo.attributes.bar`, which is extra complexity. The point is that if opts in to using dot notation for their Vue project, they can access properties conveniently and quickly. 

We could easily add a log to the proxy trap for if internal keys are set, with “could not set ‘save' on object as it is an existing internal method”

**Comment by mtrezza** [comment]
It looks as if we had 2 options:
- a) Make this feature optional with a ton of caveats
- b) Address the caveats (maybe with some compromise on the part of convenience for vue developers) and make the feature non-optional

The challenge I see with (a) is that over time developers will likely ask to get the caveats addressed anyway, if this feature becomes popular. The caveats are also counteracting the convenience this feature intends to provide.

In addition, optional features with such a profound impact introduce an new "virtual" scope in the SDK with a very low test coverage. To fully understand the impacts of this feature we'd need to run all CI tests with the feature off and with the feature on. Otherwise this becomes an "experimental feature" which we currently to do not support.

So a benefit of (b) is that there is less risk and better test coverage for vue developers.

I'm not a vue developer so I would like to better understand the level of pain this feature tries to address.

Did you mean that a vue developer currently could use the `Parse.Object.attributes` property but the only pain is the word `attributes` that has to be added (repeatedly for pointers)? If so, would it help to simply add a short alias like `Parse.Object.a`? Or maybe add an option that makes it unnecessary to repeat the `attributes` property for pointers?

**Comment by dblythy** [comment]
I will reply tomorrow, if any other vue developers would like to add thoughts, it’s very welcomed! 

It’s also worth mentioning that the naming of keys potentially clashing caveat also applies to the Swift SDK and Android KTX extension. I’m also wondering if this feature will make typescript support easier as well.

**Comment by dblythy** [comment]
Well I think the dotNotation feature should be optional anyway as certain rebinding of methods are designed to trigger reactivity. 

The “virtual coverage” issue isn’t as big of an issue in my opinion - the proxy handler only intercepts key get and sets and should respect existing functionality. The unknown side effects are whether deep in the JS or Server code some undocumented properties are being assigned to Parse.Object, or that certain existing methods or properties might not function _reactively_ as they should. 

With Vue, you can do 2 way data binding using `v-model`. When assigned to a `input` for example, the UI will automatically update whenever the input is updated. For example:

```vue
<template>
   <div>
    <input v-model="text"/>
   </div>
</template>
<script>
export default {
  data() {
    return {
      text: ""
    }
  }
}
</script>
```

What this means is you can create reactive UI with very little code, such as:

```vue
<template>
   <div>
     {{ text }}
    <input v-model="text"/>
     <div v-if="text.length < 3">
       text must be longer than 3
     </div>
   </div>
</template>
<script>
export default {
  data() {
    return {
      text: ""
    }
  }
}
</script>
```

This is similar to React’s getState and setState methods used for two way data binding and reactive UI, however as you can see no `onChange="text => setState({text})"` is needed. UI can be updated automatically just by calling `this.text = "newStr"`. 

Now this gets problematic when using Parse Objects, as the `.get` syntax cannot be passed to a `v-model`. The alternative is:

```vue
<template>
   <div>
     {{ object.name }}
    <input v-model="object.name"/>
    <button @click="save">Save</button>
   </div>
</template>
<script>
export default {
  methods: {
    async save() {
      const parseObject = new Parse.Object("test")
      parseObject.set(this.object); // this spreads all of this.objects keys
      await parseObject.save()
      this.object = parseObject.toJSON()
    }
  },
  data() {
    return {
      object: {
        name: ""
      }
    }
  }
}
</script>
```

This makes it hard to track dirty keys, and you have to create and assign a POJO every time. 

There’s a few solutions:

Subclassing Parse Objects with get and set methods for your keys, such as: 

```vue
<template>
   <div>
     {{ object.name }}
    <input v-model="object.name"/>
    <button @click="object.save()">Save</button>
  </div>
</template>
<script>
class test extends Parse.Object {
  get name() {
    return this.get("name")
  }
  set name(val) {
    this.set("name", val)
  }
}
export default {
  data() {
    return {
      object: new test()
    }
  }
}
</script>
```

Or, as you mentioned, making attributes setable, such as:

```vue
<template>
   <div>
     {{ object.attributes.name }}
    <input v-model="object.attributes.name"/>
    <button @click="object.save()">Save</button>
  </div>
</template>
<script>
export default {
  data() {
    return {
      object: new Parse.Object("test")
    }
  }
}
</script>
```

With this PR, this becomes:

```vue
<template>
   <div>
     {{ object.name }}
    <input v-model="object.name"/>
    <button @click="object.save()">Save</button>
  </div>
</template>
<script>
Parse.dotNotation = true;
export default {
  data() {
    return {
      object: new Parse.Object("test")
    }
  }
}
</script>
```

As previously mentioned, other SDKs assign properties directly:

- Swift: `object.key = true`
- KTX: `object.key = true`

Perhaps as a middle ground, we could adapt this PR to:

1. Return a proxy trap for `object.attributes` that can be assigned (need to consider whether this has any impacts on code such as `const json = object.attributes` as this will return a `Proxy` rather than a POJO
2. Allow `dotNotation` as an optional feature targeted for two way data binding for reactive frameworks. Includes additional methods to trigger UI binding. 

This way, if we introduce a new method that clashes with an existing object, a developer can still use `object.attributes.save` whilst the majority of other use cases can still use `object.key`. 

Feel free to have a look / play with the [example project](https://github.com/dblythy/Parse-Vue-Dot-Notation/blob/main/src/components/Parse-Object.vue)

mtrezza:
Thanks for the examples.

> It’s also worth mentioning that the naming of keys potentially clashing caveat also applies to the Swift SDK and Android KTX extension. 

The LLVM / JVM compilers allow for signature overloading and can differentiate the signatures in a compile step, so I think that's not a risk. JS just runs the interpreter and is missing that capability, so I'm not sure these are easily comparable. The Parse Swift SDK was developed without strategic supervision - unfortunately, as we see today - so there may be differences in syntax and usage from the rest of the Parse product suite.

We don't have experimental features, so the reason for making a feature optional cannot be because of potential conflicts. The feature must work either way without complicating the work of developers or reviewers by introducing a vague risk for conflicts.

For JS, I think the only way to avoid conflicts is to containerize these proxies instead of having them on the root.


We already have `Parse.Object.attributes` and allowing to access that for vue seems like the logical solution, considering:

- Maybe we can optimize this so to avoid `Parse.Object.attributes.pointerA.attributes.pointerB.field` and instead only use `Parse.Object.attributes.pointerA.pointerB.field`

- If we still cannot address potential conflicts then maybe we need a distinct container like `Parse.Object.vue` that contains the proxies.

dblythy:
Closing as it’s clear this PR isnt considered suitable for the JS SDK, if you’re a Vue developer and this feature helps you, you can enable it via subclassing `Parse.Object`.

dblythy:
Ok so in summary this PR adds:

2 way data binding for the currently frozen `object.attributes`
2 way data binding for `object.key` using Parse.dotNotation

Caveats:

1. If obj properties clash with the same name as public SDK properties or methods, the SDK will take precedence. This can be worked around using `object.attributes.key` instead of `object.key`
2. If in the future the SDK adds a new method or public property that creates a name conflict with their custom object properties, their app will break without prior notice, so we recommend a high test coverage as even a patch or feature release may be a breaking change for them.

mtrezza:
Thanks for reopening. As we've continued the discussion in chat, here is a list of the full effects of this PR:

> 2 caveats for vue developers:
> - They cannot use custom obj properties with the same name as public SDK properties or methods as the SDK will take precedence
> - If in the future the SDK adds a new method or public property that creates a name conflict with their custom object properties, their app will break without prior notice, so we recommend a high test coverage as even a patch or feature release may be a breaking change for them
>
>Change for SDK reviewers:
> 
> - New rule that a reviewer must check in every PR that adds or modifies a property or method name that these definition lists are updated as well; if we forget that it may break the SDK for vue developers

The only reason for these effects is the aesthetic decision of having the properties on the root like `obj.field` instead of a parent property like `obj.vue.field` or the existing `obj.attributes.field`. There is no technical reason.

Requirements for merging this PR:
- At no point can there be any logic from this feature active unless the option is enabled. The code execution paths need to be fully separated.
- Changes in contributor and reviewer instructions in the CONTRIBUTION guide regarding the additional tasks this change introduces for them (for Parse Server PRs as well?).
- Change in PR template with a new task checkbox for updating the internal list of properties if there have been any additions or renaming.

Are you sure the benefits outweigh the cost?

mtrezza:
> If obj properties clash with the same name as public SDK properties or methods, the SDK will take precedence. This can be worked around using object.attributes.key instead of object.key

I wasn't aware you want to keep both. So a vue developer has the choice of not fully knowing if their app crashes with every SDK upgrade or being safe by using `object.attributes.key`. Is that a realistic choice for vue developers - aesthetics over stability?

dblythy:
I think so, but I understand your concern. 

There’s no real difference between adding a key `vue` and the existing `attributes`, although attributes is writable now. 

Another option is the proxy _always_ returns the alias, so if the SDK introduces a clash, existing apps wouldn’t break. However, this would mean that certain methods might become unavailable, for example if a Parse.Object has the data `{save: true}`, `object.save()` would throw. We could bypass this in the proxy by adding an alias to the objects raw properties, such as `obj._raw.save()`.  This would protect the feature against future new Parse Object prototype methods. This approach could be "safer" for the long term stability of the feature.

In this case the caveat would be:

> Using dotNotation, an object’s fields can potentially clash and override its own prototype functions. In this case, the prototype function can be accessed via `obj._raw`, which will return the original object without the proxy trap. 

> I wasn't aware you want to keep both.

I don’t see any downside to allowing writing to attributes, and I think it can exist as a seperate feature to `Parse.dotNotation`

mtrezza:
We discussed this in our chat already - that's not an option. The SDK would likely break internally when it calls itself in certain scenarios, it would be very difficult to control this risk.

The risk can only lie with the vue developer in a way they can control. That means it's their responsibility to manually check the Parse SDK source code before every Parse SDK upgrade for compatibility with their custom field names. That's caveat #1.

If a vue developer can avoid caveat #1 by using `obj.attributes.field` then that's great. I'd just be surprised if a developer accepts all the trouble that comes with caveat #1 just for aesthetics. But again, I'm not a vue developer so I follow your lead on this.

SebC99:
As a "product manager", I can't help thinking this is not a feature that is asked a lot, there's not a lot of complain about it, and it's almost a custom development for a specific usage (which can be overcome by the developer with specific developments)
So my question is: is it the role of the SDK to manage this?
It's complicating the whole SDK, the documentation (as we need to explain the option), and of course it can leads to bug. Just for a specific usage.

dblythy:
> Custom development for a specific usage 

That is almost how all the new Parse features come to be - specific usage cases that are provided back to the community for others to use.

As much as I like the aesthetic, I don't think it's ideal that a feature restricts what _future_ properties can be added to `Parse.Object`, as well as creating anxiety/troubles for upgrading to newer SDKs which may contain security releases.

As a compromise, this feature has been refactored to live on the `bind` property. This should remove the caveats mentioned and allow development of the SDK to continue without having to consider `dotNotation`

So, if you are using Vue, you can create responsive UIs by using `obj.get("field")`, but if you need to assign an attribute to be 2-way bound, you can use `.bind`, such as:

```vue
<template>
  <div>
    <div>
      <h1>FOO: {{ object.get("foo") }}</h1>
      <input v-model="object.bind.foo" />
    </div>

    <div>
      <label for="checkbox">Checked: {{ object.get("checked") }}</label>
      <input type="checkbox" id="checkbox" v-model="object.bind.checked" />
    </div>

    <div>
      <div>Checked names: {{ object.get("names") }}</div>

      <input
        type="checkbox"
        id="jack"
        value="Jack"
        v-model="object.bind.names"
      />
      <label for="jack">Jack</label>

      <input
        type="checkbox"
        id="john"
        value="John"
        v-model="object.bind.names"
      />
      <label for="john">John</label>

      <input
        type="checkbox"
        id="mike"
        value="Mike"
        v-model="object.bind.names"
      />
      <label for="mike">Mike</label>
    </div>

    <div>
      <div>Picked: {{ object.get("picked") }}</div>

      <input type="radio" id="one" value="One" v-model="object.bind.picked" />
      <label for="one">One</label>

      <input type="radio" id="two" value="Two" v-model="object.bind.picked" />
      <label for="two">Two</label>
    </div>

    <div>
      <div>Selected: {{ object.get("selected") }}</div>

      <select v-model="object.bind.selected">
        <option disabled value="">Please select one</option>
        <option>A</option>
        <option>B</option>
        <option>C</option>
      </select>
    </div>

    <div>
      <button @click="object.save()">Save</button>

      <div>{{ JSON.stringify(object.toJSON(), 2) }}</div>
    </div>
  </div>
</template>

<script>
export default {
  name: "Parse-Object",
  data() {
    return {
      object: new this.$Parse.Object("Test", { names: [] }),
    };
  },
};
</script>
```

This provides the following responsive UI, directly bound to a `Parse.Object`:


https://user-images.githubusercontent.com/4974769/222996996-0c853bc8-3efd-4fd2-a2a1-b717e733ae17.mov

This approach allows safe future support for Vue/Svelte developers, and no conflicts with other features.

mtrezza:
Thanks for jumping in and voicing your concern @SebC99; I think with the latest suggestion of a `bind` syntax we have come to a good compromise that should make life easier for vue developers and avoid the previous caveats. Thanks @dblythy for all the revisions. Once this is merged, I think this new feature merits a Tweet to notify the vue developers in our community.

mtrezza:
Maybe the changelog entry could mention some keywords like "vue"? How about:

> feat: Add two-way binding of Parse object properties with `Parse.Object.bind.key = value` for use with Vue.js

dblythy:
Let me have a quick look to see if applies to Svelte too

dblythy:
https://user-images.githubusercontent.com/4974769/223304704-e45fc35a-2eba-4f74-bf07-ac3e2771f68a.mov

Feature is fully compatible with Svelte as well, with the only caveat being that to trigger reactivity on save, `object = await object.save()` must be called, whereas in Vue, `await object.save()` is enough.

Svelte demo: https://github.com/dblythy/Parse-Svelte-Binding

And for good measure, AngularJS demo: https://github.com/dblythy/Parse-Angular-Binding

mtrezza:
Can the docs all be packed into the `bind` code comments? Or should we add a new section for Reactive Binding to the html docs?

dblythy:
I have added some JSDocs, but it should be part of the JS Guide as well

mtrezza (approved):
Looks good, is this ready for merge?

dblythy:
Did you want to modify the docs at all? I didn't make too much effort with it as I figured you were likely to amend

mtrezza:
Maybe this is something we could incorporate into the JSDoc:

> Feature is fully compatible with Svelte as well, with the only caveat being that to trigger reactivity on save, object = await object.save() must be called, whereas in Vue, await object.save() is enough.

Could you formulate that in more generic terms?

Should we add code example snippets too for 2 way binding with vue, svelte and angular since you mention they use a different syntax?

dblythy:
> Should we add code example snippets too for 2 way binding with vue, svelte and angular since you mention they use a different syntax?

Perhaps we can add this detail in the expanded Javascript Guide, rather than the JS Docs?

mtrezza:
I'd only add to the docs what cannot go into JSDocs for some reason. I think keeping docs as close to the code as possible makes most sense, because:

- Easier accessible to the developer in an IDE, right when and where they need it.
- Easier to maintain than a separate document or even repository (as we know from the history of our docs).
- Tends to encourage more concise docs rather that the often too verbose docs that we currently have.

A separate documentation may be more for in-depth explanations where necessary or more tutorial-style introductions. I personally believe that's what we should aim for, but we haven't introduced a definitive policy on that yet. Apple's own SDK docs are a good example.

So depending on what you have in mind and how extensive of an introduction you want to write, you can decide for one or the other. However, we definitely want to avoid duplicate docs in both places.

src/ParseObject.js
```diff
@@ -148,6 +150,17 @@ class ParseObject {
   _objCount: number;
   className: string;
 
+  /**
+   * Bind, used for two way directonal binding using
```
mtrezza, line 154: Needs completion

mtrezza (changes requested):

dblythy:
Working on improving Array support

mtrezza:
@dblythy Couod you amend the [docs](https://github.com/parse-community/Parse-SDK-JS/pull/1484#issuecomment-1461947793)? Then let's go ahead and merge this I'd say. Unless you found a weak spot in the meantime that should be fixed first.

bpbastos:
Hello, do you have any news about this feature being released?

mtrezza:
It seems that only the docs could benefit from some of the comments mentioned here, see https://github.com/parse-community/Parse-SDK-JS/pull/1484#issuecomment-1461947793; then this could be merged.

bpbastos:
> It seems that only the docs could benefit from some of the comments mentioned here, see [#1484 (comment)](https://github.com/parse-community/Parse-SDK-JS/pull/1484#issuecomment-1461947793); then this could be merged.

Great! I'm using the proxyHandler hack, but it seems to be causing side effects. I'm still new to Parse and Vue, and this feature will greatly ease the learning process.

PR #1540, opened 2022-09-04
feat: generate `Parse.Object.objectId` automatically when `allowCustomObjectId` is enabled and no `objectId` is passed

### LINKED ISSUE #1402: Option `allowCustomObjectId` forces all objects to have custom `objectId`

- [ x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x ] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x ] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-JS/issues?q=is%3Aissue).

### Current Limitation

Currently it is not possible to have mixed environment of custom and server generated `objectId` with `allowCustomObjectId = true`. At certain database schemes it is desired that some objects have their `objectId` generated by server and some (mainly in other object class) would get a custom `objectId`. 

- Improve documentation stating clearly that if `allowCustomObjectId = true` the `createdAt` is used to decide whether the object is being created or updated. So that developers know how the create/update is handled
- Remove code blocking to create an object where `allowCustomObjectId = true` fore the `objectId` to not be nil, for example: https://github.com/parse-community/Parse-SDK-JS/blob/d196e7f4f861aa76822c59ccb78352fe0bdda80c/src/ParseObject.js#L2402

### Example Use Case

As an example - in my use case there is always one *private* `User` object and always one *public* `Profile` object. Having the ability to set custom `objectId` to `Profile` that is same as the related `User.objectId` would give a direct reference between these two objects, saving following load:
- back-end fetch and device sync when creating account (2x save reference between `User` and `Profile`)
- improve the client code readability as the client code always need to pass around two different objectId for one user, instead of one "device user" objectId
- cloud code triggers have to do extra fetch of the `Profile` or `User` object to find related partner (for example to set ACL when there is only profile's `objectId` known)
- extra field at both Profile or User to link them and for some scenarios also indexing

Currently used workaround is to use reference field in each linked objects, as mentioned above the `User.profileId` and `Profile.userId`, but when user A is saving a new message to user B and A knows only B's profiles `objectId`, the cloud code has to fetch the `Profile` object to be able to set ACL to the recipient `User.objectId`. With growing user base and larger message count this is not negligible overload

### 3rd Party References

- As another use  example can be the **mstniy**'s comments here: https://github.com/parse-community/Parse-SDK-JS/pull/1309
- ParseSwift SDK: https://github.com/parse-community/Parse-Swift/issues/220

### Additional observations
- As mentioned in JS discussion it force to always use a custom id and the `allowCustomObjectId` is a confusing name for that. I can confirm that - as a newcomer to Parse I understood that flag for a long time wrong, that I can use both, and learned the hard way the opposite. So either it should be renamed, or the SDKs updated (what I see as small change), plus added in the documentation that allowCustomObjectId will still generate objectId if it is not defined (together with mentioning that createdAt is needed when allowCustomObjectId if true.
- Testing with ParseSwift SDK I found that it works as intended [test comit](https://github.com/lsmilek1/Parse-Swift/commit/b3346df5d09dfabeeb9c49e682fb6b8e50fea2bc), although I am not sure if JS SDK is more critical as it is used by server directly.

Unfortunately I am very bad in javaScript and would most probably do more harm than profit but with some guidance I can try. Although to run test cases I will certainly fail as I am newcomer knowing only swift and running Parse with back4app... The documentation I can cover!

musthafa1996:

### Issue Description
Enabling `allowCustomObjectId` forces all objects to have custom `objectId`s instead of auto-generating when no `objectId` is passed, or when `objectId` is **null** or **undefined**. 

Related issue: https://github.com/parse-community/Parse-SDK-JS/issues/1402

### Approach
When `allowCustomObjectId` is enabled:
- Check if `objectId` has a custom value.
- If `objectId` has a value and is an empty string, throw 'objectId must not be empty' error.
- If `objectId` doesn't have a value, or is **null** or **undefined**, generate `objectId` automatically as if `allowCustomObjectId` were disabled.

### TODOs before merging

- [x ] Add tests
- [ ] Add entry to changelog


integration/test/ParseObjectTest.js
```diff
@@ -2063,18 +2063,32 @@ describe('Parse Object', () => {
...
-    const object = new Parse.Object('TestObject');
+
+    // Try to save without passing objectId
+    const object1 = new Parse.Object('TestObject');
     try {
-      await object.save();
+      await object1.save();
+    } catch (error) {
+      fail();
+    }
```
dblythy:
```suggestion
    await object1.save();
```

integration/test/ParseObjectTest.js
```diff
@@ -2092,17 +2106,23 @@ describe('Parse Object', () => {
...
+
+    // Try to save empty objectId
     const obj1 = new TestObject({ foo: 'bar' });
+    obj1.id = '';
     const obj2 = new TestObject({ foo: 'baz' });
+    obj2.id = '';
     try {
       await Parse.Object.saveAll([obj1, obj2]);
       fail();
     } catch (error) {
```
dblythy:
```suggestion
    await expectAsync(Parse.Object.saveAll([obj1, obj2])).toBeRejectedWith(new ParseError(ParseError.MISSING_OBJECT_ID, 'objectId must not be empty'))
```

dblythy (changes requested):
Looks mostly good, just some minor refactoring

musthafa1996:
Thanks for reviewing @dblythy! I have refactored the tests with your suggestions. Can you check again please?

musthafa1996:
Hi @dblythy! Were you able to have a look at the recent changes pushed to the PR?

dblythy:
You might need to run `npm run lint` and fix syntax errors. I couldn't suggest changes to be syntactically correct as I couldn't provide suggestions across deleted lines.

musthafa1996:
> You might need to run `npm run lint` and fix syntax errors. I couldn't suggest changes to be syntactically correct as I couldn't provide suggestions across deleted lines.

Thanks. I’ll run and post the results.

musthafa1996:
@dblythy, my bad. I have just fixed the syntax error and updated the PR. Please have a look now.

The following is the result of `npm run lint`:
Screenshot 2022-09-06 at 3 55 43 PM

The following is the result of `npm run test`:
Screenshot 2022-09-06 at 3 56 39 PM
Screenshot 2022-09-06 at 3 56 59 PM

integration/test/ParseObjectTest.js
```diff
@@ -2092,17 +2102,27 @@ describe('Parse Object', () => {
...
+      new Parse.Error(Parse.Error.MISSING_OBJECT_ID, 'objectId must not be empty')
+    );
+
     try {
       await Parse.Object.saveAll([obj1, obj2]);
       fail();
     } catch (error) {
-      expect(error.message).toBe('objectId must not be empty, null or undefined');
+      expect(error.message).toBe('objectId must not be empty');
     }
```
dblythy:
```suggestion
```

integration/test/ParseObjectTest.js
```diff
@@ -2063,18 +2063,28 @@ describe('Parse Object', () => {
...
+    const object2 = new Parse.Object('TestObject');
+    object2.id = '';
     try {
-      await object.save();
+      await object2.save();
       fail();
     } catch (error) {
-      expect(error.message).toBe('objectId must not be empty, null or undefined');
+      expect(error.message).toBe('objectId must not be empty');
     }
```
dblythy:
```suggestion
    await expectAsync(object2.save()).toBeRejectedWith(
      new Parse.Error(Parse.Error.MISSING_OBJECT_ID, 'objectId must not be empty')
    );

```

dblythy (changes requested):
Sorry, I just worked out how to properly suggest 😂. 

Almost there, @mtrezza can you allow the tests to run?

musthafa1996:
I had been careful enough this time while refactoring. Can you check yet again, @dblythy ? :)

musthafa1996:
@mtrezza, @dblythy, have you checked the latest changes?

dblythy:
I will review again within 24 hrs 😊

musthafa1996:
Thank you 😊

dblythy:
Can you check the integration check? `npm run integration`

musthafa1996:
> Can you check the integration check? `npm run integration`

Hi @dblythy and @mtrezza! I did get some time to look into this PR today. I have pushed the following updates:

- Remove unnecessary id existence check from `ParseObject` in case allowCustomObjectId is true.
- Fix tests for allowCustomObjectId.

Turns out that we don't need the following checks in `ParseObject` as they are already validated from ParseServer.

Screenshot 2022-09-10 at 2 09 40 AM

Screenshot 2022-09-10 at 2 09 55 AM

All the related tests seem to work fine now.

musthafa1996:
There is one thing to note here. The error received from ParseServer says 'objectId must not be empty, null or undefined'. Yes, object saving fails when the `objectId` is **empty** or **null**. However, saving succeeds when `objectId` is **undefined** with automatic objectId generated. 

We probably need to update the error message on ParseServer. Please advice.

dblythy:
> Turns out that we don't need the following checks in ParseObject as they are already validated from ParseServer

Even though the checks happen on Parse Server, the client side validation prevents unnecessary computation / network calls. There's no downside to have them run on the client side as well

musthafa1996:
Okay. Let me add them back in that case.

musthafa1996:
@dblythy, @mtrezza, I have added the validation back. All the test cases are also passed.

mtrezza:
Could you take a look at the failing CI? The lock file version needs to be 1, then the "ci build" will likely run.

integration/test/ParseObjectTest.js
```diff
@@ -2087,22 +2101,35 @@ describe('Parse Object', () => {
     const afterSave = await query.get(customId);
     expect(afterSave.get('foo')).toBe('baz');
     Parse.allowCustomObjectId = false;
+
+    done();
   });
 
-  it('allowCustomObjectId saveAll', async () => {
+  it('allowCustomObjectId saveAll', async done => {
```
dblythy: no need for done
```suggestion
  it('allowCustomObjectId saveAll', async () => {
```

integration/test/ParseObjectTest.js
```diff
@@ -2114,5 +2141,7 @@ describe('Parse Object', () => {
       expect([customId1, customId2].includes(result.id));
     });
     Parse.allowCustomObjectId = false;
+
+    done();
```
dblythy:
```suggestion
```

musthafa1996:
@mtrezza, I have fixed the **packages-lock.json** file. Can you run the build again?

integration/test/ParseObjectTest.js
```diff
@@ -2087,22 +2101,35 @@ describe('Parse Object', () => {
     const afterSave = await query.get(customId);
     expect(afterSave.get('foo')).toBe('baz');
     Parse.allowCustomObjectId = false;
+
+    done();
```
dblythy:
```suggestion
```

integration/test/ParseObjectTest.js
```diff
@@ -2060,21 +2060,35 @@ describe('Parse Object', () => {
     expect(obj.get('string')).toBeInstanceOf(String);
   });
 
-  it('allowCustomObjectId', async () => {
+  it('allowCustomObjectId', async done => {
```
dblythy:
```suggestion
  it('allowCustomObjectId', async () => {
```

dblythy (changes requested):
Almost there!

musthafa1996:
@dblythy, pushed the **ParseObjectTest.js** updates. Can you check again?

musthafa1996:
@dblythy, can you do a review?

dblythy:
It's mostly good - just need to remove the `done()` from the tests

musthafa1996:
@dblythy, the test starts failing when I remove `done()` from `allowCustomObjectId` test in **ParseObject-test.js**. Please check the screenshots below and share your thoughts.

With `done()`:
Screenshot 2022-09-16 at 9 16 24 PM

Without `done()`:
Screenshot 2022-09-16 at 9 17 58 PM

musthafa1996:
@dblythy any idea why?

dblythy:
when you remove `done()` you also must remove `async done => {`

So, remove `done()` but also change the function to `async () => {`

musthafa1996:
Thanks :) 
Let me try that. I'm new to testing by the way :D

dblythy:
No worries! It should work if you make the changes as per my [previous review](https://github.com/parse-community/Parse-SDK-JS/pull/1540#pullrequestreview-1103567948). Feel free to let us know if you need any more help!

musthafa1996:
@dblythy, this is what I get after removing `done()`. Lint doesn't identify `fail()` method at lines 3822 and 3828,
Screenshot 2022-09-19 at 12 23 32 PM

musthafa1996:
Thanks @dblythy. I have resolved the tests. Please review.

dblythy (approved):

musthafa1996:
What's pending for this PR to be merged now @dblythy, @mtrezza?

integration/test/ParseObjectTest.js
```diff
@@ -2063,18 +2063,32 @@ describe('Parse Object', () => {
   it('allowCustomObjectId', async () => {
     await reconfigureServer({ allowCustomObjectId: true });
     Parse.allowCustomObjectId = true;
+
+    // Try to save without passing objectId
```
mtrezza: If a test title doesn't tell exactly what it tests and it needs comments to explain, that's usually a sign that it should be split into individual tests. Combining multiple test scenarios in a single test increases code complexity and makes it more difficult to add changes in the future.

You could add a new `describe` layer for `allowCustomObjectId` and the add tests like:
```
describe('allowCustomObjectId', () => {
  it('fails to save with empty objectId', async () => {
```

Could you split this into individual tests?

src/ParseObject.js
```diff
@@ -2419,6 +2413,13 @@ const DefaultController = {
             const batch = [];
             const nextPending = [];
             pending.forEach(el => {
+              if (allowCustomObjectId && Object.prototype.hasOwnProperty.call(el, 'id') && !el.id) {
+                throw new ParseError(
+                  ParseError.MISSING_OBJECT_ID,
+                  'objectId must not be empty or null'
```
mtrezza, line 2419: This has changed from `'objectId must not be empty, null or undefined'` to ` 'objectId must not be empty or null'`. Why is `undefined` not allowed?

mtrezza (changes requested):

integration/test/ParseObjectTest.js
musthafa1996: Yes, sure. It was like that previously.

musthafa1996: It was like that previously.

musthafa1996: Hi @mtrezza! I have split the tests for allowCustomObjectId. Please review again and mark this request as resolved if all is good.

mtrezza:
Waiting for CI to pass

musthafa1996:
CI has been completed successfully.

mtrezza (approved):
Looks good!

mtrezza:
@dblythy Is this ready for merge also from your side?

musthafa1996:
@mtrezza, is there any actions required from me to get this merged?

dblythy (approved):

parseplatformorg:
🎉 This change has been released in version [3.5.0-alpha.4](https://github.com/parse-community/Parse-SDK-JS/releases/tag/3.5.0-alpha.4)

musthafa1996:
Thanks for merging @mtrezza and @dblythy :)

dblythy:
Thanks for your contribution!! 🔥

musthafa1996:
Hi @mtrezza! When can we expect version 3.5.0 to be released?

mtrezza:
The release cycle is monthly (around the beginning of every month): alpha -> beta -> release
This feature will be in beta in Okt. and in stable release in Nov.

musthafa1996:
Got it. Thanks for letting me know @mtrezza 😊

parseplatformorg:
🎉 This change has been released in version [3.5.0-beta.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/3.5.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [3.5.0](https://github.com/parse-community/Parse-SDK-JS/releases/tag/3.5.0)

PR #1578, opened 2022-10-21
fix: File upload fails when uploading base64 data

### LINKED ISSUE #1579: Issue uploading files
### New Issue Checklist
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description
Parse server throws the following error while uploading files:
`
The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined 
`

### Steps to reproduce
Using File API of Parse  SDK:
```
const file = new Parse.File('parse-server-logo', {
  base64: 'data:image/jpeg;base64,ParseA==',
});
await file.save();
```

Using File upload Rest API:
```
curl --location --request POST 'http://198.199.84.19:1337/parse/files/test.txt' \
--header 'X-Parse-Application-Id: <APP_ID>' \
--header 'X-Parse-REST-API-Key: <REST_API_KEY>' \
--header 'X-Parse-Master-Key: <MASTER_KEY>' \
--header 'Content-Type: application/text' \
--data-raw 'Hello world' 
```

### Actual Outcome
For File API of Parse SDK:
```
The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined
```

For File upload Rest API:
```
{
    "code": 130,
    "error": "The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined"
}
```

### Expected Outcome
Successfully upload the file and respond with the URL to the file. A sample success response would be:
```
{
    "url": "http://localhost:1337/parse/files/APP_ID/0fef055962401a0c5d23d9c10b737a07_test.txt",
    "name": "0fef055962401a0c5d23d9c10b737a07_test.txt"
}
```

### Environment
This issue occurs with Parse SDK version `^3.5.0-alpha.6`. Please note that the same version of Parse SDK must be used in Parse Server. Parse Server uses the latest released version of Parse SDK (3.4.2) by default. You will need to override the SDK version for Parse server. To do so, you can add the following in `package.json`:
```
"overrides": {
  "parse-server": {
    "parse": "^3.5.0-alpha.6"
  }
}
```

Also note that npm package overriding is only supported from npm version >=8.3.0.

Server
- Parse Server version: `5.2.8`, `5.3.0-alpha.30`
- Operating system: `macOS Monterey`, `Ubuntu 22.04.1 LTS`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `local`. `Digital Ocean`

Database
- System (MongoDB or Postgres): `Postgres`
- Database version: `PostgreSQL 14.1`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `local`, 'Digital Ocean'

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `Javascript`
- SDK version: `^3.5.0-alpha.6`

### Logs
`
error: Error creating a file:  The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined {"code":"ERR_INVALID_ARG_TYPE","stack":"TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined\n    at new NodeError (node:internal/errors:393:5)\n    at Function.from (node:buffer:328:9)\n    at createHandler (/api/node_modules/parse-server/src/Routers/FilesRouter.js:167:35)\n    at processTicksAndRejections (node:internal/process/task_queues:95:5)"}
error: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined {"code":130,"stack":"TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined\n    at new NodeError (node:internal/errors:393:5)\n    at Function.from (node:buffer:328:9)\n    at createHandler (/api/node_modules/parse-server/src/Routers/FilesRouter.js:167:35)\n    at processTicksAndRejections (node:internal/process/task_queues:95:5)"}
`

musthafa1996:

### Issue Description
Parse server throws the following error while uploading files:
`The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined`

Related issue: #1579, https://github.com/parse-community/parse-server/issues/8253
Closes #1579, https://github.com/parse-community/parse-server/issues/8253

### Approach
Add in the missing assignment of `_data` property with the uploaded file data in `ParseFile.js`

### TODOs before merging
- [x] Add tests
- [x] Remove overriding of Parse server in **package.json**
- [x] Rollback the content of **package-lock.json**
- [ ] Add entry to changelog
- [ ] Add changes to documentation (guides, repository pages, in-code descriptions)x


musthafa1996:
Hi @mtrezza! All the test cases are now passing. File uploading works now.
image
image

The overriding of Parse SDK using the git repo git@github.com:musthafa1996/Parse-SDK-JS.git#file-upload-issue for Parse server was necessary to make sure Parse Server uses the Parse SDK with the file upload fix. 

Once approved we’ll need to do the following:
- Remove overriding of Parse SDK for Parse Server
- Rollback the content of package-lock.json
- Make sure Parse Server uses the Parse SDK with file upload fix.

integration/test/ParseFileTest.js
```diff
@@ -114,4 +114,16 @@ describe('Parse.File', () => {
...
+    await file.save();
+    let data = await file.getData();
+    assert.equal(data, 'ParseA==');
+    file._data = null;
+    await file.save();
+    assert.equal(file._data, null);
+    data = await file.getData();
+    assert.equal(data, 'ParseA==');
+  });
 });
```
musthafa1996: I don’t think this test case would be necessary as all the test cases related to file upload were failing. This is just a redundant test case. Let me know your thoughts @mtrezza.

mtrezza: Do you mean the other tests were failing as well for Parse Server using the latest alpha version of the Parse JS SDK?

musthafa1996: Yes, all the existing test cases using file upload were failing with the latest alpha version of the Parse JS SDK. The below screenshot was the message where I had informed you about it previously.
Screenshot 2022-10-28 at 10 01 54 AM

musthafa1996: Do you think we should be removing that test case?

mtrezza: If a test for the same functionality already exists, you can remove it; if no other test does what your test does, we can leave it.

If we keep the test, I think it needs a rewrite, because you are setting `file._data = null;`, but that is an internal property and I'm not sure there is an expected behavior we can use in the test. You would probably want to download the file into a new object.

```
const file = new Parse.File('parse-server-logo', { base64: 'ParseA==' });
await file.save();
let data = await file.getData();
assert.equal(data, 'ParseA==');
file._data = null;
await file.save();
assert.equal(file._data, null);
data = await file.getData();
assert.equal(data, 'ParseA==');
```

musthafa1996: There are other tests with the same functionality. Let me remove it.

mtrezza: ok

musthafa1996: I have removed this test case. I'll go head and mark this conversation as resolved.

musthafa1996:
I have also replaced the Parse SDK overriding repo ssh URL to https repo URL to avoid build issues in CI.
Screenshot 2022-10-31 at 7 24 31 PM

mtrezza:
Great, could you solve the conflict so we can see whether the CI passes? I think the next step then would be to revert the changes in package and package-lock and commit the fix, right?

musthafa1996:
I've resolved the conflicts @mtrezza.
Yes, the next step would be to revert the changes in package and package-lock.

musthafa1996:
@mtrezza, did you just run the CI. Looks like the old SSH repo URL (git@github.com:musthafa1996/Parse-SDK-JS.git#file-upload-issue) was being used instead of HTTPS repo URL (git+https://github.com/musthafa1996/Parse-SDK-JS.git#file-upload-issue). Is there a way we can ensue the latter is used?
image

mtrezza:
Is the change necessary for this PR? If not, let's keep it as is and revert any changes made in package and package-lock.

musthafa1996:
> @mtrezza, did you just run the CI. Looks like the old SSH repo URL ([git@github.com](mailto:git@github.com):musthafa1996/Parse-SDK-JS.git#file-upload-issue) was being used instead of HTTPS repo URL (git+https://github.com/musthafa1996/Parse-SDK-JS.git#file-upload-issue). Is there a way we can ensue the latter is used? image

This change is necessary. Otherwise you'll keep getting **Permission denied** error while accessing the git repo.

mtrezza:
Can you try reverting the package and package-lock please

mtrezza:
Also, I noticed that there is no test in the JS SDK for this change. Is it possible to add a test that fails without the change?

musthafa1996:
You mean reverting the overrides for Parse SDK too?

mtrezza:
Yes, the only file that this PR should change is `src/ParseFile.js`. And possibly a test file if we add a test.

musthafa1996:
> Also, I noticed that there is no test in the JS SDK for this change. Is it possible to add a test that fails without the change?

The failing test case wouldn't be possible unless you force Parse server to use the latest alpha version of the Parse SDK. The issue was induce in the recent updates to Parse SDK.

src/ParseFile.js
```diff
@@ -137,7 +138,10 @@ class ParseFile {
       } else if (data && typeof data.base64 === 'string') {
         const base64 = data.base64.split(',').slice(-1)[0];
         const dataType =
-          specifiedType || data.base64.split(';').slice(0, 1)[0].split(':').slice(1, 2)[0] || 'text/plain';
+          specifiedType ||
+          data.base64.split(';').slice(0, 1)[0].split(':').slice(1, 2)[0] ||
+          'text/plain';
```
mtrezza, line 142:
```suggestion
          specifiedType || data.base64.split(';').slice(0, 1)[0].split(':').slice(1, 2)[0] || 'text/plain';
```

mtrezza (commented):
If this was not a lint change, we should also keep this line as is

musthafa1996:
You can see why the redundant test case was removed in this message: https://github.com/parse-community/Parse-SDK-JS/pull/1578#discussion_r1006572203

mtrezza:
> The failing test case wouldn't be possible unless you force Parse server to use the latest alpha version of the Parse SDK. The issue was induce in the recent updates to Parse SDK.

A test case wouldn't be possible even with the override?

src/ParseFile.js
musthafa1996: Sure, let me update that.

musthafa1996:
> > The failing test case wouldn't be possible unless you force Parse server to use the latest alpha version of the Parse SDK. The issue was induce in the recent updates to Parse SDK.
> 
> A test case wouldn't be possible even with the override?

As you already know, all the existing test cases in **ParseFileTest.js** fails with the override of Parse JS SDK.  Any additional test case would a redundancy of the existing test cases.

musthafa1996:
Shall I revert all the changes in **package.json** and **package-lock.json**?

mtrezza:
The SDK has 2 types of tests: pure SDK tests and integration tests that use Parse Server. Could we write a pure SDK test for this change? This way the override would not be needed to demo the issue.

The test would essentially look at the internals of the SDK, possibly mock some interface and expect data which there isn't because of the bug. You could look at the tests in `/src/__tests__/ParseFile-test.js`, they don't require Parse Server.

For example:
- "should save file with metadata and tag options" --> mocks the `fileController`
- "should create new ParseFile with metadata and tags" --> looks at the internal `file._source.base64` property

I guess , the simplest way would be to add a test for `this._data = ...;` and expect that `this._data` has the correct value.

> Shall I revert all the changes in package.json and package-lock.json?

yes

musthafa1996:
Sure, i can look into that test. I’ll get back.

mtrezza:
To summarize, I think the easiest approach is to:

1.  Revert package and package-lock, there should be no changed in these files for this PR
2. Add a test to `/src/__tests__/ParseFile-test.js` and test for `this._data`.

That takes maybe a minute to adapt the code, so we can merge this PR.
Note: This PR is currently blocking the beta release https://github.com/parse-community/Parse-SDK-JS/pull/1589

dplewis:
@mtrezza This issue was caused by https://github.com/parse-community/Parse-SDK-JS/pull/1467 Not only was `_data` removed, the test was changed to prevent us from catching this removal

musthafa1996:
@dplewis, thanks for mentioning that PR. The question is, we still need `_data` to be assigned, don’t we? Or should we be using `_source.base64`.

dplewis:
Yes the error you are getting occurs because you are trying to create a Buffer from undefined on the server

https://github.com/parse-community/parse-server/blob/4c1befabf2e40bf3cf41b8b3db257435684f7a62/src/Routers/FilesRouter.js#L171-L174

src/ParseFile.js
```diff
@@ -123,6 +123,7 @@ class ParseFile {
           type: specifiedType,
         };
       } else if (typeof Blob !== 'undefined' && data instanceof Blob) {
+        this._data = data;
```
dplewis: I believe this should be removed since _data is always base64. You are setting a `Blob` here

dplewis (changes requested):

musthafa1996:
Which approach do you consider is best? Using `_data` or `_source.base64`?

src/ParseFile.js
musthafa1996: Thanks. Let me correct this.

dplewis:
_data since it’s used server side. If you want to give me access to your fork I can add the changes and tests to this PR.

@mtrezza I like the override added in package.json do you think we should point the parse module in the server to this SDK ‘file:../index.js’ or allow any Parse-JS-SDK module to be passed into the Parse Server Config on creation for developers to test

src/ParseFile.js
musthafa1996: This has been removed.

musthafa1996: Please mark this conversation as resolved if that's what is required.

musthafa1996:
> If you want to give me access to your fork I can add the changes and tests to this PR.

Hi @dplewis, sorry for the late reply. Sure, I can provide the access.

musthafa1996:
> @mtrezza I like the override added in package.json do you think we should point the parse module in the server to this SDK ‘file:../index.js’ or allow any Parse-JS-SDK module to be passed into the Parse Server Config on creation for developers to test

Unfortunately overriding with the self package won't work. I have tried the following:
```
"overrides": {
    "parse-server": {
      "parse": "file:./index.js"
    }
},
```
```
"overrides": {
    "parse-server": {
      "parse": "file:./node.js"
    }
},
```
```
"overrides": {
    "parse-server": {
      "parse": "file:./"
    }
},
```

None of these work. I have also tried creating a symbolic link of the package with `ln -s . ./parse` and using like follows:
```
"overrides": {
    "parse-server": {
      "parse": "file:./parse"
    }
},
```

With all the above, `npm install --force` works. But while running `npm run integration`, the following error is thrown:
```
npm run integration

> parse@3.5.0-beta.1 integration
> cross-env TESTING=1 jasmine --config=jasmine.json

internal/modules/cjs/loader.js:905
  throw err;
  ^

Error: Cannot find module 'parse/node'
Require stack:
- /Volumes/T7/Projects/DrHCo/Parse-SDK-JS/node_modules/parse-server/lib/Adapters/Files/FilesAdapter.js
- /Volumes/T7/Projects/DrHCo/Parse-SDK-JS/node_modules/parse-server/lib/Options/index.js
- /Volumes/T7/Projects/DrHCo/Parse-SDK-JS/node_modules/parse-server/lib/ParseServer.js
- /Volumes/T7/Projects/DrHCo/Parse-SDK-JS/node_modules/parse-server/lib/index.js
- /Volumes/T7/Projects/DrHCo/Parse-SDK-JS/integration/test/helper.js
- /Volumes/T7/Projects/DrHCo/Parse-SDK-JS/node_modules/jasmine/lib/jasmine.js
- /Volumes/T7/Projects/DrHCo/Parse-SDK-JS/node_modules/jasmine/bin/jasmine.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:902:15)
    at Function.Module._load (internal/modules/cjs/loader.js:746:27)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:101:18)
    at Object.<anonymous> (/Volumes/T7/Projects/DrHCo/Parse-SDK-JS/node_modules/parse-server/lib/Adapters/Files/FilesAdapter.js:9:36)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:12)
    at Module.require (internal/modules/cjs/loader.js:974:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Volumes/T7/Projects/DrHCo/Parse-SDK-JS/node_modules/parse-server/lib/Adapters/Files/FilesAdapter.js',
    '/Volumes/T7/Projects/DrHCo/Parse-SDK-JS/node_modules/parse-server/lib/Options/index.js',
    '/Volumes/T7/Projects/DrHCo/Parse-SDK-JS/node_modules/parse-server/lib/ParseServer.js',
    '/Volumes/T7/Projects/DrHCo/Parse-SDK-JS/node_modules/parse-server/lib/index.js',
    '/Volumes/T7/Projects/DrHCo/Parse-SDK-JS/integration/test/helper.js',
    '/Volumes/T7/Projects/DrHCo/Parse-SDK-JS/node_modules/jasmine/lib/jasmine.js',
    '/Volumes/T7/Projects/DrHCo/Parse-SDK-JS/node_modules/jasmine/bin/jasmine.js'
  ]
}
```

That's the reason I ended up overriding with the git repo:
```
"overrides": {
  "parse-server": {
    "parse": "git+https://github.com/musthafa1996/Parse-SDK-JS.git#file-upload-issue"
  }
},
```

musthafa1996:
@dplewis, I did add you as a collaborator.

musthafa1996:
@mtrezza, I have reverted all the changes in **package.json** and **package-lock.json**. I have also added validation of `ParseFile._data` in all the applicable ParseFile tests of the SDK. 

The following are the tests failing without the fix in **ParseFile.js**.
Screenshot 2022-11-03 at 1 20 59 PM
Screenshot 2022-11-03 at 1 21 19 PM

The following is the report of tests passing after adding the fix to **ParseFile.js**.
Screenshot 2022-11-03 at 1 27 02 PM
Screenshot 2022-11-03 at 1 27 38 PM

musthafa1996:
Looks like there was a problem with MongoDB. @mtrezza, can you re run the CI?

Screenshot 2022-11-03 at 1 44 54 PM

mtrezza:
Great, so if you now removed the fix, the tests would fail, even without the override of Parse Server, correct?

musthafa1996:
> Great, so without the fix, the tests would now fail, even without the override of Parse Server, correct?

Yes, that’s right.

mtrezza:
@dplewis Not sure about the override suggestion maybe that would just complicate things? I think the sustainable solution is to move all into a monorepo. We are planning to do that in 2023. For this PR, all looks good now.

musthafa1996:
Anything pending here?

mtrezza:
I don't think so, unless @dplewis sees anything?

dplewis:
@musthafa1996 I added a few more tests. @mtrezza I think this is ready for review

musthafa1996:
Any idea why we get this error in Github CI?
85E715BE-A238-4F1A-8341-BF8A9090E8F7

musthafa1996:
Rerunning the CI multiple times would get the integration test to succeed.

dplewis:
@musthafa1996 Its a flaky test should be fixed when https://github.com/parse-community/parse-server/pull/8232 gets merged in. Just reran the tests and everything checked out

musthafa1996:
Got it. Thanks 😊

mtrezza (approved):
Thanks everyone, looks good!

parseplatformorg:
🎉 This change has been released in version [3.5.0-alpha.8](https://github.com/parse-community/Parse-SDK-JS/releases/tag/3.5.0-alpha.8)

dplewis:
@musthafa1996 Thank you for looking into this `_data` was supposed to be internal on the JS side. I didn't know it was used on the server side. I'll add an addtional check and performance improvements on the server side

musthafa1996:
I’m glad that i could help fix the issue 😊

musthafa1996:
The issue was kind of tricky to find too due to parse server using an older version of the SDK. I’m really looking forward to mono-repo. Development and bug finding will be much easier.

parseplatformorg:
🎉 This change has been released in version [3.5.1-beta.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/3.5.1-beta.1)

parseplatformorg:
🎉 This change has been released in version [3.5.1-alpha.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/3.5.1-alpha.1)

parseplatformorg:
🎉 This change has been released in version [3.5.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/3.5.1)

Issue #1535, opened 2022-08-28
Remove Node 12 and 15 support


SebC99:
### New Feature / Enhancement Checklist
- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-JS/issues?q=is%3Aissue).

### Current Limitation
Current build settings allow for node.js from version 8. But as far as I can see, it can only be used via parse-server which only supports nodejs from version 12.

### Feature / Enhancement Description
Could we change node support for Parse SDK JS as well? It would allow using ES7 features as object destructuring without the need for a polyfill.

### Example Use Case
As an example, [these lines](https://github.com/parse-community/Parse-SDK-JS/blob/e08f3b6f85ac16ff57da00b044b70159eaea95c1/src/ObjectStateMutations.js#L119-L121) seems to be the slowest (by far) in Parse.Object getter, which is called a lot in any code. Replacing these lines with a simple `const data = {...serverData};` is more than 80% faster according to jsbench.

mtrezza:
I'd suggest:
1. Create a PR which sets the min node version for the JS SDK to >=12 and refactors the bottlenecks you mentioned
2. Publish a minor release in Parse Server with the new JS SDK

Later, when we are bumping the node version to >=14 in Parse Server, we can also so that in the JS SDK. The bump to Node 14 in Parse Server will happen in 4 months with Parse Sever 6.x, at the end of the year.

mtrezza:
A PR for this should cover:
- [ ] Remove support for Node 12 and 15
- [ ] Add `engines.node` to package.json, not sure why this is currently missing
- [ ] Update Node compatibility table in README
- [ ] Update Node badge table in README

PR #1598, opened 2022-11-06
feat: Add Node 16 and 18 support

### LINKED ISSUE #1597: Add Node 16 and 18 support
### New Issue Checklist

### Issue Description

The SDK is currently not tested with Node 16 and 18, despite Parse Server supporting these Node versions. So we should add the same CI test environments here.

### Suggestion

Add Node 16, 18 environments to CI.

dblythy:

### Issue Description

Node 16 and 18 should be in the CI

Related issue: #1597
Closes: #1597

### Approach

Adds node 16 and 18

### TODOs before merging

- [x] Add tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
         with:
           version: 1
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     timeout-minutes: 30
     env:
       MONGODB_VERSION: 3.6.9
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,12 +24,8 @@ jobs:
     strategy:
       matrix:
         include:
-          - name: Node 12
-            NODE_VERSION: 12.22.11
           - name: Node 14
             NODE_VERSION: 14.21.1
-          - name: Node 15
-            NODE_VERSION: 15.14.0
           - name: Node 16
             NODE_VERSION: 16.18.1
           - name: Node 18
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,13 +27,13 @@ jobs:
           - name: Node 12
             NODE_VERSION: 12.22.11
           - name: Node 14
-            NODE_VERSION: 14.19.1
+            NODE_VERSION: 14.21.1
           - name: Node 15
             NODE_VERSION: 15.14.0
           - name: Node 16
-            NODE_VERSION: 16.14.2
+            NODE_VERSION: 16.18.1
           - name: Node 18
-            NODE_VERSION: 18.8.0
+            NODE_VERSION: 18.12.1
       fail-fast: false
     steps:
     - name: Fix usage of insecure GitHub protocol
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,3 +53,6 @@ jobs:
       env:
         CI: true
     - run: bash <(curl -s https://codecov.io/bash)
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,8 +19,18 @@ jobs:
   build:
     runs-on: ubuntu-18.04
     timeout-minutes: 30
-    env:      
+    env:
       MONGODB_VERSION: 3.6.9
+    strategy:
+      matrix:
+        include:
+          - name: Node 14
+            NODE_VERSION: 14.19.1
+          - name: Node 16
+            NODE_VERSION: 16.14.2
+          - name: Node 18
+            NODE_VERSION: 18.8.0
+      fail-fast: false
     steps:
     - name: Fix usage of insecure GitHub protocol
       run: sudo git config --system url."https://github".insteadOf "git://github"
@@ -28,7 +38,7 @@ jobs:
     - name: Use Node.js
       uses: actions/setup-node@v1
       with:
-        node-version: 14
+        node-version: ${{ matrix.node-version }}
     - name: Cache Node.js modules
       uses: actions/cache@v2
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,8 +19,6 @@ jobs:
   build:
     runs-on: ubuntu-latest
     timeout-minutes: 30
-    env:
-      MONGODB_VERSION: 3.6.9
     strategy:
       matrix:
         include:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
     - name: Use Node.js
       uses: actions/setup-node@v1
       with:
-        node-version: ${{ matrix.node-version }}
+        node-version: ${{ matrix.NODE_VERSION }}
     - name: Cache Node.js modules
       uses: actions/cache@v2
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
         with:
           version: 1
   build:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     timeout-minutes: 30
     env:
       MONGODB_VERSION: 3.6.9
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,8 +24,12 @@ jobs:
     strategy:
       matrix:
         include:
+          - name: Node 12
+            NODE_VERSION: 12.22.11
           - name: Node 14
             NODE_VERSION: 14.19.1
+          - name: Node 15
+            NODE_VERSION: 15.14.0
           - name: Node 16
             NODE_VERSION: 16.14.2
           - name: Node 18
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -11,7 +11,7 @@
 ### Issue Description
 <!-- Add a brief description of the issue this PR solves. -->
 
-Related issue: #`FILL_THIS_OUT`
+Closes: FILL_THIS_OUT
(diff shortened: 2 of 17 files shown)


mtrezza (approved):
Looks good!

mtrezza (changes requested):
- Could you add Node 16 and 18 to the node tag in README?
image

- Could you add Node 12 and 15 to the CI? We're currently indicating that the SDK is compatible with those.

We have https://github.com/parse-community/Parse-SDK-JS/issues/1535 where we'll drop Node 12 and 15 support and add the `engines` block to package.json.

dblythy:
`npm run integration` seems to be a bit flaky on Node 18

.github/workflows/ci.yml
```diff
@@ -19,16 +19,30 @@ jobs:
...
+      fail-fast: false
     steps:
     - name: Fix usage of insecure GitHub protocol
       run: sudo git config --system url."https://github".insteadOf "git://github"
     - uses: actions/checkout@v2
     - name: Use Node.js
       uses: actions/setup-node@v1
       with:
-        node-version: 14
+        node-version: ${{ matrix.node-version }}
```
dplewis: This should be `matrix.NODE_VERSION` if you look at the CI all the test run against `v16.18.0`.

Also should we worry about the tests running against package-lock.json v2?

dplewis (changes requested):

.github/workflows/ci.yml
mtrezza:
```suggestion
        node-version: ${{ matrix.NODE_VERSION }}
```

mtrezza: @dblythy Did you copy/paste this from another workflow file? Maybe it needs to be corrected there as well.

.github/workflows/ci.yml
```diff
@@ -19,16 +19,30 @@ jobs:
...
+          - name: Node 12
+            NODE_VERSION: 12.22.11
+          - name: Node 14
+            NODE_VERSION: 14.19.1
+          - name: Node 15
+            NODE_VERSION: 15.14.0
+          - name: Node 16
+            NODE_VERSION: 16.14.2
+          - name: Node 18
+            NODE_VERSION: 18.8.0
```
mtrezza:
```suggestion
            NODE_VERSION: 18.12.1
```

.github/workflows/ci.yml
```diff
@@ -19,16 +19,30 @@ jobs:
...
+      matrix:
+        include:
+          - name: Node 12
+            NODE_VERSION: 12.22.11
+          - name: Node 14
+            NODE_VERSION: 14.19.1
+          - name: Node 15
+            NODE_VERSION: 15.14.0
+          - name: Node 16
+            NODE_VERSION: 16.14.2
```
mtrezza:
```suggestion
          - name: Node 16
            NODE_VERSION: 16.18.1
```

.github/workflows/ci.yml
```diff
@@ -19,16 +19,30 @@ jobs:
...
-    env:      
+    env:
       MONGODB_VERSION: 3.6.9
+    strategy:
+      matrix:
+        include:
+          - name: Node 12
+            NODE_VERSION: 12.22.11
+          - name: Node 14
+            NODE_VERSION: 14.19.1
```
mtrezza:
```suggestion
            NODE_VERSION: 14.21.1
```

.github/workflows/ci.yml
dplewis: @mtrezza @dblythy Should we add `npm install --no-save` to the CONTRIBUTION.md guide? I run npm > 7 and have to keep downgrading before my PR. This will prevent overriding `lockfileVersion: 2` in the `package-lock.json` from being commited.

https://github.blog/2021-02-02-npm-7-is-now-generally-available/#changes-to-the-lockfile

.github/workflows/ci.yml
```diff
@@ -19,16 +19,30 @@ jobs:
   build:
     runs-on: ubuntu-18.04
```
mtrezza:
```suggestion
    runs-on: ubuntu-latest
```

dplewis: I'd rather pin it to `20.04` as the latest `20.10` isn't an LTS version.

mtrezza: GitHub is using "latest" in all their code examples, I think they are only offering LTS anyway. The CI ran on:

> Operating System
> Ubuntu
> 20.04.5
> LTS

Also, looking at the [PRs](https://github.com/actions/runner-images/issues/6399) that bump the "latest" tag on GitHub, it seems they are testing this throughly before bumping. So GitHub's "latest" doesn't seem to mean Ubuntu's "latest", but the latest version they find stable and suitable for GitHub Actions.

mtrezza: GitHub is suggesting to use latest in all their code examples, I think they are only offering LTS anyway:

> Operating System
> Ubuntu
> 20.04.5
> LTS

mtrezza: Interesting that the CI fails with “latest" (20.04.5) though... Maybe a GH issue, will re-run CI later

dplewis: Puppeteer doesn't support Node 12 which we were going to remove anyway.

mtrezza: So we'll stay on latest for now, it should work.

.github/workflows/ci.yml
mtrezza: We didn't really have much of an issue with this. Whether the option is needed depends on the PR, so we'd need a whole paragraph devoted to explaining this. This is most relevant for beginners, but for them the contribution docs are already complex. I'd rather have someone open a PR with a wrong lock file than no PR because they are discouraged by the contribution docs. I think the docs needs a fundamental rework anyway.

mtrezza: Resolving as node ref is fixed

.github/workflows/ci.yml
```diff
@@ -17,18 +17,32 @@ jobs:
...
       MONGODB_VERSION: 3.6.9
+    strategy:
+      matrix:
+        include:
+          - name: Node 12
+            NODE_VERSION: 12.22.11
+          - name: Node 14
+            NODE_VERSION: 14.21.1
+          - name: Node 15
+            NODE_VERSION: 15.14.0
```
mtrezza:
```suggestion
          - name: Node 14
            NODE_VERSION: 14.21.1
```

.github/workflows/ci.yml
```diff
@@ -17,18 +17,28 @@ jobs:
         with:
           version: 1
   build:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
mtrezza, line 20:
```suggestion
    runs-on: ubuntu-18.04
```

Trying to switch back to fixed version due to failing integration tests

mtrezza:
Ubuntu 18.04.6 (LTS): Node 14, 16 pass; Node 18 fails
Ubuntu 20.04.5 (LTS): Node 14, 16 fail; Node 18 passes

Found [this](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) regarding "latest":

> Note: The -latest runner images are the latest stable images that GitHub provides, and might not be the most recent version of the operating system available from the operating system vendor.

So staying on "latest" should be fine.
Switching back to "latest"...

.github/workflows/ci.yml
mtrezza: Found [this](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) regarding "latest":

> Note: The -latest runner images are the latest stable images that GitHub provides, and might not be the most recent version of the operating system available from the operating system vendor.

.github/workflows/ci.yml
```diff
@@ -19,16 +19,26 @@ jobs:
   build:
     runs-on: ubuntu-18.04
```
mtrezza:
```suggestion
    runs-on: ubuntu-latest
```

dplewis:
@mtrezza @dblythy `matrix.node-version` is used in the parse-server CI want to change it?

dplewis:
Why is there an extra build step?

Screen Shot 2022-11-10 at 8 13 51 AM

mtrezza:
build step is still required in repo settings, we'll remove after this PR

> matrix.node-version is used in the parse-server CI want to change it?

yes, would be good to use the same nomenclature across repos

mtrezza:
Note 18 tests are failing randomly, could someone look into this?

mtrezza:
In Parse Server adding Node 18 support required to set the DNS resolution order, because it first tries to resolve for IPv6. So instead of changing `localhost` to an IPv4 `127.0.0.1`, we set this in [helper.js](https://github.com/parse-community/parse-server/blob/6e66b20e28462ae7d6456c2519682e29365556fa/spec/helper.js#L8-L11):

```
// Ensure localhost resolves to ipv4 address first on node v17+
if (dns.setDefaultResultOrder) {
  dns.setDefaultResultOrder('ipv4first');
}
```

This has the benefit that `localhost` can still be used and reviewers don't need to keep that in mind when reviewing PRs in the future.

mtrezza (approved):
Looks good!

mtrezza:
Rerunning the failed tests; then this should be ready for merge, right?

mtrezza:
Rerunning CI

mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [4.0.0-alpha.4](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.0.0-alpha.4)

parseplatformorg:
🎉 This change has been released in version [4.0.0-beta.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.0.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [4.0.0](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.0.0)

PR #1619, opened 2022-11-25
fix: ParseError message handling 

stewones:

- [ ] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-SDK-JS/issues?q=is%3Aissue).

### Issue Description

The problem is that log is unclear when some random error happens in the server, eg: MongoDB query has exceeded memory limit

```
error ParseError: [object Object]
    at handleError (/my-project/node_modules/parse-server/node_modules/parse/lib/node/RESTController.js:418:17)
    at processTicksAndRejections (internal/process/task_queues.js:95:5) {
  code: 1
}
```

### Approach

My proposal is just to make sure the error message is a string otherwise, stringify it.

So this would be the returned log for the example above
```
error ParseError: "{\"ok\":0,\"code\":292,\"codeName\":\"QueryExceededMemoryLimitNoDiskUseAllowed\"}"
    at handleError (/my-project/node_modules/parse-server/node_modules/parse/lib/node/RESTController.js:418:17)
    at processTicksAndRejections (internal/process/task_queues.js:95:5) {
  code: 1
}
```

Please note that it's strictly necessary making sure those logs are readable, especially when relying on monitoring services like GCP Error Reporting, New Relic, Data Dog, etc...

- [x] Add tests

### DIFF
--- a/src/ParseError.js
+++ b/src/ParseError.js
@@ -22,7 +22,14 @@ class ParseError extends Error {
     this.code = code;
     Object.defineProperty(this, 'message', {
       enumerable: true,
-      value: typeof message === 'string' ? message : JSON.stringify(message),
+      value:
+        typeof message === 'string'
+          ? message
+          : typeof message === 'object' &&
+            typeof message.toString === 'function' &&
+            message.toString() !== '[object Object]'
+            ? message.toString()
+            : JSON.stringify(message),
     });
   }
 
--- a/src/ParseError.js
+++ b/src/ParseError.js
@@ -20,6 +20,15 @@ class ParseError extends Error {
   constructor(code, message) {
     super(message);
     this.code = code;
+
+    const stringify = obj => {
+      let log = '';
+      for (const k in obj) {
+        log += `${obj[k]} `;
+      }
+      return log.trim();
+    };
+
     Object.defineProperty(this, 'message', {
       enumerable: true,
       value:
@@ -27,9 +36,9 @@ class ParseError extends Error {
           ? message
           : typeof message === 'object' &&
             typeof message.toString === 'function' &&
-            message.toString() !== '[object Object]'
+            !message.toString().includes('[object Object]')
             ? message.toString()
-            : JSON.stringify(message),
+            : stringify(message),
     });
   }
 
--- a/src/ParseError.js
+++ b/src/ParseError.js
@@ -22,7 +22,7 @@ class ParseError extends Error {
     this.code = code;
     Object.defineProperty(this, 'message', {
       enumerable: true,
-      value: message,
+      value: typeof message === 'string' ? message : JSON.stringify(message),
     });
   }
 
--- a/src/__tests__/ParseError-test.js
+++ b/src/__tests__/ParseError-test.js
@@ -29,11 +29,25 @@ describe('ParseError', () => {
   });
 
   it('message must be a string', () => {
+    /**
+     * error as object
+     */
     const someRandomError = { code: 420, message: 'time to chill' };
     const error = new ParseError(1337, someRandomError);
     expect(JSON.parse(JSON.stringify(error))).toEqual({
       message: JSON.stringify(someRandomError),
       code: 1337,
+    });
+
+    /**
+     * error as an Error instance
+     */
+    const someRandomError2 = new Error('time to relax');
+    const error2 = new ParseError(420, someRandomError2);
+
+    expect(JSON.parse(JSON.stringify(error2))).toEqual({
+      message: 'Error: time to relax',
+      code: 420,
     });
   });
 });
--- a/src/__tests__/ParseError-test.js
+++ b/src/__tests__/ParseError-test.js
@@ -27,4 +27,13 @@ describe('ParseError', () => {
       code: 123,
     });
   });
+
+  it('message must be a string', () => {
+    const someRandomError = { code: 420, message: 'time to chill' };
+    const error = new ParseError(1337, someRandomError);
+    expect(JSON.parse(JSON.stringify(error))).toEqual({
+      message: JSON.stringify(someRandomError),
+      code: 1337,
+    });
+  });
 });
--- a/src/__tests__/ParseError-test.js
+++ b/src/__tests__/ParseError-test.js
@@ -28,26 +28,50 @@ describe('ParseError', () => {
     });
   });
 
-  it('message must be a string', () => {
-    /**
-     * error as object
-     */
-    const someRandomError = { code: 420, message: 'time to chill' };
+  it('message can be a string', () => {
+    const someRandomError = 'oh no';
+
     const error = new ParseError(1337, someRandomError);
+
     expect(JSON.parse(JSON.stringify(error))).toEqual({
-      message: JSON.stringify(someRandomError),
+      message: someRandomError,
       code: 1337,
     });
+  });
 
-    /**
-     * error as an Error instance
-     */
-    const someRandomError2 = new Error('time to relax');
-    const error2 = new ParseError(420, someRandomError2);
+  it('message can be an object passed trough some external dependency', () => {
+    const someRandomError = { code: '420', message: 'time to chill', status: '🎮' };
 
-    expect(JSON.parse(JSON.stringify(error2))).toEqual({
-      message: 'Error: time to relax',
-      code: 420,
+    const error = new ParseError(1337, someRandomError);
+
+    expect(JSON.parse(JSON.stringify(error))).toEqual({
+      message: '420 time to chill 🎮',
+      code: 1337,
+    });
+  });
+
+  it('message can be an Error instance *receiving a string* passed trough some external dependency', () => {
+    const someRandomError = new Error('good point');
+
+    const error = new ParseError(1337, someRandomError);
+
+    expect(JSON.parse(JSON.stringify(error))).toEqual({
+      message: 'Error: good point',
+      code: 1337,
+    });
+  });
+
+  it('message can be an Error instance *receiving an object* passed trough some external dependency', () => {
+    const someRandomErrorWrong = new Error({
+      code: 'WRONG',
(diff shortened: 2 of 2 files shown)


src/__tests__/ParseError-test.js
```diff
@@ -27,4 +27,27 @@ describe('ParseError', () => {
       code: 123,
     });
   });
+
+  it('message must be a string', () => {
+    /**
+     * error as object
```
mtrezza: Could you please split this into 2 separate tests and remove the comments. If a comment is needed to explain a test that is usually an indication that it should be a separate test.

src/ParseError.js
```diff
@@ -22,7 +22,14 @@ class ParseError extends Error {
     this.code = code;
     Object.defineProperty(this, 'message', {
       enumerable: true,
-      value: message,
+      value:
+        typeof message === 'string'
+          ? message
+          : typeof message === 'object' &&
+            typeof message.toString === 'function' &&
```
mtrezza, line 38: Why `typeof message.toString === 'function' &&` and `message.toString() !== '[object Object]'`?

src/__tests__/ParseError-test.js
stewones: sure thing

mtrezza:
> error ParseError: "{\"ok\":0,\"code\":292,\"codeName\":\"QueryExceededMemoryLimitNoDiskUseAllowed\"}"

I would say that this is actually an unwanted error message style. `[object object]` is really an indication that an error is not properly logged, but just passed through from a dependency. So maybe it's not even a good idea to just stringily and output whatever is passed to ParseError.

The problem is that it's not clear whether this error message is from Parse Server, the Parse JS SDK or MongoDB. People may then search for "Parse Server error 292" and not find an answer for the issue, or worse references to a completely unrelated Parse Server error 292. Whereas "MongoDB error 292" would likely yield a better search result. It's also unclear for the developer what the `"\"ok\":0,\` part means and what relevance it has, because we are just passing everything through without context.

Maybe the better solution is to properly write an error message and package the dependency error into it, like:
> error ParseError: MongoDB error 292: QueryExceededMemoryLimitNoDiskUseAllowed

If the MongoDB errors vary in structure, then we may not be able to that; but we should at least add the error source, like:

> error ParseError: MongoDB error: "{\"ok\":0,\"code\":292,\"codeName\":\"QueryExceededMemoryLimitNoDiskUseAllowed\"}"

Maybe we can also get rid of the quotes and format it more nicely, with [util.inspect()](https://nodejs.org/api/util.html#utilinspectobject-options).

stewones:
> [object object] is really an indication that an error is not properly logged, but just passed through from a dependency

this is what mongodb is giving us, no third party or anything, just to be clear. so Parse isn't handling it properly right now. create a collection with more than 10k records and try to sort it with no custom index, that's what you'd get.

stewones:
so the way I look at it this solution is just printing whatever message is being passed as a string, regardless of its type. so instead of `[object, Object]` we have a nice readable error message. after all, that's not the sense of a Error Instance?

src/ParseError.js
stewones: yeah that's crazy I know, I'm not sure why yet but inside the test env `.toString()` is giving `[object Object]`.

stewones:
I think I get your point @mtrezza 
let me see if I can improve it

src/ParseError.js
mtrezza: Maybe [util.inspect()](https://nodejs.org/api/util.html#utilinspectobject-options) will give you a more reliable and better formatted output without quotes.

stewones: but isn't `util` package from node? does it work in browser too? cause this ParseError thing is from the js sdk

stewones: nevermind we really need this check.

answering your question, due to this edge case that might be uncommon but who knows 

Screenshot 2022-11-26 at 14 55 43

that's a case where nobody will be ever able to extract a readable message so Parse would just throw `Parse Error: `  instead of `Parse Error: [object Object]`

stewones: unless I'm missing something, that's not how Errors should be thrown, but I see lots of `Parse Error: [object Object]` in my server logs and would like to cover this case as well. I've already added a test case exclusive for it, going to push soon.

stewones: take a look now @mtrezza it's more refined and test covered

stewones: and this is how it's looking like

Screenshot 2022-11-26 at 15 58 11

stewones:
this is how it's looking like now

Screenshot 2022-11-26 at 15 58 11

src/ParseError.js
mtrezza: > but isn't util package from node? 

yes you are right, sorry for mixing things up here

I'll post a broader comment in the review and close this.

mtrezza (commented):
Going back, I think we should never throw an object directly without context. These [edge cases](https://github.com/parse-community/Parse-SDK-JS/pull/1619/files#r1032816441) should not even make it through a PR review, because looking at the [spec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error), I would say that `new Error(message)` expects a string, not an object. From that I would say that `ParseError(message)` also expects `message` to be a string.

If we want to include the original error object for reference, we can use the options of the `Error` object, see [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error#rethrowing_an_error_with_a_cause):
```
throw new Error("New error message", { cause: err });
```

It seems the issue is not that `ParseError` doesn't handle non-string values. The issue is that the code throws non-string values, which seems to be incorrect from a type perspective. I would say this PR doesn't need any of these object to string conversions. It should just fix how the error in thrown originally.

stewones:
> These [edge cases](https://github.com/parse-community/Parse-SDK-JS/pull/1619/files#r1032816441) should not even make it through a PR review

Yeah, I agree but we don't control third-party code running on our servers. 

while I'm ok removing the edge case it doesn't resolve the issue. if it happens to Parse catch some error wrongly thrown, the server will still log `ParseError: [object Object]` giving no clue. Even worse, it's co-relating one external dumb mistake with Parse.

> It should just fix how the error in thrown originally.

what do you mean?

stewones:
> If we want to include the original error object for reference, we can use the options of the Error object, see [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error#rethrowing_an_error_with_a_cause):

not at all, we just want a nice and readable error message. I personally think my approach by stringifying objects could be useful for those scenarios where we don't control what is coming externally. developers are humans and make mistakes, and unfortunately for JS engines `new Error({...})` is a valid syntax, even though docs say it's not.

matheusdavidson:
Hey @stewones @mtrezza, I also have this problem, I agree the PR seems to be solving the problem directly but what is the best way to solve this @mtrezza? cause we do have an [object, object] going on.

Should we add a `cause` like you suggested? even tho that's no implemented yet. Or is it in a higher class?

mtrezza:
Where is the code that throws the error and creates a new ParseError object?

stewones:
> Where is the code that throws the error and creates a new ParseError object?

@mtrezza  did you see my screenshot above? it starts from the RESTController, I'm not an expert in Parse codebase (yet :D) but I don't think it's something worth it to go deep into every place where ParseError is consumed just to verify if an object or string is being passed.

why not treat this for general purpose like I'm doing here? at least accept the fact that we can't always expect a string in ParseError (i.e. MongoDB is throwing a `new Error('string')`)  and we *must* decode this message to avoid the [object Object] bubble.

image

stewones:
> MongoDB is throwing a new Error('string')

how do I know this?

because if I simply put a `message.toString()` inside ParseError it shows the log correctly ^

I just expanded it to a broader situation where the edge cases mentioned in the beginning may happen. I'm a Parse user for years and started to contribute only recently, but this ParseError: [object Object] thing is smth that always disgusted me so much. we need to fix this asap! also, it will be "required" at some point for the allowDiskUse option I'm going to PR, cause users will get the [object Object] if they don't set a sort specification (as shown in the ^ screenshot)

happy to remove those edge cases if you don't like it and let only `message.toString()`. but keep in mind that people will continue co-relating those errors with Parse giving a negative perception of the library.

just lmk

stewones:
and I just expanded it to a broader situation where the edge cases mentioned in the beginning may happen. I'm a Parse user for years, started to contribute only recently, but this object bubble thing is smth that always disgusted me so much. we need to fix this asap!

mtrezza:
I get your point, we should get rid of these `[obj obj]` entries. If the error is thrown in `RESTController`, then maybe the solution is to construct a proper error there, and not just create an error as `new ParseError(anything)`. See my [argument](https://github.com/parse-community/Parse-SDK-JS/pull/1619#issuecomment-1327940635) above for why I think that would be better. I think outputting a dependency error object without context is not good error logging.

We could take some inspiration from the `Error` object (from which `ParseError` inherits) and its argument types. The `Error` spec has a dedicated [cause](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error#parameters) argument for re-throwing another error. Like we are re-throwing the MongoDB error. But that is separate from the `message` argument, which seems to be a string type. In other words, if we pass an object to a string type argument, then of course we will log `[obj obj]` and the solution is to pass a string. 

In fact, the type of `message` is already a string:
https://github.com/parse-community/Parse-SDK-JS/blob/5c140299109d2850982155ac1e1b907e1895fde0/src/ParseError.js#L18

If we had typescript already rolled out, then it would show a type mismatch error in Parse Server where an object is passed, not in here the Parse JS SDK. I get that just stringifying `message` is a quick solution. But we'd be violating our own type definitions.

I think it would help to understand this better if we would first find the code in Parse Server that throws the MongoDB error.

PR #1623, opened 2022-11-27
feat: Add ability to allowDiskUse

stewones:

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues/7765)

### Issue Description
this is so we can pass a boolean to `ParseQuery`

i.e.

```
const GameScore = Parse.Object.extend("GameScore");
const query = new Parse.Query(GameScore);

query.allowDiskUse(true); // <----------

query.equalTo("playerName", "Dan Stemkoski");
const results = await query.find();
```

### TODOs before merging

- [ ] Add tests

### DIFF
--- a/src/ParseQuery.js
+++ b/src/ParseQuery.js
@@ -40,6 +40,7 @@ export type QueryJSON = {
   hint?: mixed,
   explain?: boolean,
   readPreference?: string,
+  allowDiskUse?: Boolean,
   includeReadPreference?: string,
   subqueryReadPreference?: string,
 };
@@ -239,6 +240,7 @@ class ParseQuery {
   _count: boolean;
   _order: Array<string>;
   _readPreference: string;
+  _allowDiskUse: boolean;
   _includeReadPreference: string;
   _subqueryReadPreference: string;
   _queriesLocalDatastore: boolean;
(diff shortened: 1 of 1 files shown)


mtrezza:
`query.allowDiskUse(true); // <----------`

I'm not sure about this syntax. This is not a `Parse.Query` property but a meta property of the query, actually it's a database driver parameter. Such meta properties are usually set in the options, like: `await query.find({ useMasterKey: true, allowDiskUse: true });`

stewones:
yeah I just followed what `readPreferences` does, actually was planning to add support for both ways

stewones:
see this line @mtrezza 
https://github.com/parse-community/Parse-SDK-JS/blob/ad1070568633c01d418c833f29fe188efd55f935/src/ParseQuery.js#L1959

mtrezza:
Right, I guess there wasn't a clear strategy so far how to add options. Allowing to set as Parse.Query property *and* meta property is a code duplication. Let's decide for one way or the other. What do you think makes more sense?

One strategy to decide is to look at what the option does:
- An option that influences the query response (i.e. response content) should be a `Parse.Query` property
- An option that doesn't influence the query response but only changes how a query is executed should be a meta option

stewones:
makes sense @mtrezza 
so we should go as a meta option

mtrezza:
Alright sounds good

mtrezza:
What's the state of this PR? Do you plan to continue with it or should we close it for now?

stewones:
Yeah you can close this and any other PR of mine. Lost interest in Parse unfortunately since it's blocking me more than helping. And I understand that, not every tool is the right choice for every need.

Thanks.

mtrezza:
Sorry to hear that and thanks for the heads up. If you have any other feedback, please feel free to share. All the best!

PR #1641, opened 2022-12-23
ci: Add support for Parse Server 6

### LINKED ISSUE #1637: Prepare for Parse Server 6

### Current Limitation

The SDK needs to be prepared for the breaking changes coming with Parse Server 6.
As we can see in https://github.com/parse-community/Parse-SDK-JS/pull/1636, the tests are failing.

A PR to fix this may break compatibility with Parse Server <6; a compatibility table needs to be added to the README for Parse Server.

n/a

### Example Use Case
n/a

n/a

dblythy:

### Issue Description

SDK does not currently support Parse Server 6, as the startup method has changed.

Closes: #1637

### TODOs before merging

- [x] Add tests

### DIFF
--- a/integration/test/ParseLiveQueryTest.js
+++ b/integration/test/ParseLiveQueryTest.js
@@ -285,7 +285,7 @@ describe('Parse LiveQuery', () => {
   it('connectPromise does throw', async () => {
     await reconfigureServer({
       cloud({ Cloud }) {
-        Cloud.beforeConnect((params) => {
+        Cloud.beforeConnect(params => {
           if (params.sessionToken === 'testToken') {
             throw 'not allowed to connect';
           }
--- a/integration/test/ParseObjectTest.js
+++ b/integration/test/ParseObjectTest.js
@@ -1644,49 +1644,36 @@ describe('Parse Object', () => {
       });
   });
 
-  it('merges user attributes on fetchAll', done => {
+  it('merges user attributes on fetchAll', async () => {
     Parse.User.enableUnsafeCurrentUser();
-    let sameUser;
+    const acl = new Parse.ACL();
+    acl.setPublicReadAccess(true);
     let user = new Parse.User();
     user.set('username', 'asdf');
     user.set('password', 'zxcv');
     user.set('foo', 'bar');
-    user
-      .signUp()
-      .then(() => {
-        Parse.User.logOut();
-        const query = new Parse.Query(Parse.User);
-        return query.get(user.id);
-      })
-      .then(userAgain => {
-        user = userAgain;
-        sameUser = new Parse.User();
-        sameUser.set('username', 'asdf');
-        sameUser.set('password', 'zxcv');
-        return sameUser.logIn();
-      })
-      .then(() => {
-        assert(!user.getSessionToken());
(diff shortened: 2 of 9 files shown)


dblythy:
@mtrezza The SDK should still be compatible with previous versions, it's only the startup method which is a `devDependancy` that breaks the CI

dplewis:
@mtrezza Is this ready to merge? I have to copy the startup code to make PR's

mtrezza (approved):
Looks good!
@dblythy is this ready for merge?

Is this a breaking change? If so, we'd need to compose a note for the changelog.

If this breaks compatibility with Parse Server <6, a compatibility table needs to be added to the README for Parse Server.

dplewis:
@mtrezza Looks like its linted code and the server is only here for testing. No breaking changes.

mtrezza:
There is also a change in package lock, that's strange if there are only test and lint changes.

dplewis:
@mtrezza You can't test against latest of Parse Server without bumping it. In the package-lock.json they are all dev dependencies. `dev: true`. Dev deps don't affect the build

mtrezza:
Ah right, we import the alpha branch of Parse Server here, hence the lock file changes. Really not an ideal solution.

Let's think this through once more. So this is actually not a `feat` commit, but a `ci` commit, since it's only adapting the tests. If we are adapting the tests for Parse Server 6, we are not testing anymore against Parse Server 5. Since Parse Server 6 has breaking changes, I assume there is a also a chance that the Parse JS SDK becomes incompatible with Parse Server 5 at some point.

It may be problematic that we are only testing with the latest Parse Server alpha branch. If we could add a CI test for at least the latest Parse Server 5 (branch `release-5.x.x`), we would know when the compatibility breaks.

mtrezza:
So I guess this is also not a `feat` commit, but actually just a `ci` commit, since it's only test related, right?

mtrezza:
Since we are preparing the Parse JS SDK 4 major release, I guess we could add a compatbility table to the README and say that JS SDK 4 is only compatible with Parse Server 6. That may not be true (yet), but it aligns with our premise that we only officially support versions that the CI tests against.

mtrezza:
To get ahead here, let's merge this for now and assume the SDK stays compatible with Parse Server 5. We should add Parse Server 5 testing to the CI to verify compatibility, therefore opened https://github.com/parse-community/Parse-SDK-JS/issues/1651.

mtrezza:
Node 18 fails repeatedly with the same error:

> 1) Parse User cannot save non-authed user
>  - Unhandled promise rejection: AssertionError [ERR_ASSERTION]: 101 == 206

parseplatformorg:
🎉 This change has been released in version [4.0.0-beta.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.0.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [4.0.0](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.0.0)

parseplatformorg:
🎉 This change has been released in version [4.0.0-alpha.6](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.0.0-alpha.6)

PR #1662, opened 2023-01-23
fix: Saving a new `Parse.Object` with an unsaved `Parse.File` fails

### LINKED ISSUE #1660: unsaved children does not work for unsaved Parse.File
### New Issue Checklist

### Issue Description

When saving a new Parse.Object that is saved with a new Parse.File, the following error occurs.

`TypeError: object._getId is not a function. (In 'object._getId()', 'object._getId' is undefined)`

Likely it's from `_getId()` not being implemented for the `ParseFile` class? Since `_getId()` is called here when saving unsaved children. https://github.com/parse-community/Parse-SDK-JS/blob/8115e959533d1676fe5e5551bc81888b21fc12ef/src/unsavedChildren.js#L37 

### Steps to reproduce

````ts
// a File object
const file : File;
const parseFile = new Parse.File('myFileName',  file);
const parseObject = new Parse.Object({ prop1: "hello", file: parseFile });
parseObject.save();
````
### Actual Outcome

`TypeError: object._getId is not a function. (In 'object._getId()', 'object._getId' is undefined)`

### Expected Outcome

No errors

###  Environment

Node.js v16.13.1

Server
- Parse Server version: `^5.4.0`
- Operating system: `Mac OS Monterey 12.5.1, Ubuntu 22.04`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): Local network served Node.JS

Database
- Database version: `6.0.2`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): Local network served

Client
- Parse JS SDK version: `3.5.1`

### Logs

`TypeError: object._getId is not a function. (In 'object._getId()', 'object._getId' is undefined)`

### PR DESCRIPTION
**Pull request by swittk**
Fixes https://github.com/parse-community/Parse-SDK-JS/issues/1660 (Caused by passing a Parse.File into the LocalDataStore, which is unexpected)

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-SDK-JS/issues/1660).

### Issue Description

Saving a new Parse.Object with an unsaved Parse.File fails.

Closes: https://github.com/parse-community/Parse-SDK-JS/issues/1660

### Approach

Checks that the object being pinned to the localDataStore is a ParseObject.

### CODE REVIEW

**Comment by mtrezza** [comment]
Could you add test ?

**Comment by swittk** [comment]
I've added the test case

**Comment by dplewis** [comment]
@swittk I wrote the entire LocalDatastore code myself. Can you add an integration test to replicate the issue? I might have overlooked this

**Comment by swittk** [comment]
It's not a bug related to the LocalDatastore itself I think; but rather in the save() method of ParseObject which allows both `ParseObject | ParseFile`
https://github.com/parse-community/Parse-SDK-JS/blob/f27d043a50b18ea17d49aaf12a0ae1ad3bc68b8a/src/ParseObject.js#L2371

But then in the `then` clause after saving all; it pins them all into the LocalDatastore, which, as far as I understand, expects only Parse.Objects, so when a Parse.File goes in it crashes.
https://github.com/parse-community/Parse-SDK-JS/blob/f27d043a50b18ea17d49aaf12a0ae1ad3bc68b8a/src/ParseObject.js#L2494

I've spent the past few hours trying to get it to fail in the integration test, but it won't. The situation does occur in-browser with Parse SDK 3.5.1 and 4.0.0 though (Parse.File reaching LocalDatastore causing the crash).

The test cases in `ParseFile-test.js` (not the integration ones) do fail due to the same _getId() calls, though, with Parse.Files getting passed into the `save()` method of ParseObject. So right now I'm not sure what the exact difference in code paths are between `npm run test` and `npm run integration`, and can't really demonstrate it; but the error really is there in-browser.

(Here's an example of what my current hotfix to the problem looks like in node_modules)
Screen Shot 2566-01-28 at 11 56 01

**Comment by dplewis** [comment]
@swittk  I can verify that this is an issue. Can you add this failing test to your PR and rebase?

```
it('can save file to localDatastore', async () => {
    Parse.enableLocalDatastore();
    const file = new Parse.File('parse-js-sdk', [61, 170, 236, 120]);
    const object = new Parse.Object('TestObject');
    await object.pin();

    object.set('file', file);
    await object.save();

    query.fromLocalDatastore();
    const results = await query.find();

    const url = results[0].get('file').url();
    assert.equal(results.length, 1);
    assert.notEqual(url, undefined);
    assert.equal(url, file.url());
  });
```

File: src/ParseObject.js [hunk 1]
```diff
@@ -2492,8 +2492,11 @@ const DefaultController = {
             return Promise.reject(objectError);
           }
           for (const object of target) {
-            await localDatastore._updateLocalIdForObject(mapIdForPin[object.id], object);
-            await localDatastore._updateObjectIfPinned(object);
+            // Make sure that it is a ParseObject before updating it into the localDataStore
+            if(object instanceof ParseObject) {
```
dplewis: Can you add a space after the if?
```
if (object instanceof ParseObject) {
```

There is another place where `_updateLocalIdForObject` is used can you add the check there and update the tests if needed

dplewis (changes requested):
Everything looks good so far

src/ParseObject.js
swittk: I've edited to include the space.

Regarding the other place that references `_updateLocalIdForObject`, do you mean this?
https://github.com/parse-community/Parse-SDK-JS/blob/fce7f31f2db8293279ee77a6a2dd08d40e725c2a/src/ParseObject.js#L2529-L2534
If so it's ok to not check again I think, since `target instanceof ParseObject` has already been checked as part of its parent `if` condition.
https://github.com/parse-community/Parse-SDK-JS/blob/fce7f31f2db8293279ee77a6a2dd08d40e725c2a/src/ParseObject.js#L2504

dplewis: Ok an additional check isn't needed.

dplewis (approved):
LGTM! Just waiting for the CI to pass

swittk:
It seems like the CI hasn't run yet? ._.

mtrezza:
Now it is running

swittk:
I've tried running `npm run integration` on NodeJS 19.3.0 (same as the runner) and encountered the same issue.
It was solved by simply running `npm install` again after switching to node 19, and integration tests finished fine afterwards.
Screen Shot 2566-02-04 at 15 16 55

mtrezza:
The failing test in Node 19 is likely a flaky test, I'll re-run the CI and it should hopefully pass.

dplewis (approved):
LGTM! The flaky test will be fixed shortly.

mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [4.0.0-alpha.9](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.0.0-alpha.9)

parseplatformorg:
🎉 This change has been released in version [4.1.0-beta.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.1.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [4.1.0-alpha.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.1.0-alpha.1)

parseplatformorg:
🎉 This change has been released in version [4.1.0](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.1.0)

PR #1666, opened 2023-01-25
fix: Local datastore query with `containedIn` not working when field is an array

dplewis:

### Issue Description

Similar to https://github.com/parse-community/parse-server/pull/8128

Closes: Added additional test case for https://github.com/parse-community/parse-server/issues/8335

### TODOs before merging

- [x] Add tests
- [x] Add changes to documentation (guides, repository pages, in-code descriptions)

### DIFF
--- a/integration/test/ParseLiveQueryTest.js
+++ b/integration/test/ParseLiveQueryTest.js
@@ -257,38 +257,6 @@ describe('Parse LiveQuery', () => {
     await object.save();
   });
 
-  it('can handle containsAll query', async (done) => {
-    const object = new TestObject({ arrayField: [1, 2, 3] });
-    await object.save();
-
-    const query = new Parse.Query(TestObject);
-    query.containsAll('arrayField', [2, 3, 4]);
-    const subscription = await query.subscribe();
-
-    subscription.on('enter', object => {
-      expect(object.get('arrayField')).toEqual([1, 2, 3, 4]);
-      done();
-    });
-    object.addUnique('arrayField', 4);
-    await object.save();
-  });
-
-  it('can handle containedIn query', async (done) => {
-    const object = new TestObject({ arrayField: [1, 2, 3] });
-    await object.save();
-
-    const query = new Parse.Query(TestObject);
-    query.containedIn('arrayField', [4]);
-    const subscription = await query.subscribe();
-
-    subscription.on('enter', object => {
-      expect(object.get('arrayField')).toEqual([1, 2, 3, 4]);
-      done();
-    });
-    object.addUnique('arrayField', 4);
-    await object.save();
-  });
-
   it('live query can handle beforeConnect and beforeSubscribe errors', async () => {
     await reconfigureServer({
       cloud({ Cloud }) {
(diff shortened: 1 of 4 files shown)


mtrezza:
Thanks for the PR; could you rephrase the PR title to make it easier to understand in the changelog what this actually fixes?

The 2 references are related to LiveQuery - is this also LiveQuery related?

dplewis:
@mtrezza The LocalDatastore is internally code copied from LiveQuery meaning queries that LiveQuery supports the LocalDatastore should support too. Those are just additional test cases that doesn't exist server side.

mtrezza:
How would you best describe the fix in `src/OfflineQuery.js`?

> Local datastore not working properly for `Parse.Query.containedIn` where field is an array

Or something more specific?

dplewis:
> How would you best describe the fix in `src/OfflineQuery.js`?

It's the same fix as the `QueryTools.js` in the latest 5.4.0 release of the server

https://github.com/parse-community/parse-server/pull/8128/files#diff-13c52f9da4d37c7dc0bd3e9e845716de38a13e2a8632465ea167422762d5a9b6

dplewis:
> Or something more specific?

For non array fields you are comparing a single value against an array. This fix checks a array against an array recursively. As I write this I realized that `nonContainedIn` is broken and will return always return an object.

Without this fix we would have to rely on.

`haystack.indexOf(needle) > -1`

Array.indexOf(Array) will always return -1

mtrezza:
The other PR is called "fix: liveQuery with containedIn not working when object field is an array". So how about this:

> fix: Local datastore query with `containedIn` not working when field is an array

What do you conclude from the `nonContainedIn`? I haven't looked into the logic so far...

dplewis:
@mtrezza I write a few additional test cases, I also think there might be an issue with `containedBy` queries.

dplewis:
@mtrezza This is ready for review

mtrezza:
Re-running CI

I'm curious about these fails: https://github.com/parse-community/Parse-SDK-JS/actions/runs/4019619971/jobs/6906679057; it seems a connection error, do you have any idea what that could be since you've been looking at the flaky tests?

dplewis:
> it seems a connection error, do you have any idea what that could be since you've been looking at the flaky tests?

@mtrezza I suspect it might be an issue with `reconfigureServer`. `allowCustomObjectId` tests seem most common but I have seen `eventuallyQueue` tests also fail with his error. You can see where the first occurrence of the error is thrown.

parseplatformorg:
🎉 This change has been released in version [4.0.0-alpha.6](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.0.0-alpha.6)

parseplatformorg:
🎉 This change has been released in version [4.0.1-beta.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.0.1-beta.1)

parseplatformorg:
🎉 This change has been released in version [4.0.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.0.1)

PR #1682, opened 2023-01-30
fix: Request execution time keeps increasing over time when using `Parse.Object.extend`

### LINKED ISSUE #1683: Request execution time keeps increasing over time when using `Parse.Object.extend`
### New Issue Checklist

### Issue Description

There are numerous reports that Parse Server execution time increases over time:
- https://github.com/parse-community/parse-server/issues/7404
- https://github.com/parse-community/parse-server/issues/7036
- https://github.com/parse-community/parse-server/issues/6543
- https://github.com/parse-community/parse-server/issues/6300
- https://github.com/parse-community/parse-server/issues/5471
- https://community.parseplatform.org/t/parse-server-performance-loss-after-time

The issue may well be a JS SDK issue if the code causing the issue runs in Cloud Code using the Parse Server-integrated JS SDK.

### Steps to reproduce

See reported issues above.

### Actual Outcome

Server request execution times increase over time.

### Expected Outcome

Server request execution times should stay the same over time.

###  Environment

Server
- Parse Server version: various

dblythy:

### Issue Description

`new Parse.Object.extend("")` takes increasing execution time when ran in a loop, as it seems that the constructor somehow gets deep assigned each time, meaning that it is exponentially called.

Closes: https://github.com/parse-community/Parse-SDK-JS/issues/1683

### Approach

First commit shows the issue, with:

```js
 for (let i = 0; i < 100000; i++) {
      // eslint-disable-next-line
      const Parent = ParseObject.extend('Parent');
      // eslint-disable-next-line
      const parent = new Parent();
}
```

Timing out the CI, taking too long.

With these changes, this function only takes 38ms.

### TODOs before merging

- [ ] Add tests

### DIFF
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -121,6 +121,12 @@ class ParseObject {
       this.initialize.apply(this, arguments);
     }
 
+    if (this._initializers) {
+      for (const initializer of this._initializers) {
+        initializer.apply(this, arguments);
+      }
+    }
+
     let toSet = null;
     this._objCount = objectCount++;
     if (typeof className === 'string') {
@@ -1968,6 +1974,12 @@ class ParseObject {
         this.initialize.apply(this, arguments);
       }
 
+      if (this._initializers) {
+        for (const initializer of this._initializers) {
+          initializer.apply(this, arguments);
+        }
+      }
+
       if (attributes && typeof attributes === 'object') {
         if (!this.set(attributes || {}, options)) {
           throw new Error("Can't create an invalid Parse Object");
@@ -1991,6 +2003,15 @@ class ParseObject {
 
     if (protoProps) {
       for (const prop in protoProps) {
+        if (prop === 'initialize') {
+          Object.defineProperty(ParseObjectSubclass.prototype, '_initializers', {
+            value: [...(ParseObjectSubclass.prototype._initializers || []), protoProps[prop]],
+            enumerable: false,
+            writable: true,
+            configurable: true,
+          });
+          continue;
+        }
         if (prop !== 'className') {
           Object.defineProperty(ParseObjectSubclass.prototype, prop, {
             value: protoProps[prop],
@@ -2015,14 +2036,15 @@ class ParseObject {
       }
     }
 
-    ParseObjectSubclass.extend = function (name, protoProps, classProps) {
-      if (typeof name === 'string') {
-        return ParseObject.extend.call(ParseObjectSubclass, name, protoProps, classProps);
-      }
-      return ParseObject.extend.call(ParseObjectSubclass, adjustedClassName, name, protoProps);
-    };
-    ParseObjectSubclass.createWithoutData = ParseObject.createWithoutData;
-
+    if (!classMap[adjustedClassName]) {
+      ParseObjectSubclass.extend = function (name, protoProps, classProps) {
+        if (typeof name === 'string') {
+          return ParseObject.extend.call(ParseObjectSubclass, name, protoProps, classProps);
+        }
+        return ParseObject.extend.call(ParseObjectSubclass, adjustedClassName, name, protoProps);
+      };
+      ParseObjectSubclass.createWithoutData = ParseObject.createWithoutData;
+    }
     classMap[adjustedClassName] = ParseObjectSubclass;
     return ParseObjectSubclass;
   }
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -121,12 +121,6 @@ class ParseObject {
       this.initialize.apply(this, arguments);
     }
 
-    if (this._initializers) {
-      for (const initializer of this._initializers) {
-        initializer.apply(this, arguments);
-      }
-    }
-
     let toSet = null;
     this._objCount = objectCount++;
     if (typeof className === 'string') {
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -1959,10 +1959,8 @@ class ParseObject {
     let parentProto = ParseObject.prototype;
     if (this.hasOwnProperty('__super__') && this.__super__) {
       parentProto = this.prototype;
-    } else if (classMap[adjustedClassName]) {
-      parentProto = classMap[adjustedClassName].prototype;
     }
-    const ParseObjectSubclass = function (attributes, options) {
+    let ParseObjectSubclass = function (attributes, options) {
       this.className = adjustedClassName;
       this._objCount = objectCount++;
       // Enable legacy initializers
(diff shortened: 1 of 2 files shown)


dblythy:
Testing this on a Parse Server using `saveAll`, with 500 objects:

Pre-patch:

Screenshot 2023-01-30 at 4 22 12 pm

Post-patch:

Screenshot 2023-01-30 at 4 22 30 pm

dblythy:
Related: https://github.com/parse-community/parse-server/issues/7036

dblythy:
Test function:

```js
Parse.Cloud.define("test", async (request) => {
  const testObj = Parse.Object.extend("Temp");
  await new testObj().save();
  return true;
});
```

`autocannon -m POST -H "X-Parse-Application-Id: myAppId" -c 1000  http://localhost:1337/parse/functions/test`

Pre patch:

Screenshot 2023-01-30 at 4 36 00 pm

Post patch:

Screenshot 2023-01-30 at 4 36 47 pm

mtrezza:
That's an amazing improvement! Hats off!

Could this have an effect on the general performance of Parse Server if the JS SDK dependency of Parse Server is updated with this fix? I’m thinking if `extend` is used internally in Parse Server, even if it’s not used explicitly by the developer in their custom code.

dblythy:
> Does this need to get merged into Parse Server to have an effect when calling saveAll from a client 

No, unless there are cloud code triggers that uses `extend`

> I’m thinking if extend is used internally in Parse Server

I don't think extend is used at all internally

mtrezza:
I’ve updated my comment while you posted. If `extend` is used anywhere in the server code internally (haven’t checked yet), could it affect these parts of the code as well, essentially causing a leak?

dblythy:
I'm pretty sure the method isn't used internally. I might just double check that this change doesn't break the tests on Parse Server before we merge

mtrezza:
Sure, thanks for checking

dblythy:
Tests passing

mtrezza (approved):
Looks good! This will hopefully address these vague perf issue reports we're received over the years.

From the investigations this seems to go beyond a mere perf improvement and actually fix a bug. It seems that the execution time keeps increasing until the server slows down significantly. So I'll merge this as a bug fix.

parseplatformorg:
🎉 This change has been released in version [4.0.0-alpha.7](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.0.0-alpha.7)

parseplatformorg:
🎉 This change has been released in version [4.0.1-beta.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.0.1-beta.1)

parseplatformorg:
🎉 This change has been released in version [4.0.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.0.1)

vscorpio:
Issue seems to be still existing even in Parse 4.1.0? I just ran into the same issue when calling Parse.Object.extend in a loop and it gets progressively slower in time. I initially thought my queries are poorly written but then I found this.

dblythy:
@vscorpio can you share some pseudo-code, or write a failing test?

vscorpio:
@dblythy I have tested an whatever code creating rows gets slower in time. For example the following code:
```js
    // Create a new ticket purchase event similar to the one produced by the SolanaIndexer class.
    const TicketPurchaseEvent = Parse.Object.extend("EVENT_TicketPurchase");
    const ticketPurchaseEvent = new TicketPurchaseEvent();
    ticketPurchaseEvent.set("lottery_id", activeLottery?.get("lottery_id"));
    ticketPurchaseEvent.set(
      "ticket_number",
      latestTicket?.get("ticket_number")
        ? latestTicket?.get("ticket_number") + 1
        : 1
    );
    ticketPurchaseEvent.set("owner", owner);

    // Save the object to the database.
    await ticketPurchaseEvent.save(null, { useMasterKey: true });
```
Works perfect for the first 10 tries, but after the 10th iteration it gets progressively slower. CPU usage is never above 7%, Memory usage is negligible.

mtrezza:
I think it should be possible to write a test case for this, and test for example whether the execution time mean decreases more than x% over time.

dblythy:
I have written this test and have **not** observed the issue @vscorpio - is there any cloud code / server rate limit or slowdown to consider?

```js
it('execution time should not increase', async () => {

    const activeLottery = new Parse.Object('Lottery');
    activeLottery.set('lottery_id', 'lotteryID')
    const owner = new Parse.User();
    owner.set({username: 'abc', password: '123'});

    await Promise.all([
      activeLottery.save(), owner.signUp()
    ])

    let latestTicket = null;

    for (let i = 0; i < 10000; i++) {

      let start = new Date();
      const TicketPurchaseEvent = Parse.Object.extend("EVENT_TicketPurchase");
      const ticketPurchaseEvent = new TicketPurchaseEvent();
      ticketPurchaseEvent.set("lottery_id", activeLottery?.get("lottery_id"));
      ticketPurchaseEvent.set(
        "ticket_number",
        latestTicket?.get("ticket_number")
          ? latestTicket?.get("ticket_number") + 1
          : 1
      );
      ticketPurchaseEvent.set("owner", owner);

      // Save the object to the database.
      await ticketPurchaseEvent.save(null, { useMasterKey: true });
      latestTicket = ticketPurchaseEvent;
      const end = new Date();
      console.log('duration: ', end.getTime() - start.getTime());
    }
  });
```

mtrezza:
Did you verify that this test would have failed pre-https://github.com/parse-community/Parse-SDK-JS/pull/1682? It's not part of the PR so if it hasn't been added to the tests yet and we can verify the test works it would be good to add it.

There are also some caveats regarding code execution time measurement, so maybe rewrite the test using the Node built-in [perf API](https://nodejs.org/dist/latest-v18.x/docs/api/perf_hooks.html#performance-measurement-apis) and test the execution time variance between the mean of the first few and last few runs. That would also be a good template for other perf tests in the future.

dblythy:
Testing with 50,000 requests and averaging execution times per 10,000 requests (using `performance.now()`:

Parse JS SDK 4.0.0 (pre #1682):

0-9999 average duration: 7.8ms
10000-19999 average duration: 25.1ms
20000-29999 average duration: 43.1ms
30000-39999 average duration: 90.9ms
40000-49999 average duration: 148.8ms

Parse JS SDK 4.1.0 (post #1682):

0-9999 average duration: 1.9ms
10000-19999 average duration: 1.9ms
20000-29999 average duration: 2.1ms
30000-39999 average duration: 2.05ms
40000-49999 average duration: 2.1ms

mtrezza:
Nice, I think you can just calculate the standard deviation of all runs (not of batches) for the test to pass or fail:

pre-𝜎 = 51.02825
post-𝜎 = 0.09165

So the test could fail if 𝜎 >= 0.1
If that turns out to create a flaky test, we can increase it.

PR #1719, opened 2023-02-01
ci: Remove MongoDB runner

### LINKED ISSUE #1730: Node 14 CI takes long to finish
### New Issue Checklist

### Issue Description

- The mongodb runner dependency takes 3x as long to download in the Node 14 environment compared to other environments of the CI.
- mongodb runner is a prod dependency (for unknown reasons) but it should be a dev dependency

The issue doesn't seem to occur in the Parse Server CI, so maybe the CI is set up differently here; needs to be looked at.

See https://github.com/parse-community/Parse-SDK-JS/actions/runs/4089983465/jobs/7053083540

It should finish in a comparable time to the other Node environments.

dplewis:
## Pull Request

## Issue

The mongodb runner has a dependency that has to be downloaded over ssh and causes the CI to take ~17 minutes on Node 14

Closes https://github.com/parse-community/Parse-SDK-JS/issues/1730

## Approach

* Update CI dependencies
* Improve node cache
* Add mongodb github action
* Saves 14 minutes on CI time


mtrezza:
- If we replace mongodb runner with  a GH action won't it be more difficult for developers to develop and test a PR locally for various MongoDB version?
- Could we just configure git to define the protocol that git uses to download that specific dependency?

dplewis:
> If we replace mongodb runner with a GH action won't it be more difficult for developers to develop and test a PR locally for various MongoDB version?

Thats what the contribution guide is for, they can install mongodb-runner themselves, use MongoDB CLI or any method they choose for testing.

> Could we just configure git to define the protocol that git uses to download that specific dependency?

Seems like extra work for a package that isn't a dev dependency.

mtrezza:
MongoDB runner is only a dev dependency, so this issue doesn't have any effect on production deployments, right?

How come we don't see this issue with parse-server? Or does it occur there as well?

Maybe we could avoid adding a barrier for contributors and instead fix this temporary issue (we'll remove Node 14 support in 11 months) with a git config.

Configuring git shouldn't be a lot of work. We can instruct git to use a specific protocol for a specific dependency with a 1 liner. What is the problematic URL and what URL should the npm installer use instead?

dplewis:
> MongoDB runner is only a dev dependency

It isn't, check package.json. You must be thinking about the server.

Should we add it as a dev dependency it would add 5 minutes to the CI?

mtrezza:
Do you know why mongodb runner isn't a dev dependency? Does the SDK actually need it?

dplewis:
The SDK doesn't need it. I forgot why it isn't a dev dependency, it's always been like that. We have had a lot of issues with mongodb-runner over the years.

mtrezza:
Then we can fix 2 issues at once here. Let's move it to dev and add a git config rule. 

And down the road we can look for a replacement for the runner. I still think it would be good to keep it easier for contributors to test against different mongoDB versions. That's why we've added the package scripts in parse server. Maybe we could add them here too.

dplewis:
I think updating the Contributing Guide would solve all your concerns for the developers. I'm against adding unmaintained packages here.

> Maybe we could add them here too.

I'm against that as we have flaky tests. Maybe once we figure out those I would consider it.

dplewis:
@mtrezza I actually tested moving as a dev dependency before this PR and `npm ci` took 7 minutes. I don't think developer will want to wait that long when testing locally.

https://github.com/dplewis/Parse-SDK-JS/actions/runs/4058503220/jobs/6986789849

Can we merge this?

.github/workflows/ci.yml
```diff
@@ -37,21 +37,17 @@ jobs:
...
+      uses: actions/setup-node@v3
       with:
         node-version: ${{ matrix.NODE_VERSION }}
-    - name: Cache Node.js modules
-      uses: actions/cache@v2
+        cache: 'npm'
+    - name: Start MongoDB
+      uses: supercharge/mongodb-github-action@1.8.0
       with:
-        path: ~/.npm
```
mtrezza, line 48: See https://github.com/parse-community/parse-server/issues/7614#issuecomment-945178352

I think that still holds true?

dplewis: Technically `actions/cache` used here creates 1 cache for all node versions and hasn’t caused a problem. I'll revert it back

dplewis:
@mtrezza Can you rerun the CI? I want to make sure the cache works

Screen Shot 2023-02-02 at 11 46 58 AM

dplewis:
@mtrezza i think this is good to merge

mtrezza:
Still not sure about removing mongodb runner for contributors to fix a temp issue with Node 14. I'll look into the git URL fix.

Also the dev dependency situation seems still unclear, why is the runner not a dev dependency if it's only needed in dev.

dplewis:
> Still not sure about removing mongodb runner for contributors to fix a temp issue with Node 14. I'll look into the git URL fix.

I don’t see this being an issue for contributors. If it is an issue we can move this unmaintained package to dev dependency. There are many ways to install mongodb locally.

mtrezza:
What would be the steps required for a contributor to run a test with 3 different versions of MongoDB?

In Parse Server it's currently:

```
npm run test:mongodb:4.2.19
npm run test:mongodb:5.3.2
npm run test:mongodb:6.0.2
```

It is strange that mongodb runner works fine for Node 14 in the Parse Server CI, but not here.

mtrezza:
I'v opened https://github.com/parse-community/Parse-SDK-JS/issues/1730 since this PR doesn't have a ref to an issue and has been closed.

PR #1729, opened 2023-02-04
feat: Upgrade Node Package Manager lock file `package-lock.json` to version 2

### LINKED ISSUE #1728: Regenerate package-lock.json as v2

With the removal of node 12 we can safely upgrade to v2. This will improve CI times significantly. The has already been implemented on version 6 of the server.

dplewis:
## Pull Request

## Issue

I want to see what the effects are against the CI.

Closes: https://github.com/parse-community/Parse-SDK-JS/issues/1728

## Approach

* Update package-lockfile to V2
* Update lockfile checker in CI
* Improve CI caching to individual cache per node version
* Move mongodb runner from CI to Dev Dependencies
* Update semantic-release packages


dplewis:
@mtrezza Looks like mongodb-runner and node 14 are working well here.

mtrezza:
> Looks like mongodb-runner and node 14 are working well here.

Great to hear, that's why we didn't see the issue in the Parse Server CI. So we can keep mongodb-runner. Btw, they made a recent release 3 days ago (albeit just a tag), so there is some maintenance going on it seems. What's left is to move it to the dev dependencies, but we'll do that in a separate PR related to #1730 since we want it to have a distinct changelog entry.

package.json
```diff
@@ -47,9 +47,9 @@
     "@babel/preset-react": "7.10.1",
     "@parse/minami": "1.0.0",
     "@saithodev/semantic-release-backmerge": "2.1.3",
-    "@semantic-release/changelog": "5.0.1",
```
mtrezza, line 50: Could you please only do the package-lock v2 upgrade in this PR and remove all other changes, except maybe the "actions/checkout@v3" and "${{ matrix.NODE_VERSION }}-${"?

- We do dependency upgrades in separate commits for easier debugging.
- We have to research before upgrade semantic-release as it's not tested in the CI.
- We want to keep the workflow files conceptually coherent across repos; if we replace the CI step with a `mongo:start` npm script here, we also need to do this in other repo, all at once; so we'd do this in distinct PRs. We've largely cleaned up the mess of every repo having it's own flavor of a CI workflow and npm scripts over the last year.

mtrezza (changes requested):

package.json
dplewis: > What's left is to move it to the dev dependencies mongodb-runner

I've already moved it to dev dependency. The combination of V2 + Caching makes it possible it seems

> We do dependency upgrades in separate commits for easier debugging.

This is true but cause issues. V2 is restrictive on peer dependencies.
https://github.com/parse-community/Parse-SDK-JS/actions/runs/4089591152/jobs/7052405616

Most of our packages are grouped together and outdated.

* babel-jest + jest
* eslint + flowtype + jsdocs
* babel/runtime-corejs + core js + babel/core + babel/plugins
* semantic-release

> We have to research before upgrade semantic-release as it's not tested in the CI.

You downgraded semantic-release instead of bumping its peer dependencies. 
Can you research it now?

> We want to keep the workflow files conceptually coherent across repos

The current CI has been nearly the same for 4 years and was originally copied from the PHP-SDK which still uses Travis instead of Github Action. I would also like consistency in the CI.

mtrezza: Please revert:
- any semantic release related changes; we don't need to research this, that was just a general remark to explain why we cannot just upgrade it, see https://github.com/parse-community/Parse-SDK-JS/pull/1706#issuecomment-1410441296
- the mongodb runner scripts; we should use the same script names that we already have in Parse Server to make it more intuitive for developers; if they should be different then we'd also need to change them in PS and should discuss that separately why we want to change them

We can keep the dependency upgrades that are necessary for the lockfile upgrade, that makes sense.

dplewis: > We can keep the dependency upgrades that are necessary for the lockfile upgrade, that makes sense.

The semantic-release changes are necessary for the V2 upgrade. As shown in the previous failing job. Either we keep the current changes here or downgrade semantic-release even further from 19.0.5

Feel free to tackle it in a separate PR, I'll merge it in

https://github.com/parse-community/Parse-SDK-JS/actions/runs/4089591152/jobs/7052405616

mtrezza: OK, let me take a quick look the semantic release versions here

mtrezza: These are the versions used in Parse Server which is on lock-file v2:

```
"@semantic-release/changelog": "5.0.1",
"@semantic-release/commit-analyzer": "8.0.1",
"@semantic-release/git": "9.0.0",
"@semantic-release/github": "7.2.3",
"@semantic-release/npm": "7.1.3",
"@semantic-release/release-notes-generator": "9.0.3",
 "semantic-release": "17.4.6",
```

You can probably just use the versions we're using in Parse Server to make this work? 

I see that some dependencies here are even higher here, like `"semantic-release": "19.0.5",`, so maybe for v2 we need to do a downgrade here and in a separate PR look into making sure it all plays together. Alternative would be to look at the 7 changelogs and see whether it all plays together, keeing in mind that `semantic-release` needs to stay on <20.

dplewis: @mtrezza You missed the most important one, this SDK uses `19.0.5`
```
"semantic-release": "17.4.6",
```

mtrezza: You were reading my comment faster than I was able to edit it :-) So I've looked through all the changelogs now and I guess we can merge this. If anything happens during release I'll deal with it.

package.json
```diff
@@ -123,7 +127,7 @@
     ]
   },
   "engines": {
-    "node": ">=14.21.0 <17 || >=18 <19"
```
mtrezza: Is this change required for this PR?

dplewis: Technically we do support node 19+ right now, wanted to get rid of these warnings

```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'parse@4.0.0',
npm WARN EBADENGINE   required: { node: '>=14.21.0 <17 || >=18 <19' },
npm WARN EBADENGINE   current: { node: 'v19.3.0', npm: '9.2.0' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'parse-server@6.0.0-alpha.30',
npm WARN EBADENGINE   required: { node: '>=14.21.0 <17 || >=18 <19' },
npm WARN EBADENGINE   current: { node: 'v19.3.0', npm: '9.2.0' }
npm WARN EBADENGINE }
```

mtrezza: Got it, we need to do this in a separate PR please. This merits a distinct changelog entry, it's a bug fix I'd say since we forgot to change this from <19 to <20 (we need to keep an upper range) and I think there is even a strict setting in npm where the installer will abort if there is such a conflict. Good you found that!

mtrezza (approved):
Looks good! Get's get this merged if the CI passes...

package.json
mtrezza: Opened https://github.com/parse-community/Parse-SDK-JS/issues/1731 to track this

mtrezza:
Uh, there is a conflict after merging alpha, could you take a look?

mtrezza (approved):
Looks good!

mtrezza:
This PR seems to be done, can we merge? Any other issues please in separate PRs.

dplewis:
Its good to merge

mtrezza:
Glad to see this Node 14 issue finally fixed

parseplatformorg:
🎉 This change has been released in version [4.0.0-alpha.10](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.0.0-alpha.10)

parseplatformorg:
🎉 This change has been released in version [4.1.0-beta.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.1.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [4.1.0-alpha.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.1.0-alpha.1)

parseplatformorg:
🎉 This change has been released in version [4.1.0](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.1.0)

PR #1803, opened 2023-03-02
feat: Allow logout with invalid session token

### LINKED ISSUE #307: Improving Invalid Session Handling
According to the documentation on [Handling Invalid Session Token Error](https://parse.com/docs/js/guide#sessions-handling-invalid-session-token-error), the appropriate resolution is to attach an error handler to every call out to the Parse server.

Applying this change to an existing aplication is a bit of a chore in terms of finding all of the appropriate places to add the error handler. It also places a burden on developers to remember to include this handler whenever writing new code.

I can think of a couple ways to address this:

- ### Always log the user out on INVALID_SESSION_TOKEN
  We build in to the framework a special case for this error that directly invokes `Parse.User.logOut`. This has the benefit of immediately changing the state of `Parse.User.current` to reflect the new information, which may be sufficient for some applications to understand the state change. The downside is that this may be a breaking change in the SDK for some applications that aren't expecting it.
  
- ### Allow registration of a global error handling callback
  Imagine initializing the SDK could look something like:
  ```js
  Parse.initialize(APP_ID);
  Parse.serverErrorHandler(function(err) {
    if (err.code === Parse.Error.INVALID_SESSION_TOKEN) {
      Parse.User.logOut();
    }
  });
  ```
  When this handler is present, it could be automatically invoked whenever we receive an error response from the Parse server. This has the downside of being extra configuration, but is extremely flexible in how it could be used. It also doesn't present any possibility of being an API breaking change, as the behavior would remain the same if the global handler is not registered.
  
How would folks feel about a pull request for either of these strategies? Are there other options to improve this that folks would rather see?

dblythy:
## Pull Request

## Issue

Currently, if trying to logout with an invalid session token, you will need to catch the error, and call logout again. It can be a bit tedious.

Closes: #307

## Approach

Improves clearing session by allowing `Parse.User.logOut({ clearSession: true })`, which resolves even if invalid session token is returned.

## Tasks

- [x] Add tests

## Summary by CodeRabbit

- **New Features**
	- Added an optional setting to the logout process that allows users to specify whether to clear their session if the session token is invalid.

- **Bug Fixes**
	- Improved handling of invalid session tokens during logout, preventing unnecessary errors when the new session clearing option is enabled.

- **Tests**
	- Introduced new test cases to verify logout behavior with the session clearing option and invalid session tokens.

- **Documentation**
	- Updated documentation to reflect the new logout option and its behavior.


mtrezza:
> Currently, if trying to logout with an invalid session token, you will need to catch the error, and call logout again. 

What would be the difference between the first and second logout request? What makes the first fail and the second succeed?

mtrezza:
@dblythy Friendly ping regarding the previous question. If the 1st logout attempt fails, why does the 2nd attempt (after catching the error) succeed?

dplewis:
@dblythy We should add this to the server side like https://github.com/parse-community/parse-server/pull/8722 so that it can work for all SDK's

@mtrezza I have no idea why there is a need to call `Parse.User.logOut` twice when there is a invalid session error to log out. I've been running into this issue a lot while fixing sessionToken invalid issues.

mtrezza:
Could you please open a separate issue for this?

dblythy:
@dplewis see https://github.com/parse-community/parse-server/issues/7277#issuecomment-801953151

kishanio:
are we still tracking this?

dplewis:
@dblythy Nice job! Can you fix the merge conflicts?

dplewis:
@dblythy @mtrezza This is ready to merge

src/ParseUser.ts
```diff
@@ -812,11 +812,13 @@ class ParseUser<T extends Attributes = Attributes> extends ParseObject<T> {
    * <code>current</code> will return <code>null</code>.
    *
    * @param {object} options
+   * @param {boolean} [options.clearSession] If true, the session token will be
```
mtrezza, line 815: I think this needs a better explanation. It's now a rather technical description but not so clear what the purpose of the option is and why / when one should use it.

dplewis: What do you recommend we change it to?

mtrezza: I don't have a suggestion because I don't understand what the option exactly is for, from a practical perspective.

See https://github.com/parse-community/Parse-SDK-JS/pull/1803#issuecomment-2855565390.

mtrezza:
I understand that with the new option this PR introduces, the session token will be removed if the server returns that it's invalid, but:

- Why would I set this for a specific method instead of as a general option for the whole SDK, i.e. why is this not relevant for other API calls as well?
- If this is an option, why would I want to maintain an invalid session token at all?
- How are other Parse SDKs handling this for comparison?

parseplatformorg:
### :tada: **Snyk checks have passed. No issues have been found so far.**

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/acinader/pr-checks/82be2741-861f-417e-a27f-db147aac220c?refs=ccbe7319a62f4a54a648caa62bc04d71e6413172&source=prChecksComment)

dplewis:
> How are other Parse SDKs handling this for comparison?

The PHP SDK just ignores it as nobody really cares.

https://github.com/parse-community/parse-php-sdk/blob/081690dd822629933b27b8fa6b8adf143828a48d/src/Parse/ParseUser.php#L499C5-L511C6
```
public static function logOut()
    {
        $user = static::getCurrentUser();
        if ($user) {
            try {
                ParseClient::_request('POST', 'logout', $user->getSessionToken());
            } catch (ParseException $ex) {
                // If this fails, we're going to ignore it.
            }
            static::$currentUser = null;
        }
        ParseClient::getStorage()->remove('user');
    }
```

> Why would I set this for a specific method instead of as a general option for the whole SDK, i.e. why is this not relevant for other API calls as well?

If any other calls get an invalid session token error the next step logical step would be to logout. Then logout throws the same error and I have to logout again. Luckily the second logout doesn't throw this error or it would be an infinite loop.

mtrezza:
> The PHP SDK just ignores it as nobody really cares.

That's what I was thinking. I don't think we need that option. The way the PHP SDK handles it makes sense. I also can't think of a reasonable customization in that regard that would justify introducing an option here. Once the server determines that a token is invalid, it won't become valid ever again as they are designed to be ephemeral.

dplewis:
One thing the PHP SDK and Swift SDK does is clear the current user which I think the current SDK doesn't do if an invalid sessionToken is thrown.

mtrezza:
I think that makes sense as well. @dblythy Would you want to adapt this? Or if anyone else wants to pick up this PR...

Issue #1362, opened 2021-05-13
TypeError when bundling latest Parse JS SDK using VITE (the vue bundler)


tremendus:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-JS/issues?q=is%3Aissue).
- [-] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse JS SDK](https://github.com/parse-community/Parse-SDK-JS/releases). 

### Issue Description

When bundling Parse JS SDK client through VITE - vue's new bundler, based on snowpack - compilation fails due to an error caused in parse.

### Steps to reproduce

- Check out repo: https://github.com/tremendus/vite-vue3-tailwind-starter-parse-issue
- Branch -> main
- run `npm i`
- open browser if it doesn't automatically http://localhost:3000
- open developer console - see error

NOTE: parse lib is imported in src/App.vue

### Actual Outcome

```
 Uncaught TypeError: Super expression must either be null or a function
    _inherits Babel
    Subscription LiveQuerySubscription.js:149
    js LiveQuerySubscription.js:197
    __require chunk-BIJEITIM.js:6
    js LiveQueryClient.js:65
    __require chunk-BIJEITIM.js:6
    js ParseLiveQuery.js:25
    __require chunk-BIJEITIM.js:6
    js Parse.js:286
    __require chunk-BIJEITIM.js:6
    js index.js:1
    __require chunk-BIJEITIM.js:6
    <anonymous> parse:1
```

This is the code at 149: 

```
var Subscription = /*#__PURE__*/function (_EventEmitter) { << 149
  (0, _inherits2.default)(Subscription, _EventEmitter);

  var _super = _createSuper(Subscription);
  // ....
```

### Expected Outcome

I expected Parse to be bundled without error

###  Environment

`"parse": "^3.2.0",`
- see package.json in repos for other dependencies.

Server
- Parse Server version: - Not applicable

Database
- System (MongoDB or Postgres): Not applicable

Client
- Parse JS SDK version: "parse": "^3.2.0",
- localhost on node 15.11 M1 using "vite": "^2.3.0"

### Logs

No logs - see error in console

[...]

PR #1954, opened 2023-06-23
feat: Add typescript support

### LINKED ISSUE #1950: Add TypeScript support

### Current Limitation

The code base does not have native TypeScript support but relies on 3rd part TypeScript definitions. These 3rd party definitions are often inconsistent, leading to complaints from developers and support cases being opened in Parse Platform resources, even though Parse Platform is not responsible for managing these definitions.

There have been discussions and attempts in the past about strategies, sometimes trying to use the existing (not 100% coherent) JSDoc annotation as a basis to transition to TypeScript. As usual, without a committed lead pushing the topic forward, such a large scale change becomes stale over time and don't get implemented.

- https://github.com/parse-community/Parse-SDK-JS/pull/388
- https://github.com/parse-community/Parse-SDK-JS/pull/1390
- https://github.com/parse-community/parse-server/issues/7334

Develop a strategy to gradually implement TypeScript support.

It would be beneficial to develop for a strategy where a TypeScript migration can be implemented gradually, in tiny PRs and over time. This makes the presence of a lead less relevant, as the efforts is split up into manageable tasks that be picked up by the community over time. That could mean that we won't take advantage of the existing JSDoc annotations, requiring more work in total, but that seems preferable compared to a large effort that never comes to fruition.

This issue closes with a strategy for gradual TypeScript implementation and a first, small PR that begins the implementation. We can then start to give out bounties for individual scopes to facilitate the migration to TypeScript in the long run.

dblythy:
## Pull Request

## Issue

Currently Parse JS SDK relies on external typings

Closes: #1950

## Approach

Copies DefinitelyTyped `types` into this repo

## Tasks

- [ ] test types


dblythy:
@sadortun could you have a look at the approach here?

.github/workflows/ci.yml
```diff
@@ -51,6 +51,7 @@ jobs:
             ${{ runner.os }}-node-${{ matrix.NODE_VERSION }}-
     - run: npm ci
     - run: npm run lint
+    - run: npm run test:types
```
mtrezza: Do we need to run this for various Node and DB environments of can we move this into its a separate job `check-typescript`?

types/index.d.ts
mtrezza, line 1: See https://github.com/parse-community/Parse-SDK-JS/issues/1950#issuecomment-1604017156

mtrezza:
What's the strategy here? Shouldn't types be part of the code comment of the respective function rather than in a separate files? If the types are defined separately, they need to be maintained separately from the code itself or they'll deviate, same issue as we're having now with 3rd party types, right?

Is there a way we can actually add types gradually to the code comments (replacing JSDoc) without requiring major PRs?

types/index.d.ts
sadortun: Is this `.d.ts` auto-generated?

mtrezza: It's copied from DefinitelyTyped (as I understand the PR description)

dblythy:
At the moment the `d.ts` files are just copied. I wonder if the better approach would be to convert this whole repo to typescript, perhaps it would resolve #1362

mtrezza:
I agree, but how can we do that gradually in smaller PRs, instead of in 1 large PR? Also, we'd need to ensure that the API docs are still correctly generated, currently using JSDoc.

I think if we could start a simple PR with just converting 1 method (or 1 class?) to TS, while demonstrating the API docs generation still works, we'd have really solved what has been holding us back from adding TS support and can close https://github.com/parse-community/Parse-SDK-JS/issues/1950. It will be easy to find others to convert the rest, following such an example.

sadortun:
> gradually in smaller PRs

That's the way to go! The repo already have a great pipeline, adding TS and ensuring the pipeline and 3rd party tools still work would be a huge step forward.

Converting the whole repo will take time (maybe a year?), and it'll be easier to troubleshoot problems with multiple small PRs.

mtrezza:
@dblythy Would you want to:
- rewrite the PR and try to convert just 1 method of a class to TS
- see if the JSDoc generation still works, or if we need a new tool ([TSDoc](https://tsdoc.org)?)
- add a CI check `check-typescript` which should fail on type conflicts, so we know whether we can merge these tiny PRs in the future to convert to TS

And discard the 3rd party TS file - it's not attached to the methods anyway and partly outdated, but we can keep a reference in the GitHub issue, as it may help others as a guidelines when converting to TS in their PRs. 

I've also added a bounty to #1950 to hopefully get this going.

dblythy:
I've updated it so the `d.ts` file is auto-generated

I think the problem is that as types are imported across the project by `Flow`, all files will need to be updated to typescript in one go:

Screenshot 2023-06-30 at 4 10 41 pm

Screenshot 2023-06-30 at 7 03 19 pm

I've updated a few classes but the problem is they keep importing types from a file that isn't `TS`. And then when I change that file to `.ts`, the issue deepens.

The approach of this PR has been to:

- rename files to `.ts`
- Fix any `ts` issues in the file
- Auto-generate definition files

It seems to be working pretty well.

mtrezza:
You wrote "the issue deepens" and then "It seems to be working pretty well"; so I don't understand whether there is currently an obstacle or whether you've found an approach that works?


> rewrite the PR and try to convert just 1 method of a class to TS

  This PR contains a lot of changes. I wonder were these changes necessary in order to convert 1 method and fix subsequent TS errors, or were these just unrelated, additional TS conversions that would not have been required?

> see if the JSDoc generation still works, or if we need a new tool ([TSDoc](https://tsdoc.org/)?)

Looking at this change:

```
class Subscription extends EventEmitter {
  /*
   * @param {string} id - subscription id
   * @param {string} query - query to subscribe to
   * @param {string} sessionToken - optional session token
   */
  id: string;
  query: ParseQuery;
  sessionToken: string;
  subscribePromise: resolvingPromise<any>
  unsubscribePromise: resolvingPromise<any>
  subscribed: boolean;
  emit: any;
  constructor(id: string, query: ParseQuery, sessionToken: string) {
```

It seems that we have two (conflicting) type definitions for the same var; in JSDoc and in TS. Do we need to keep the `@param` and `@returns` entries in JSDoc to generate the API docs, or could we remove them in lieu of the added TS types by switching to TSDoc?

dblythy:
It works well in the sense of generating correct types. But the types aren’t usable if the code imports types from a non-ts file (which causes the typescript error). 

A fair few of the changes are auto-generated definition files. 90% of the changes are just file renames. 

This PR bloated out as I’m conscious that if we offer supported, incomplete types, a lot of the intellesense + current typescript support will disappear. 

Most of the changes are minor type corrections, as typescript detects issues as soon as a file is renamed. 

I’d assume conflicting types in JSDocs / typescript means that the docs are wrong and need to be updated as well.

mtrezza:
> the types aren’t usable if the code imports types from a non-ts file (which causes the typescript error)

That's exactly what we want to solve with this PR. Is there a way to convert only one method of a class to TS while maybe adding some annotations (like lint annotations) that could silence these warnings so that a CI types check only checks what is already converted to TS.

> This PR bloated out as I’m conscious that if we offer supported, incomplete types, a lot of the intellesense + current typescript support will disappear.

We'd have to naturally accept that during the transition. This PR is to verify the concept, without making a lot of conversions, because we want to explicitly know if it's possible to convert only a tiny part (like 1 method of a class) with a working CI type check and the API docs being generated. I would maybe reduce the type conversions in this PR (or open an additional new one as PoC and merge this one later?), because that will be the sort of PRs that contributors should be able to submit.


> I’d assume conflicting types in JSDocs / typescript means that the docs are wrong and need to be updated as well.

Ideally we could which to TSDoc already if it uses TS where specified and fall back to taking types from JSDoc annotations where no TS types are specified. We'd need to test out if that works.

dplewis:
@dblythy In your tsconfig.json why not add `allowJS: true`?  It will allow mixing ts and js files, it should also only type check ts files. Once all the files are statically typed then we can generate the .d.ts files. Probably fix those `require` modules too.

dplewis:
@dblythy @mtrezza  I'd recommend using `allowJS: true` in your tsconfig.json. This will allow both js and ts files. Type checking should only be done on TS files. Once all files are statically typed properly (and flow is removed) then we can generate the .d.ts files in a separate PR to remove DefinitelyTyped.

dplewis:
@dblythy What is the status of this PR? Do you mind if I contribute to it?

dblythy:
This PR has gone stale as large wholesale changes are not preferred for this repo, although, I'm not sure how it will be possible to offer non-broken inbuilt types _without_ large changes (see #1985)

dplewis:
Closing via https://github.com/parse-community/Parse-SDK-JS/pull/1985

PR #1985, opened 2023-07-28
refactor: Add TypeScript support

### LINKED ISSUE #1950: Add TypeScript support

dblythy:
## Pull Request

## Issue

Closes: #1950

## Tasks

- [ ] Add tests


dblythy:
This PR uses `allowJS: true` to auto-generate `d.ts` files, however, quite a few issues occur when building types:

Screenshot 2023-07-28 at 1 00 00 pm

And a few more when running `test:types`:

Screenshot 2023-07-28 at 1 00 34 pm

dblythy:
@dplewis I have added you to my repo if you would like to contribute. I'm not sure how we can resolve the typescript issues _without_ making large changes to the codebase. The issue seems to be that one file depends on invalid ts, and when that file is fixed, it depends on another file with invalid ts, and so forth. @mtrezza?

dplewis:
> without making large changes to the codebase

@dblythy I have no problem reviewing large changes. This PR is just syntactic sugar without any functionality.

mtrezza:
@dplewis I think dblythy was referring to the fact that we want to add TypeScript support incrementally without *requiring* large parts of the codebase to be converted at once, for the reasons described in https://github.com/parse-community/Parse-SDK-JS/issues/1950.

dplewis:
@dblythy If you ignore `test:types`  what else would be needed to merge this PR as is? As long as the project transcompiles and the tests pass I would count this as Typescript support incrementally.

@mtrezza Thoughts?

mtrezza:
Can we also have a TS test in the CI that ensures that the parts of the code that are already converted to TS are type-correct? Otherwise we'll have a hard time reviewing the incremental TS PRs.

dplewis:
@dblythy @mtrezza For this to work you need to fix the test suite and treat it like the unit tests. Bypass the `Parse` module and test against the files themselves. Luckly we already have the test suite from definitely typed

```
import ParseObject from 'ParseObject';
class GameScore extends ParseObject {
  constructor(options?: any) {
      super('GameScore', options);
  }
}
```

instead of 

```
import Parse from 'Parse';
class GameScore extends Parse.Object {
  constructor(options?: any) {
      super('GameScore', options);
  }
}
```

mtrezza:
I was referring to the CI job that currently fails, the `"test:types": "dtslint --expectOnly types",` npm script. How can we get `dtslint` to pass?

dplewis:
> How can we get dtslint to pass?

Thats what I'm talking about, you do know how dtslint works right? You build a test file against your typescript types. Since the test file isn't correctly set up it fails. `src/Parse.ts` is broken so you can't test against it until the entire project is typed correctly.

dplewis:
@dblythy I only changed Parse.ts and ParseSession.ts. Here are the steps if you want to convert other files.

* Change file extension to `.ts`
* Remove @flow from file if it exists
* Add file to `./tsconfig.json`
* `npm run build:types`
* Fix types
* Add / enable tests  in `types/test.ts`
* `npm run test:types`
* Compare types to [@types/parse](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b23a36e669fa127d1035e22ca93faab85b98e49f/types/parse/index.d.ts#L11)

dplewis:
> src/Parse.ts is broken so you can't test against it until the entire project is typed correctly

I stand corrected, we can test against Parse.ts but first we need to fix these. We should also consider moving to es6.

Screenshot 2023-07-28 at 7 10 15 PM

dplewis:
@dblythy I used allowJs to generate all the typings available with a ts-no-check to remove warnings. Regenerating the typings would remove the no-check, I set it up to prevent that and allow for incremental regeneration by adding files to tsconfig. 

index.d.ts shouldn’t be changed for now as every release will have some typing like DefinitelyTyped

dplewis:
@mtrezza I think this is ready for review

mtrezza:
Nice, the checks pass. @dblythy What do you think?

So [these](https://github.com/parse-community/Parse-SDK-JS/pull/1985#issuecomment-1656471016) are the instructions for incrementally convert to TS? If so, I'll copy them to https://github.com/parse-community/Parse-SDK-JS/issues/1950, so we have that documented for contributors.

> index.d.ts shouldn’t be changed for now as every release will have some typing like DefinitelyTyped

What does that mean? Is there anything new we need to consider when reviewing PRs? If so, could you amend the [instructions](https://github.com/parse-community/Parse-SDK-JS/pull/1985#issuecomment-1656471016) with what not to do, so it also serves as guideline for reviewers?

dplewis:
> What does that mean?

To be honest I don't use typescript and never published a typescript package. I don't think users will have to install `npm install @types/parse` after this code gets release. This is my assumption.

Instructions have been updated.

mtrezza:
That's good point. We know that `@types/parse` is currently more complete than our gradual conversion. For developers that probably means an immediate loss of types and IDE auto-complete, if they don't use or cannot use `@types/parse` anymore. What happens if a package installs the Parse JS SDK with our partial TS conversion and also has the `@types/parse` package installed? Will one overrule the other? Will there be conflicts? Does `@types/parse` have to be uninstalled? That's info we should include in the change log when merging this.

dplewis:
We would remove the package from DefinitelyTyped to avoid confusion.

mtrezza:
Would it make sense to accept TS conversion PRs but not publish the type files on NPM just yet, i.e. not include them in the published npm package, only keep them in the GitHub repo? We could wait until we have reached a good level of TS conversion so that developers won't loose the current level of type coverage so drastically.

dplewis:
>  so that developers won't loose the current level of type coverage so drastically

Since index.d.ts is copied directly from DefinitelyTyped no coverage is loss. As far as publishing to npm, we should plan for the next major release of the SDK

dplewis:
Can we get this merged?

mtrezza:
I think this needs a closer examination before merge.

> Since index.d.ts is copied directly from DefinitelyTyped no coverage is loss.

Don't the other `.d.ts` files contain types as well? For example `ParseQuery.d.ts` contains the types for `ParseQuery.or(...)` that I can't seem to find in `index.d.ts`. 

> As far as publishing to npm, we should plan for the next major release of the SDK

Why major release, what's the breaking change?

dplewis:
> Don't the other .d.ts files contain types as well? 

We can include those in the index.d.ts now or later. I choose later after the files have been converted to typescript.

> Why major release, what's the breaking change?

~~Converting files to es6. Can't generate typings without it. Some users import internal Parse files in their projects using commonjs for patches, etc. Those imports would change.~~ 

Edit: Nevermind

```
require('../CoreManager') -> require('../CoreManager').default
```

mtrezza:
> We can include those in the index.d.ts now or later. I choose later after the files have been converted to typescript.

Does that mean the type files that are being generated are not actually used, only the static index.d.ts file that was copied from DefinitelyTyped? And once we have good type coverage we replace that static file with our generated file?

Has anyone looked at the license under which the DT index file is provided? To include that in our repository we'd need to ensure their license allows that and we comply with the license requirements.

dplewis:
Let's make this easy. Users will have to keep using definitelyTyped until we finish converting the entire project to typescript.

dplewis:
> Does that mean the type files that are being generated are not actually used

That is correct because they have typing errors. But that are being tested against 

Another benefit for using a static type file is it’s easy to review against the current definitlyTyped types. It will make converting files easier too since we know most of the types

mtrezza:
That sounds good, but the CI ensures that our own generated type files are coherent with correct types?

So I guess the license question remains.

dplewis:
It’s MIT license so we can use it

mtrezza:
I think we need to add that into our NOTICE file and probably add the MIT license file as well, not sure what MIT requires regarding redistribution of code.

dplewis:
@mtrezza Do what you think is best.

dplewis:
@dblythy can you review the changes? Let me know if you have any questions or if I missed something.

mtrezza:
I think the license topic still needs to be investigated.

dplewis:
> I think the license topic still needs to be investigated.

@mtrezza I'd like the code to be reviewed at some point. I'll remove the definitelyTyped from this PR. Users will still be able to use `@types/parse` until the SDK is fully converted to typescript.

src/Parse.ts
```diff
@@ -0,0 +1,449 @@
+import decode from './decode';
```
dplewis, line 1: @dblythy Since you deleted the old Parse.js and created Parse.ts instead of renaming wouldn't this delete the commit history?

mtrezza:
What was the purpose of adding DT to this PR in the first place? Just to understand the implication of its removal.

dplewis:
To get rid of the dependency on external typings. See https://github.com/parse-community/Parse-SDK-JS/pull/1954

types/index.d.ts
```diff
@@ -0,0 +1,8 @@
+// https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b23a36e669fa127d1035e22ca93faab85b98e49f/types/parse/index.d.ts#L11
+// TODO(dplewis): Possibly add @types/node module reference types
+
+/// <reference path="node.d.ts" />
+/// <reference path="react-native.d.ts" />
```
mtrezza: Could you remove your name? We don't assign tasks to specific contributors via code comments.
Can we remove the rest of the comments, or rephrase them so contributors can understand?

mtrezza (changes requested):

types/index.d.ts
dplewis: Done. Documentation for reference types directives are [here](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-types-)

mtrezza (approved):
Looks good; maybe someone else could take a look at this as well before we merge? @dblythy @sadortun @radandevist

types/index.d.ts
dplewis: @mtrezza Reference types are for `parse/node` and `parse/react-native` types. Since they aren’t in the src folder and at the root like index.js they have to be manually created instead of generated.

mtrezza:
I think we can leave this open for another 2 days; if there are no opposing reviews we'll go ahead and merge.

sadortun (approved):

dblythy:
I'm concerned that the `allowJS` will allow broken types that will throw issues for developers. An example is that the type `mixed` (which is compatible with flow) is used throughout.

Also as previously mentioned there are a lot of cases where the types are wrong (optionals are not specified, there's wrong  types throughout, etc) 

I've also enabled a few of the tests because I think we should at least have a few of them activated if we're going to merge. 

Another thing that I've observed, even though the `Session` file has been renamed to `.ts`, issues still occur when `build` is ran, as typescript can't interpret the types from `ParseObject`.

Screenshot 2023-08-10 at 9 14 26 pm

dplewis:
> I'm concerned that the allowJS will allow broken types that will throw issues for developers.

We should change the title of the PR to `Add initial TypeScript support`. I purposefully removed `types/index.d.ts` from the package.json to prevent issues for developers until the we have full TypeScript support.

> issues still occur when build is ran, as typescript can't interpret the types from ParseObject.

There are a few reasons why I specifically chose ParseSession to be the first file to convert to TypeScript. It has few functions and the limited dependencies it uses are the most important in the project. Its a good place to start since we are about to go down a typing rabbit hole.

dblythy:
Sounds good - apologies, I'll revert my changes

dblythy:
LGTM ✅ 

Thanks @dplewis

mtrezza:
Great to see this is ready for merge; just one last thing: it's unclear what "initial TypeScript support" is supposed to mean. Instead of adding "initial", if we want to communicate anything specific, we can spell it out in the changelog. I can amend the changelog entry in the commit message, so we don't have to write an unreadably-long PR title. Could someone compose a short sentence what we mean by "initial"?

dplewis:
> if we want to communicate anything specific

These are internal changes with no effect on the developers yet. Nothing much to communicate unless to you want to show TS progress through changelogs. We can treat this the same as CI changes.

mtrezza:
Then this should not be a `feat` but a `refactor` PR. I have changed it, and since there won't be any changelog entry I also removed the "initial", since it will only show in the commit history and one can only guess what that could mean anyway.

Please confirm that this is correct and I'll go ahead and merge.

dplewis:
@mtrezza LGTM!

mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [4.2.0-alpha.5](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.2.0-alpha.5)

parseplatformorg:
🎉 This change has been released in version [4.3.0-beta.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.3.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [4.3.0-alpha.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.3.0-alpha.1)

parseplatformorg:
🎉 This change has been released in version [4.3.0](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.3.0)

PR #2002, opened 2023-08-25
ci: Improve test coverage

dplewis:
## Pull Request

## Issue

Add missing coverage

## Approach

* Test unsupported live queries
* Test destroy with session token
* Test live query connections / disconnection
* Test internal nested object states

- [x] Add tests

### DIFF
--- a/src/OfflineQuery.js
+++ b/src/OfflineQuery.js
@@ -538,7 +538,7 @@ function matchesKeyConstraints(className, object, objects, key, constraints) {
         const distance = point.radiansTo(centerPoint);
         return distance <= maxDistance;
       }
-      break;
+      return false;
     }
     case '$geoIntersects': {
       const polygon = new ParsePolygon(object[key].coordinates);
--- a/src/ObjectStateMutations.js
+++ b/src/ObjectStateMutations.js
@@ -150,7 +150,9 @@ export function estimateAttributes(
 function nestedSet(obj, key, value) {
   const path = key.split('.');
   for (let i = 0; i < path.length - 1; i++) {
-    if (!(path[i] in obj)) obj[path[i]] = {};
+    if (!(path[i] in obj)) {
+      obj[path[i]] = {};
+    }
     obj = obj[path[i]];
   }
   if (typeof value === 'undefined') {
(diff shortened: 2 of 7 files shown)


src/OfflineQuery.js
```diff
@@ -538,7 +538,7 @@ function matchesKeyConstraints(className, object, objects, key, constraints) {
         const distance = point.radiansTo(centerPoint);
         return distance <= maxDistance;
       }
-      break;
+      return false;
```
mtrezza, line 541: Why this change? Seems to go beyond "test coverage" - is this a bug fix?

dplewis: It’s technically unreachable code. The change is for the test to see what happens if it is reached

mtrezza: Are you sure it's unreachable? I haven't looked much into this, but what if the constraint is neither `$polygon` nor `$centerSphere`, wouldn't it be reachable?

dplewis: Using the SDK as is it’s unreachable. You could mess around with the internals to reach it like I do with the tests

mtrezza: Got it, and have you looked into what the difference is between `break` and `return false`? For example `case '$maxDistance'` breaks, it doesn't return false, so maybe there is a reason why this here has a `break` and is not returning false - even if it's (currently) unreachable?

dplewis: $maxDistance is ignored so the break statement falls to true (matching) ignoring the default (false). Here in $geoWithin this would be non matching invalid query and should return false instead of break which would be true.

It has a break statement because a test wasn’t written for it and was overlooked

parseplatformorg:
🎉 This change has been released in version [4.2.0-alpha.7](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.2.0-alpha.7)

parseplatformorg:
🎉 This change has been released in version [4.3.0-beta.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.3.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [4.3.0-alpha.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.3.0-alpha.1)

parseplatformorg:
🎉 This change has been released in version [4.3.0](https://github.com/parse-community/Parse-SDK-JS/releases/tag/4.3.0)

PR #2011, opened 2023-08-31
feat: Allow custom server connection fail message

### LINKED ISSUE #1469: Change default connection failure message

### Current Limitation

If Parse JS SDK is unable to connect to the server, the error thrown is `'XMLHttpRequest failed: "Unable to connect to the Parse API"` which is rather confusing for client users.

It would be good to change the default error message to something more helpful such as `Could not connect to the server` which non-technical users could understand. Or, allow JS SDK to specify custom error messages via something like `Parse.customError.101`

### Example Use Case

n/a

n/a

### 3rd Party References

[related](https://github.com/parse-community/parse-server/issues/7915)

dplewis:
## Pull Request

## Issue

Allows users to change the default connection failed message. Developers will be able to display more helpful messages. This is a breaking change for developers that use `Parse._ajax` which is unlikely since ajax is used internally.

Closes: https://github.com/parse-community/Parse-SDK-JS/issues/1469

## Approach

* Add `CONNECTION_FAILED_MESSAGE` to CoreManager
* Return `Parse.Error` in `RESTController.ajax`
* Update EventuallyQueue API to support custom messages

## Tasks

- [x] Add tests

### DIFF
--- a/src/ParseObject.js
+++ b/src/ParseObject.js
@@ -1219,7 +1219,7 @@ class ParseObject {
     try {
       await this.save(null, options);
     } catch (e) {
-      if (e.message === 'XMLHttpRequest failed: "Unable to connect to the Parse API"') {
+      if (e.message === CoreManager.get('CONNECTION_FAILED_MESSAGE')) {
         await EventuallyQueue.save(this, options);
         EventuallyQueue.poll();
       }
@@ -1363,7 +1363,7 @@ class ParseObject {
     try {
       await this.destroy(options);
     } catch (e) {
-      if (e.message === 'XMLHttpRequest failed: "Unable to connect to the Parse API"') {
+      if (e.message === CoreManager.get('CONNECTION_FAILED_MESSAGE')) {
         await EventuallyQueue.destroy(this, options);
         EventuallyQueue.poll();
       }
--- a/integration/test/ParseServerTest.js
+++ b/integration/test/ParseServerTest.js
@@ -16,7 +16,7 @@ describe('ParseServer', () => {
     const object = new TestObject({ foo: 'bar' });
     await parseServer.handleShutdown();
     await new Promise((resolve) => parseServer.server.close(resolve));
-    await expectAsync(object.save()).toBeRejectedWithError('XMLHttpRequest failed: "Unable to connect to the Parse API"');
+    await expectAsync(object.save()).toBeRejectedWithError('The connection to the Parse servers failed.');
     await reconfigureServer({});
     await object.save();
     assert(object.id);
--- a/integration/test/ParseServerTest.js
+++ b/integration/test/ParseServerTest.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const assert = require('assert');
+const Parse = require('../../node');
 
 describe('ParseServer', () => {
   it('can reconfigure server', async () => {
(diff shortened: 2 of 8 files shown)


mtrezza:
> This is a breaking change for developers that use Parse._ajax which is unlikely since ajax is used internally.

Internal methods are not considered breaking changes, so we can ignore that fact.

mtrezza (changes requested):
This doesn't seem to be the correct fix for the bug. The reported bug is that the error message `'XMLHttpRequest failed: "Unable to connect to the Parse API"` "is rather confusing for client users." We do have an existing Parse.Error that has a clear error message `The connection to the Parse servers failed.`, which should be thrown, see:

https://github.com/parse-community/Parse-SDK-JS/issues/1469#issuecomment-1086982822

dplewis:
This is a feature not a bug fix. I want to change the message entirely as I don't want people to know I use `Parse servers` or `xmlhttprequest` under the hood for security reasons. The correct error code (100) is thrown here. If you want you can do.

```
Parse.CoreManager.set('CONNECTION_FAILED_MESSAGE', 'The connection to the Parse servers failed.');
```

mtrezza:
I see, but building a feature on top of incorrect code could make the bug fix more complicated. If the bug is fixed and the error message of Parse.Error 100 is thrown, then the concept of your feature may need to change significantly.

I also think that we may need to think about how we override with custom error messages. The way this is currently designed in this PR seem to require a condition everywhere the error is, but when a Parse.Error is thrown (the bug is fixed), it may make more sense to override the predefined Parse.Error message at it's definition.

dplewis:
> building a feature on top of incorrect code

I'm not sure I'm following you. What is missing from this PR?

mtrezza:
I think it makes more sense to fix the bug first, and then think about how to override the message of the Parse.Error 100. Overriding the Parse.Error message may require a different solution than this PR currently implements.

dplewis:
> I think it makes more sense to fix the bug first

Basically you want me to throw this by default?

```
new Parse.Error(Parse.Error.CONNECTION_FAILED, 'The connection to the Parse servers failed.')
```

mtrezza:
Yes, apologies for the confusion, I assumed the error messages were hard coded in `Parse.Error` as well, but I just look it up and they don't seem to be, which is a strange design.

Let's assume we wants to allow customizing other error messages as well in the future. So we would end up with a whole lot of string definitions in the Core Manager. That doesn't seem to be an elegant solution. I wonder if this PR can be rewritten in a way so that it's more flexible. For example instead of creating a separate key for each error message, why not let the developer define the error code and the overriding message? Like so:

```
[
  {
    code: 100,
    message: "Can't connect to server",
  },
  {
    code: 101,
    message: "Oops.",
  },
]
```

Then, in the Parse.Error constructor you could simply override the `message` argument. And that works for all errors, not just error 100, without any additional code.

So I'd change:

1. All default error messages for code 100 to 'The connection to the Parse servers failed.'. (= the bug fix)
2. Add the overriding logic to Parse.Error (= the new overriding feature)

dplewis:
I think that could work. Let me look into it.

src/ParseObject.js
```diff
@@ -1363,7 +1363,7 @@ class ParseObject {
     try {
       await this.destroy(options);
     } catch (e) {
-      if (e.message === 'XMLHttpRequest failed: "Unable to connect to the Parse API"') {
+      if (e.message === CoreManager.get('CONNECTION_FAILED_MESSAGE')) {
```
mtrezza, line 1366: This seems like a super fragile design, testing against an error message. Unrelated to this PR, but that code really smells. Why is this not testing against the error code?

dplewis: The error message here can only be thrown from one place in the entire SDK. Error code 100 can have multiple error messages since it’s the fallback in the RESTController.

This was the safest approach at the time, I can look into change it to test check the code.

dplewis:
`Parse.Error.CONNECTION_FAILED` will have to be a special case handled separately since it could be an xmlhttprequest error or an unparseable response from the server and the developer should be able to see those. All other messages can be customized using the contructor method.

dplewis:
Closing in favor of https://github.com/parse-community/Parse-SDK-JS/pull/2014

PR #2015, opened 2023-09-03
fix: Data lost when calling `saveEventually`, `destroyEventually` with invalid session token

### LINKED ISSUE #2008: saveEventually - "code":209,"stack":"Error: Session token is expired"
### New Issue Checklist

### Issue Description

I am using saveEventually and have noted the following:

> Data is pinned(saved locally on the device) correctly during a valid session when no network is found.
> I noted data is not saved at all locally if a session expires...
> Upon the network connection returning data gets sent correctly [ although receiving doubles or multiples ] the data is at least found on parse server.
> However if the connection returns and meanwhile the session has expired an attempt is still made and essentially the data gets transmitted and discarded, never saved to parse server and getting essentially lost. 

I have checked the code in the SDK for saveEventually and it seems the server only polls for status ok..... and then attempts to send the queue.

### Steps to reproduce

to reproduce one has to set a very short session expiry on parse server.
### Actual Outcome

The outcome is lost data upon session expiry both locally and on parse server.

### Expected Outcome

I expected the data to live locally until a user gets logged in again(not just network restored again) and has a valid session. 
I expected the data to still be saved locally on the device even if the session expires. 

###  Environment

Server
- Parse Server version: 5.1.0
- Operating system: n/a
- Local or remote host Local

Database
- System (MongoDB or Postgres): n/a
- Database version: n/a
- Local or remote host n/a

Client
- Parse JS SDK version:  parse@4.1.0

### Logs

info: Create new client: f5e9c41b-2378-4d16-a167-d849dbef9d9c
error: Session token is expired. {"code":209,"stack":"Error: Session token is expired.\n    at Object.getAuthForSessionToken (/usr/local/parse-server/node_modules/parse-server/lib/Auth.js:132:11)\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)\n    at async handleParseSession (/usr/local/parse-server/node_modules/parse-server/lib/middlewares.js:292:21)"}

dplewis:
## Pull Request

## Issue

When sessionToken is invalid during saveEventually / destroyEventually the data is lost.

Closes: https://github.com/parse-community/Parse-SDK-JS/issues/2008

## Approach

* Keep data local on invalid sessionToken
* Retry eventually queue when the user changes

## Tasks

- [x] Add tests

### DIFF
--- a/src/EventuallyQueue.js
+++ b/src/EventuallyQueue.js
@@ -30,6 +30,7 @@ const QUEUE_KEY = 'Parse/Eventually/Queue';
 let queueCache = [];
 let dirtyCache = true;
 let polling = undefined;
+let processing = false;
 
 /**
  * Provides utility functions to queue objects that will be
@@ -235,6 +236,7 @@ const EventuallyQueue = {
     if (queueData.length === 0) {
       return false;
     }
+    processing = true;
     for (let i = 0; i < queueData.length; i += 1) {
       const queueObject = queueData[i];
       const { id, hash, className } = queueObject;
@@ -247,6 +249,7 @@ const EventuallyQueue = {
         await this.process.create(ObjectType, queueObject);
       }
     }
+    processing = false;
     return true;
   },
 
@@ -276,7 +279,7 @@ const EventuallyQueue = {
         await object.save(queueObject.object, queueObject.serverOptions);
         await this.remove(queueObject.queueId);
       } catch (e) {
-        if (e.code !== ParseError.CONNECTION_FAILED) {
+        if (![ParseError.CONNECTION_FAILED, ParseError.INVALID_SESSION_TOKEN].includes(e.code)) {
           await this.remove(queueObject.queueId);
         }
       }
@@ -286,7 +289,7 @@ const EventuallyQueue = {
         await object.destroy(queueObject.serverOptions);
         await this.remove(queueObject.queueId);
       } catch (e) {
-        if (e.code !== ParseError.CONNECTION_FAILED) {
+        if (![ParseError.CONNECTION_FAILED, ParseError.INVALID_SESSION_TOKEN].includes(e.code)) {
           await this.remove(queueObject.queueId);
         }
       }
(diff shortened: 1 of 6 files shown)


src/EventuallyQueue.js
```diff
@@ -348,20 +351,36 @@ const EventuallyQueue = {
...
-      const { sessionToken } = queueObject.serverOptions;
+      const { sessionToken = '' } = queueObject.serverOptions;
       const query = new ParseQuery(ObjectType);
       query.equalTo('objectId', queueObject.id);
       const results = await query.find({ sessionToken });
       return EventuallyQueue.sendQueueCallback(results[0], queueObject);
     },
     async byHash(ObjectType, queueObject) {
-      const { sessionToken } = queueObject.serverOptions;
+      const { sessionToken = '' } = queueObject.serverOptions;
```
mtrezza, line 383: Why should it execute a query if the session token is an empty string? Wouldn't that query always fail?

src/EventuallyQueue.js
```diff
@@ -276,7 +279,7 @@ const EventuallyQueue = {
         await object.save(queueObject.object, queueObject.serverOptions);
         await this.remove(queueObject.queueId);
       } catch (e) {
-        if (e.code !== ParseError.CONNECTION_FAILED) {
+        if (![ParseError.CONNECTION_FAILED, ParseError.INVALID_SESSION_TOKEN].includes(e.code)) {
```
mtrezza, line 282: If I read this correctly, pre-PR, an operation in the queue will be discarded if the session token is invalid; post-PR, the operation stays in the queue.
- What type of queue is this? Will a such a pending operation block the queue for others to be executed?
- When will these failed operations due to invalid session token be discarded? After n failed retries?
- How is the queue managed when the current user logs out - will the queue emptied?
- How is the queue managed when another user logs in - will the queue be emptied?

src/EventuallyQueue.js
```diff
@@ -235,6 +236,7 @@ const EventuallyQueue = {
     if (queueData.length === 0) {
       return false;
     }
+    processing = true;
```
mtrezza, line 239: Could `sendQueue` be called twice and lead to a race condition regarding the value `processing`?

src/EventuallyQueue.js
dplewis: The opposite, the query always runs regardless if the user's session token is invalid.

RESTController.js
```
        const userController = CoreManager.getUserController();
        if (options && typeof options.sessionToken === 'string') {
          return Promise.resolve(options.sessionToken);
        } else if (userController) {
          return userController.currentUserAsync().then(user => {
            if (user) {
              return Promise.resolve(user.getSessionToken());
            }
            return Promise.resolve(null);
          });
        }
        return Promise.resolve(null);
```

src/EventuallyQueue.js
dplewis: `processing` was add for integration testing.

src/EventuallyQueue.js
dplewis: > What type of queue is this? Will a such a pending operation block the queue for others to be executed?

A standalone non-blocking queue

> When will these failed operations due to invalid session token be discarded? After n failed retries?

If a valid sessionToken exists and the save fails for another reason, then it will be discarded

> How is the queue managed when the current user logs out - will the queue emptied?

Currently the queue isn't cleared on logout

> How is the queue managed when another user logs in - will the queue be emptied?

The queue runs when a user logs in

src/EventuallyQueue.js
mtrezza: What's the point of sending an empty token to the server? Could such a query ever succeed?

src/EventuallyQueue.js
mtrezza: Thanks for the insight, I've invited you to a separate thread to discuss.

src/EventuallyQueue.js
dplewis: The query does succeed unless there is a permission issue

mtrezza: You mean for example if there is a public read/write permission the query would succeed?

dplewis: Yes

mtrezza: Got it, and why send an empty string instead of not setting the sessionToken? An empty string could technically be considered a token of zero length, i.e. an invalid token.

dplewis:
Closing as more planning is needed.

PR #2063, opened 2023-12-05
feat: Add support for Node 20, remove support for Node 14 and 16

nikeee:
According to the readme, parse follows the supported node versions:

> Parse JS SDK is continuously tested with the most recent releases of Node.js to ensure compatibility.
> We follow the [Node.js Long Term Support plan](https://github.com/nodejs/Release) and only
> test against versions that are officially supported and have not reached their end-of-life date.

This commit aligns the repository with the currently supported Node.js versions.
See:
https://github.com/nodejs/Release

Closes: #2065

Issue #1979, opened 2023-07-23
Tests fail with Parse Server upgrade


mtrezza:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-JS/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse JS SDK](https://github.com/parse-community/Parse-SDK-JS/releases). 

### Issue Description

Upgrading Parse Server for integration testing fails in https://github.com/parse-community/Parse-SDK-JS/pull/1975.

### Steps to reproduce

Failing tests in https://github.com/parse-community/Parse-SDK-JS/actions/runs/5633637975/job/15262758943?pr=1975:

> 1) Parse.File can get file data from base64 (saved)
>   - Error: File upload of extension plain is disabled.
> 
> 2) Parse.File can get file data from byte array
>   - Error: File upload of extension plain is disabled.
> 
> 3) Parse.File can save file to localDatastore
>   - Error: File upload of extension plain is disabled.
> 
> 4) Schema set multiple required and default values
>   - Error: File upload of extension plain is disabled.

### Actual Outcome

Tests fail.

### Expected Outcome

Test pass.

dblythy:
This seems to be as a result of an unexpected breaking change, fixed in https://github.com/parse-community/parse-server/issues/8698

mtrezza:
Possibly related to https://github.com/parse-community/parse-server/issues/8818

PR #2088, opened 2024-02-19
feat: Add comment to MongoDB query via `Parse.Query.comment`

Meglali20:
SDK changes for : https://github.com/parse-community/parse-server/issues/8799


mtrezza:
CI fails with:

> 1) Parse Query can query and send a comment
>   - Error: Invalid parameter for query: comment

I believe the tests are not running with the latest alpha version of Parse Server because the package-lock file usually specifies a commit hash so it's locked to a specific version. Running `npm i -DE parse-server@7.0.0-alpha.23` should fix that. However, there's https://github.com/parse-community/Parse-SDK-JS/issues/1979 which we should fix in the coming days, so some tests may fail that are unrelated to this PR.

Meglali20:
Yes, I was expecting this error. for me, all tests passed when I used parse server as local package after making changes locally related to https://github.com/parse-community/parse-server/pull/8928

mtrezza:
Could you upgrade Parse Server like I've mentioned?

mtrezza:
There has been some issue with the package-lock file; I've reverted it and ran `npm i -DE parse-server@7.0.0-alpha.25`.

mtrezza:
I believe we need to upgrade the Parse JS SDK to support only Node 18 and 20, the same node versions as Parse Server 7. In a separate PR, before we can merge this PR. https://github.com/parse-community/Parse-SDK-JS/pull/2063

mtrezza:
Tests are failing because the Parse JS SDK test suite needs to be updated for Parse Server 7. For example, allowClientClassCreation needs to be set to true. https://github.com/parse-community/Parse-SDK-JS/issues/1979

Meglali20:
When running tests locally most of them pass, did you figure out the issue for this?

mtrezza:
Did you download the latest commit of this branch? I would be surprised if it passed locally. They likely pass when testing with an older version of Parse Server, but not with the latest alpha version as in this branch.

Meglali20:
You're correct tests are failing, didn't notice you changed to alpha 25 I had to run `npm i -DE parse-server@7.0.0-alpha.25` after that tests started failing.

mtrezza:
Yes, I recommend to pull the latest commit in this PR to your local clone, so that you use the same code, delete your current `node_modules` directory, then run just `npm i`. Otherwise you may see different errors.

mtrezza:
Tests are passing now; could you check the codecov? It seems that the new tests are not testing all lines of the new feature. After that it should be good to merge.

src/ParseQuery.js
```diff
@@ -2111,6 +2122,26 @@ class ParseQuery {
...
+
+  /**
+   * Sets a comment to the query so that the query
+   *can be identified when using a the profiler for MongoDB.
+   *
+   * @param {string} value a comment can make your profile data easier to interpret and trace.
+   * @returns {Parse.Query} Returns the query, so you can chain this call.
+   */
+  comment(value: string): ParseQuery {
+    if (typeof value === 'undefined') {
```
mtrezza:
```suggestion
    if (value == null) {
```

Simplify, so that undefined and null values are considered "no value"?

src/ParseQuery.js
```diff
@@ -2111,6 +2122,26 @@ class ParseQuery {
...
+      delete this._comment;
+      this._comment = value;
```
mtrezza: Why is this needed after delete?

```suggestion
```

mtrezza (approved):
Looks good! Codecov is fixed and all tests pass, thanks for the PR!

parseplatformorg:
🎉 This change has been released in version [5.0.0-alpha.3](https://github.com/parse-community/Parse-SDK-JS/releases/tag/5.0.0-alpha.3)

parseplatformorg:
🎉 This change has been released in version [5.0.0-beta.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/5.0.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [5.0.0](https://github.com/parse-community/Parse-SDK-JS/releases/tag/5.0.0)

PR #2090, opened 2024-03-16
fix: `Parse.Object.saveEventually` fails if server is restarting while calling it

### LINKED ISSUE #2028: saveEventually / EventuallyQueue doesn't work when killing server -> saveEventually object -> restarting server
### New Issue Checklist

### Issue Description

Loading an object in client, then calling saveEventually on it after killing the parseserver, then starting it up again doesn't work for me. It seems to be a few different issues. Maybe I'm just doing something really wrong, and if so I'm sorry in advance :) 

Test code is really simple. I find object "Foo" (with server up) in the console, kill server, client executes this code: 

```
result[0].set("counter", result[0].get("counter")+1); await result[0].saveEventually(); 
```

I can see the `Parse.EventuallyQueue.getQueue()` have the following

```
'[{"queueId":"save_Foo_T6r7uWTcU9_","action":"save","object":{"counter":13,"createdAt":"2023-09-22T16:00:06.541Z","updatedAt":"2023-09-22T18:48:30.396Z","ACL":{"G2lwffEEaD":{"read":true,"write":true}},"objectId":"T6r7uWTcU9"},"serverOptions":{},"id":"T6r7uWTcU9","className":"Foo","createdAt":"2023-09-22T18:48:44.083Z"}]
```

When I restart the server, I don't see any attempts to save to server but queue is empty. It seems to fail this test in `EventuallyQueue.sendQueueCallback` under save:

```
// Queued update was overwritten by other request. Do not save
      if (
        typeof object.updatedAt !== 'undefined' &&
          object.updatedAt > new Date(queueObject.object.createdAt)
      )
```

The code is checking is the object's updatedAt is greater than the object's createdAt, which will always be the case. My guess was that this should be `object.updatedAt > queueObject.createdAt` (when the **queue object** was created not the original object). hacking on the JS code did get me past this problem but it still wouldn't save.

Next issue is that the EventuallyQueue stores a .toJSON of my object, which includes among other things the `ACL` object. 

With some extra logging, the ParseObject.save gets sent that ACL object. It fails on `ParseObject.validate(attrs)` (silently) with this:

```
Error: ACL must be a Parse ACL.
    at ParseObjectSubclass.value (ParseObject.js:1259:16)
    at ParseObjectSubclass.value (ParseObject.js:1578:31)
    at EventuallyQueue.js:386:27
```

I'm not sure here if the best fix is to delete the ACL out of the json stored in EventuallyQueue or if ParseObject.validate should allow for JSON versions of the ACL, but it does fail my very simple `saveEventually` regardless. 

I don't normally modify ACLs on client, only server using save hooks, but taking them out would be one thing.

Another fix (although still would fail with ACLs changed on client) is to not do (in `EventuallyQueue.enqueue`):

```
 queueData[index] = {
      queueId,
      action,
      object: object.toJSON(), // <-- This stores ACL in json format..
      serverOptions,
      id: object.id,
      className: object.className,
      hash: object.get('hash'),
      createdAt: new Date(),
    };
```

Instead of `object` being the entire `toJSON()` storing only the values of the changed keys (`_getSaveJSON()` ?). Or maybe change the ACL object if present back into a valid version. 

### Actual Outcome
Data is lost

### Expected Outcome
Expected data to be saved when server starts up again.

###  Environment

Server
- Parse Server version: `5.2.3`
- Operating system: `OSX`

Database
- Database version: `5.5`

Client
- Parse JS SDK version: `4.1.0`

mortenmo:
## Pull Request

## Issue

Closes: [2028](https://github.com/parse-community/Parse-SDK-JS/issues/2028)

As issue describes, save fails due to the ACL object being stored to JSON and back causing the following error when connection is back:

```
Error: ACL must be a Parse ACL.
    at ParseObjectSubclass.value (ParseObject.js:1259:16)
    at ParseObjectSubclass.value (ParseObject.js:1578:31)
    at EventuallyQueue.js:386:27
```

## Approach

This is a quick fix using the saveJson format instead when storing the object locally. There might be a better representation of the object in a string store if someone can tell me what that is.

### DIFF
--- a/src/EventuallyQueue.js
+++ b/src/EventuallyQueue.js
@@ -117,7 +117,7 @@ const EventuallyQueue = {
     queueData[index] = {
       queueId,
       action,
-      object: object.toJSON(),
+      object: object._getSaveJSON(),
       serverOptions,
       id: object.id,
       className: object.className,
--- a/src/__tests__/EventuallyQueue-test.js
+++ b/src/__tests__/EventuallyQueue-test.js
@@ -25,6 +25,9 @@ class MockObject {
   toJSON() {
     return this.attributes;
   }
+  _getSaveJSON() {
+      return this.attributes;
+  }
   static extend(className) {
     class MockSubclass {
       constructor() {


src/__tests__/EventuallyQueue-test.js
```diff
@@ -25,6 +25,9 @@ class MockObject {
   toJSON() {
     return this.attributes;
   }
+  _getSaveJSON() {
```
mtrezza, line 28: What's the diff to `toJSON`?

mortenmo: The main thing that makes this work is that `_getSaveJSON()` (outside the mock) does not include the `ACL` object while `toJSON` does. In the mocks, there is no real different why the tests didn't fail already they don't 

When using saveEventually and connection is down, the object gets stored in the queue as json. Today it uses 'getJSON' and that includes a json representation of ACL. `_getSaveJSON` doesn't. 

This json object is later used (line 276) to save the queued object when the server is back up:

`await object.save(queueObject.object, queueObject.serverOptions);`

save here does not support ACL in json format (or ACL at all, don't think you can change access outside of server). 

There are TBH probably even better fixes here. We should probably only store the dirty fields of the object and not overwrite everything (but it does already make sure before line 276 that its not overwriting other things).

mortenmo: Reason why the EventuallyQueue tests didn't catch this is because ACLs aren't on the objects before they are loaded from the servers.

Doing this..

```
x = new Parse.Object('TestObject');
  ParseObject2 {id: undefined, _localId: undefined, _objCount: 1762, className: 'TestObject'}
x.set('foo', 'bar');
x.toJSON()
{foo: 'bar'}
x.save();
```

Would work fine. But in reality you often modify existing objects. When objects are loaded from server, the `toJSON` includes a lot more properties. Here is the same object:

```
x = await new Parse.Query('TestObject').find();
x[0].toJSON()

   {"foo":"bar",
    "createdAt":"2024-03-17T20:52:35.213Z",
    "updatedAt":"2024-03-17T20:53:30.133Z",
    "ACL":{"*":{"read":true,"write":true}},
    "objectId":"sZhK7r5owf"}

```

All tests in `EventuallyQueue-test` just creates non-server saved objects which doesn't catch this problem with "real" objects from servers.

example:
```
const object = new ParseObject('TestObject');
    object.set('foo', 'bar');
    object.set('test', '1234');
```

dplewis:
@mortenmo Thanks for the PR! I think the validation check on ACL should removed instead of removing ACL from Eventually Queue. I honestly don't know why there ACL check is there. `.validate` is called in both `save` and `set` , `save` also calls `set`. The `validate` call is redundant. `set` transform the ACL json into `Parse.ACL` (which has it own checks) then performs the `.validate` then throws the error.

The following integration test breaks.

```
it('can set ACL from toJSON', async () => {
    Parse.User.enableUnsafeCurrentUser();
    user.set('username', 'torn');
    user.set('password', 'acl');
    await user.signUp();
    const acl = new Parse.ACL(user);
    object.setACL(acl);
    const json = object.toJSON();
    await object.save(json); // This breaks ACL must be a Parse ACL.
 });
```

Here is an integration test replicate the issue for this PR.

```
it('can saveEventually on object with ACL', async () => {
    Parse.User.enableUnsafeCurrentUser();
    const parseServer = await reconfigureServer();
    user.set('username', 'torn');
    user.set('password', 'acl');
    await user.signUp();

    const acl = new Parse.ACL(user);
    const object = new TestObject({ hash: 'saveSecret' });
    object.setACL(acl);

    await new Promise((resolve) => parseServer.server.close(resolve));

    await object.saveEventually();

    let length = await Parse.EventuallyQueue.length();
    assert(Parse.EventuallyQueue.isPolling());
    assert.strictEqual(length, 1);

    await reconfigureServer({});

    while (Parse.EventuallyQueue.isPolling()) {
      await sleep(100);
    }
    assert.strictEqual(Parse.EventuallyQueue.isPolling(), false);

    length = await Parse.EventuallyQueue.length();
    while (length) {
      await sleep(100);
    }
    length = await Parse.EventuallyQueue.length();
    assert.strictEqual(length, 0);

    const query = new Parse.Query('TestObject');
    query.equalTo('hash', 'saveSecret');
    let results = await query.find();
    while (results.length === 0) {
      results = await query.find();
    }
    assert.strictEqual(results.length, 1);
});
```

Removing the ACL check will fix this issue.

src/EventuallyQueue.js
```diff
@@ -117,7 +117,7 @@ const EventuallyQueue = {
     queueData[index] = {
       queueId,
       action,
-      object: object.toJSON(),
+      object: object._getSaveJSON(),
```
dplewis, line 120: @mortenmo Thanks for the PR! I think the validation check on ACL should removed instead of removing ACL from Eventually Queue. I honestly don't know why the ACL check is there. `.validate` is called in both `save` and `set` , `save` also calls `set`. `set` transform the ACL json into `Parse.ACL` (which has it own checks) then performs the `.validate` then throws the error.

The following integration test fails.

```
it('can set ACL from toJSON', async () => {
    Parse.User.enableUnsafeCurrentUser();
    user.set('username', 'torn');
    user.set('password', 'acl');
    await user.signUp();
    const acl = new Parse.ACL(user);
    object.setACL(acl);
    const json = object.toJSON();
    await object.save(json); // This breaks ACL must be a Parse.ACL.
 });
```

Here is an integration test replicate the issue for this PR.

```
it('can saveEventually on object with ACL', async () => {
    Parse.User.enableUnsafeCurrentUser();
    const parseServer = await reconfigureServer();
    user.set('username', 'torn');
    user.set('password', 'acl');
    await user.signUp();

    const acl = new Parse.ACL(user);
    const object = new TestObject({ hash: 'saveSecret' });
    object.setACL(acl);

    await new Promise((resolve) => parseServer.server.close(resolve));

    await object.saveEventually();

    let length = await Parse.EventuallyQueue.length();
    assert(Parse.EventuallyQueue.isPolling());
    assert.strictEqual(length, 1);

    await reconfigureServer({});

    while (Parse.EventuallyQueue.isPolling()) {
      await sleep(100);
    }
    assert.strictEqual(Parse.EventuallyQueue.isPolling(), false);

    length = await Parse.EventuallyQueue.length();
    while (length) {
      await sleep(100); // <---------------------- stalls as data is lost
    }
    length = await Parse.EventuallyQueue.length();
    assert.strictEqual(length, 0);

    const query = new Parse.Query('TestObject');
    query.equalTo('hash', 'saveSecret');
    let results = await query.find();
    while (results.length === 0) {
      results = await query.find();
    }
    assert.strictEqual(results.length, 1);
});
```

Removing the ACL check will fix this issue.

dplewis (changes requested):

dplewis:
Fixed via https://github.com/parse-community/Parse-SDK-JS/pull/2097

PR #2097, opened 2024-03-21
feat: Add support for setting `Parse.ACL` from JSON

### LINKED ISSUE #2096: Cannot set ACL from JSON
### New Issue Checklist

### Issue Description

Saving ACL from JSON throws `ACL must be a Parse.ACL` error. The validation on ACL prevents the json from properly encoding. This may also fix the issue https://github.com/parse-community/Parse-SDK-JS/issues/2028

### Steps to reproduce

```
it('can set ACL from json', async () => {
    Parse.User.enableUnsafeCurrentUser();
    user.set('username', 'torn');
    user.set('password', 'acl');
    await user.signUp();
    const acl = new Parse.ACL(user);
    object.setACL(acl);
    const json = object.toJSON();
    await object.save(json); // This breaks ACL must be a Parse.ACL.
    assert.equal(acl.equals(object.getACL()), true);
 });
```

This error also happens when on `object.set({ ACL: acl.toJSON() })`;

### Actual Outcome

Throws error

### Expected Outcome

Should save object properly

###  Environment

Server
- Parse Server version: `6.0.0`
- Operating system: `MacOS`

Database
- Database version: `4.4`

Client
- Parse JS SDK version: `4.3.1`

dplewis:
## Pull Request

## Issue

Setting ACL from JSON throws `ACL must be Parse.ACL` error. ACL should be handled the same as other special fields like id, createdAt, updatedAt

Closes: https://github.com/parse-community/Parse-SDK-JS/issues/2096, https://github.com/parse-community/Parse-SDK-JS/issues/2028

## Approach

* Properly handles invalid ACL by returning null instead of error

## Tasks

- [x] Add tests

### DIFF
--- a/src/__tests__/ParseObject-test.js
+++ b/src/__tests__/ParseObject-test.js
@@ -909,12 +909,9 @@ describe('ParseObject', () => {
 
   it('validates attributes on save()', async () => {
     const o = new ParseObject('Listing');
-    try {
-      await o.save({ '$$$': 'o_O' });
-      expect(true).toBe(false);
-    } catch (e) {
-      expect(e.code).toBe(ParseError.INVALID_KEY_NAME);
-    }
+    await expect(o.save({ '$$$': 'o_O' })).rejects.toEqual(
+      new ParseError(ParseError.INVALID_KEY_NAME)
+    );
   });
 
   it('ignores validation if ignoreValidation option is passed to set()', () => {
--- a/src/__tests__/ParseObject-test.js
+++ b/src/__tests__/ParseObject-test.js
@@ -364,6 +364,13 @@ describe('ParseObject', () => {
     expect(o.getACL()).toEqual(ACL);
   });
 
+  it('encodes ACL from json', () => {
+    const ACL = new ParseACL({ user1: { read: true } });
+    const o = new ParseObject('Item');
+    o.set({ ACL: ACL.toJSON() });
+    expect(o.getACL()).toEqual(ACL);
+  });
+
   it('can be rendered to JSON', () => {
     let o = new ParseObject('Item');
     o.set({
@@ -871,12 +878,6 @@ describe('ParseObject', () => {
     const o = new ParseObject('Listing');
     expect(
       o.validate({
-        ACL: 'not an acl',
-      })
-    ).toEqual(new ParseError(ParseError.OTHER_CAUSE, 'ACL must be a Parse ACL.'));
-
-    expect(
-      o.validate({
         'invalid!key': 12,
       })
     ).toEqual(new ParseError(ParseError.INVALID_KEY_NAME));
@@ -896,8 +897,6 @@ describe('ParseObject', () => {
 
   it('validates attributes on set()', () => {
     const o = new ParseObject('Listing');
-    expect(o.set('ACL', 'not an acl')).toBe(false);
-    expect(o.set('ACL', { '*': { read: true, write: false } })).toBe(o);
     expect(o.set('$$$', 'o_O')).toBe(false);
 
     o.set('$$$', 'o_O', {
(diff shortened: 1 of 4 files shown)


dplewis:
@mtrezza can you a member of @parse-community/js-sdk ?

mtrezza:
How can I help? :-)

mtrezza (commented):
Is this a breaking change?

dplewis:
This isn't a breaking change.

src/__tests__/ParseObject-test.js
```diff
@@ -908,6 +907,16 @@ describe('ParseObject', () => {
     });
   });
 
+  it('validates attributes on save()', async () => {
+    const o = new ParseObject('Listing');
+    try {
+      await o.save({ '$$$': 'o_O' });
+      expect(true).toBe(false);
```
mtrezza: Looks odd, should it be asyncAwait?

src/__tests__/ParseObject-test.js
```diff
@@ -869,12 +876,6 @@ describe('ParseObject', () => {
 
   it('can validate attributes', () => {
     const o = new ParseObject('Listing');
-    expect(
-      o.validate({
-        ACL: 'not an acl',
-      })
-    ).toEqual(new ParseError(ParseError.OTHER_CAUSE, 'ACL must be a Parse ACL.'));
```
mtrezza: So now instead of throwing an error, setting an invalid ACL silently fails by setting the ACL to `null`?

src/__tests__/ParseObject-test.js
dplewis: I wrote it this way since there is only an error code and no error message. I'll change it

src/__tests__/ParseObject-test.js
dplewis: Yes, also users have the option to pass `ignoreValidation` option to `set`, in that case `getACL` would return `null`.

mtrezza: So that's a change in behavior, right? How is this change related to the feature "add setting ACL from JSON"? Or can we say that throwing an error as it currently does is a bug?

dplewis: I would consider this to be a bug. I'm not entirely sure why its there. There are no integration tests in this SDK or the Server. Also this error message doesn't exist in any other SDK.

mtrezza: I see, you say it's possible to set an invalid ACL in other Parse client SDKs and they fail silently without throwing an error. Both behaviors sound normal to me. For someone using this SDK, it would be considered normal behavior to get an error. If they built code around catching an error when setting an invalid ACL, that error would not be thrown anymore. In other words, they wouldn't be getting any ACL validation feedback anymore, right? In a practical use case, that means that an object would be saved without ACL rather than not saved at all on an invalid ACL. That sounds like a severe change actually.

dplewis: I don't see how users would encounter this error if they follow the ACL documentation unless they set ACL without Parse.ACL like I'm doing in this PR. Thats probably the reason why there aren't any integration tests. If they some how do bypass the sdk internals with invalid ACL the server throws `Tried to create an ACL with an invalid permission type`.

mtrezza: Do you mean that using the official methods of the SDK it's currently not possible to invoke `ParseError.OTHER_CAUSE, 'ACL must be a Parse ACL.'`? With "official methods" I mean the publicly accessible methods of the SDK, regardless of whether they are documented in the guides or not, as these docs usually lag behind.

dplewis: Correct, you can try to write a failing test yourself

mtrezza: What about `o.setACL('invalid');`, wouldn't that throw the error? Even though when setting an ACL, it has to be an instance of [Parse.ACL](http://parseplatform.org/Parse-SDK-JS/api/5.0.0/Parse.ACL.html) according to the docs, so this would be a programmer's mistake in the first place.

Edit: I just noticed that there's test for this called `validates attributes on set()` which does `expect(o.set('ACL', 'not an acl')).toBe(false);`. So I guess it wouldn't throw the error, just return `false`? But this expect has been removed from the test, how come?

dplewis: Surprisingly `o.setACL('invalid');` wouldn't throw an error it would return an error which wouldn't be seen by the developer unless they pass the appropriate options to it.

dplewis: @mtrezza I have an idea that could work. Both `save` and `set` both validates the attributes. `save` calls `set` internally, removing the redundant validation should keep the error and allow for this feature.

mtrezza: Yes, maybe that's a way forward. Just so that we avoid a behavior change that could break existing apps. The change of the existing tests hints that there's a behavior change. But I agree that the current behavior looks odd as well.

What would be the ideal behavior a developer would want to see? I think `o.setACL('invalid');` should throw an error, right? At worst, a developer tries to set a ACL incorrectly and it fails silently, creating objects without ACL. We could create an issue for that so we can make a braking change with JS SDK 6 in a few months.

dplewis:
@mtrezza I was able to keep the original error handling while adding this new feature. This is ready for review.

src/__tests__/ParseObject-test.js
dplewis: Backbone support was removed in 2.0.0 I don't know how this got missed.

mtrezza (approved):
Looks good, ready for merge?

dplewis:
Yep

parseplatformorg:
🎉 This change has been released in version [5.1.0-alpha.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/5.1.0-alpha.1)

parseplatformorg:
🎉 This change has been released in version [5.1.0-beta.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/5.1.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [5.1.0](https://github.com/parse-community/Parse-SDK-JS/releases/tag/5.1.0)

PR #2099, opened 2024-03-25
fix: Fail on infinite recursion in `encode.js`

### LINKED ISSUE #2098: Infinite loop in `encode.js`
### New Issue Checklist

### Issue Description

The `encode` function is called recursively without a ceiling, causing the entire service to crash with no hint as to the originating call, even with `VERBOSE=1` enabled. All I know is that the error is thrown from this line:

```diff
if (value && typeof value === 'object') {
  const output = {};
  for (const k in value) {
+   output[k] = encode(value[k], disallowObjects, forcePointers, seen, offline);
  }
  return output;
}
```

### Steps to reproduce

Unsure, inherited legacy code and there is no hint as to which network call is causing this issue.

### Actual Outcome

Program errors out without any hint as to where the error is coming from.

### Expected Outcome

The function to execute properly. Two possible improvements to the library come to mind:
1. Give further context about where the error is coming from
2. Have a limit on the number of calls within the function and truncate after the limit is hit, giving a warning in the console but not causing the program to error out

###  Environment

Server
- Parse Server version: `7.0.0`
- Operating system: `node:18 Docker image`

Database
- Database version: `5.9.2`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `Local`

Client

### Logs

```shell
An uncaught exception occurred: Maximum call stack size exceededStack Trace:
RangeError: Maximum call stack size exceeded
    at encode (/api/node_modules/.pnpm/parse@5.0.0/node_modules/parse/lib/node/encode.js:23:16)
    at /api/node_modules/.pnpm/parse@5.0.0/node_modules/parse/lib/node/encode.js:61:14
    at Array.map (<anonymous>)
    at encode (/api/node_modules/.pnpm/parse@5.0.0/node_modules/parse/lib/node/encode.js:60:18)
    at encode (/api/node_modules/.pnpm/parse@5.0.0/node_modules/parse/lib/node/encode.js:67:19)
    at encode (/api/node_modules/.pnpm/parse@5.0.0/node_modules/parse/lib/node/encode.js:67:19)
    at encode (/api/node_modules/.pnpm/parse@5.0.0/node_modules/parse/lib/node/encode.js:67:19)
    at encode (/api/node_modules/.pnpm/parse@5.0.0/node_modules/parse/lib/node/encode.js:67:19)
    at encode (/api/node_modules/.pnpm/parse@5.0.0/node_modules/parse/lib/node/encode.js:67:19)
```

ajmeese7:
## Issue

Closes: #2098 

## Approach

Implemented the changes suggested by @mtrezza. Will give some clarity to the issues being encountered, and may illuminate further test cases that cause issues that can be handled in a better way in the future.


src/encode.js
```diff
@@ -10,13 +10,30 @@ import ParseObject from './ParseObject';
 import { Op } from './ParseOp';
 import ParseRelation from './ParseRelation';
 
+const MAX_RECURSIVE_CALLS = 999;
+let recursiveCallsCount = 0;
```
mtrezza: This implementation doesn't work. The counter resides at the module scope and is "global" for the module. Each subsequent call to `encode` uses the same global instance of the variable.

Instead, you could add a counter argument to the `encode` function, so that it has a call-scoped counter.

Could you please also add a test to the PR to verify this works?

mtrezza (changes requested):

src/encode.js
ajmeese7: I made the requested argument counter change, but I was struggling to write a test because I don't know much about how Parse works behind the scenes. The `ParseObject` import is mocked in the `encode` test file and I was going down a rabbit hole trying to instantiate the class like I saw in some other test files.

mtrezza: You could just write a test for the encode function directly. Maybe a new test file for `encode.js`. Then just run it with dummy args so that it runs recursively. But writing a higher level test would be better, I agree. For that you could look at in what cases `encode` runs recursively, then write something that makes it run recursively. I'm just guessing but maybe an object that references itself would cause a recursing. Maybe something like a circular Parse Pointer, where pointer A references pointer B which references pointer A... any maybe that is even what is causing the issue you are observing in our deployment.

ajmeese7: I can't get it to work whatsoever, this is what I have:

```js
it('handles circular references in ParseObjects', () => {
  const RealParseObject = jest.requireActual('../ParseObject').default;
  const obj1 = new RealParseObject();
  const obj2 = new RealParseObject();
  obj1.set('obj2', obj2);
  obj2.set('obj1', obj1);

  expect(() => encode(obj1, false, false, [], false)).toThrow();
});
```

This throws a TypeError that has nothing to do with my code:

```shell
TypeError: Cannot read properties of undefined (reading 'estimateAttributes')

      153 |   get attributes(): AttributeMap {
      154 |     const stateController = CoreManager.getObjectStateController();
    > 155 |     return Object.freeze(stateController.estimateAttributes(this._getStateIdentifier()));
          |                                          ^
      156 |   }
      157 |
      158 |   /**

      at ParseObject.estimateAttributes [as attributes] (src/ParseObject.js:155:42)
      at ParseObject.attributes [as set] (src/ParseObject.js:747:36)
      at Object.set (src/__tests__/encode-test.js:192:10)
```

And if I try including `obj1.save()` before setting anything, I get this error:

```shell
TypeError: Cannot read properties of undefined (reading 'save')

      1333 |     const controller = CoreManager.getObjectController();
      1334 |     const unsaved = options.cascadeSave !== false ? unsavedChildren(this) : null;
    > 1335 |     return controller.save(unsaved, saveOptions).then(() => {
           |                       ^
      1336 |       return controller.save(this, saveOptions);
      1337 |     });
      1338 |   }

      at ParseObject.save (src/ParseObject.js:1335:23)
      at Object.save (src/__tests__/encode-test.js:192:10)
```

I have no idea how to proceed from here, so if anyone else knows how this library works and can implement a test, that would be great.

mtrezza: Are you able to run the existing tests locally without issues? Also, when you make any changes, in /src, you need to rebuild the SDK, either with `npm run build` or continuously rebuild on file change with `npm run watch`.

ajmeese7: Yes, I can run the existing tests without issues. And re-running `npm run build` did not make a difference to the tests, this test:

```js
it('handles circular references in ParseObjects', () => {
  const RealParseObject = jest.requireActual('../ParseObject').default;
  const obj1 = new RealParseObject();
  const obj2 = new RealParseObject();
  obj1.set('obj2', obj2);
  obj2.set('obj1', obj1);

  expect(() => encode(obj1, false, false, [], false)).toThrow();
});
```

Still gives this error:

```js
TypeError: Cannot read properties of undefined (reading 'estimateAttributes')

      153 |   get attributes(): AttributeMap {
      154 |     const stateController = CoreManager.getObjectStateController();
    > 155 |     return Object.freeze(stateController.estimateAttributes(this._getStateIdentifier()));
          |                                          ^
      156 |   }
      157 |
      158 |   /**

      at ParseObject.estimateAttributes [as attributes] (src/ParseObject.js:155:42)
      at ParseObject.attributes [as set] (src/ParseObject.js:747:36)
      at Object.set (src/__tests__/encode-test.js:192:10)
```

mtrezza: Then I guess it's an issue with your test. Why are you using `jest.requireActual`?  I suggest you take an existing test, clone and adapt it.

ajmeese7: Because like I said [here](https://github.com/parse-community/Parse-SDK-JS/pull/2099#discussion_r1537763838):

> The `ParseObject` import is mocked in the `encode` test file

This is how you have `ParseObject` defined in `encode-test.js`:

```JS
const mockObject = function (className) {
  this.className = className;
};
mockObject.registerSubclass = function () {};
mockObject.prototype = {
  _getServerData() {
    return this._serverData;
  },
  toPointer() {
    return 'POINTER';
  },
  toOfflinePointer() {
    return 'OFFLINE_POINTER';
  },
  _getId() {
    return 'local1234';
  },
  dirty() {},
  toJSON() {
    return this.attributes;
  },
  _toFullJSON(seen, offline) {
    const json = {
      __type: 'Object',
      className: this.className,
    };
    for (const attr in this.attributes) {
      json[attr] = encode(this.attributes[attr], false, false, seen.concat(this), offline);
    }
    return json;
  },
};
jest.setMock('../ParseObject', mockObject);
```

mtrezza: Doesn't it work to encode the mocked obj with the pointers set? Or maybe it's easier to test on a higher level. Why not just create a test in `ParseObject-test.js` in which you create a normal Parse Object and set the properties in order to invoke the recursive encode? Did you already find out in which cases encode is called recursively?

mtrezza: See https://github.com/parse-community/Parse-SDK-JS/issues/2098#issuecomment-2027459521 for how you may reproduce this issue.

src/unsavedChildren.js
```diff
@@ -48,8 +50,20 @@ function traverse(
...
   encountered: EncounterMap,
   shouldThrow: boolean,
-  allowDeepUnsaved: boolean
+  allowDeepUnsaved: boolean,
+  counter: number = 0
 ) {
+  counter++;
+
+  if (counter > MAX_RECURSIVE_CALLS) {
+    const message = 'Maximum recursive calls exceeded in traverse function. Potential infinite recursion detected.';
```
mtrezza: Could you simply the error message that is thrown into a single sentence that is "less technical" and easier to understand? In case the error message gets returned to the request response as is, it cannot contain any object information for security reasons; instead you may be able to log objects details internally somehow.

mtrezza (changes requested):

src/__tests__/ParseObject-test.js
```diff
@@ -3862,4 +3863,41 @@ describe('ParseObject pin', () => {
     });
     CoreManager.set('ALLOW_CUSTOM_OBJECT_ID', false);
   });
+
+  it('handles unsaved circular references', async () => {
+    const xhrs = [];
```
mtrezza: Why do you need to mock this?

src/__tests__/ParseObject-test.js
```diff
@@ -3862,4 +3863,41 @@ describe('ParseObject pin', () => {
...
+    const object = new ParseObject('Test');
+    object.set('a', a);
+    expect(() => {
+      object.save();
+    }).toThrowError(
+      'Maximum recursive calls exceeded in traverse function. Potential infinite recursion detected.'
+    );
+  });
+
+  it('throws error for infinite recursion', () => {
```
mtrezza: This should probably move into a separate file for testing the encode file, as this test is unrelated to the higher level Parse Object?

src/unsavedChildren.js
ajmeese7: What are you thinking, something like "Infinite recursion detected in traverse function"?

Also, *cannot contain any object information for security reasons; instead you may be able to log objects details internally somehow* is why I have the `console.error` logging there, that way the server admins (such as myself) can debug what went wrong and see the objects causing these errors. Are you saying you are good with the server-side logging, or that you want it removed?

src/__tests__/ParseObject-test.js
ajmeese7: This was [your suggestion](https://github.com/parse-community/Parse-SDK-JS/pull/2099#discussion_r1544635679):

> Doesn't it work to encode the mocked obj with the pointers set? Or maybe it's easier to test on a higher level. Why not just create a test in ParseObject-test.js in which you create a normal Parse Object and set the properties in order to invoke the recursive encode?

I can *try* to create a new separate file if you think it's necessary, but I know it was having issues in the `encode.js` file as we discussed above.

mtrezza: Every source file usually has an associated test file. If you test on a higher level ParseObject, then you would add the test to `ParseObject-test.js`. If you test on a lower level encode method, then you'd add the test to `encode-test.js`. You have now added two tests, one using Parse Object, another using encode directly. If you believe we need the encode test in addition to the higher level test, then it should be moved into a separate file. If you believe that they both cover the same, then you could just remove the encode test.

src/unsavedChildren.js
mtrezza: Got it, I missed that you are logging part of that in the console. Maybe something like `Encoding object failed due to high number of recursive calls, likely caused by circular reference within object.`?

mtrezza:
After looking at the `encode` method again I wonder whether we really need to introduce a max limit of recursive calls, or whether `encoding` can be "fixed" to properly handle recursive properties. The method already has a `seen` argument which seems to be used for a similar purpose. If an object has a circular reference, then maybe it could just stay circular in its encoded form. That is, if we consider a circular reference to make sense in the first place. In fact, we even have a existing test for circular pointers, `does not stack-overflow when encoding recursive pointers`.

ajmeese7:
@mtrezza found a method that seems to be working on my end with complex objects containing cyclical references, I will be able to fully test by installing locally from my branch later tonight. Will update once I have done so.

Let me know if there's anything else you want changed and I can take a stab at it.

src/CryptoUtils.js
```diff
@@ -0,0 +1,21 @@
...
+    h1 = Math.imul(h1 ^ ch, 2654435761);
+    h2 = Math.imul(h2 ^ ch, 1597334677);
+  }
+  h1  = Math.imul(h1 ^ (h1 >>> 16), 2246822507);
+  h1 ^= Math.imul(h2 ^ (h2 >>> 13), 3266489909);
+  h2  = Math.imul(h2 ^ (h2 >>> 16), 2246822507);
+  h2 ^= Math.imul(h1 ^ (h1 >>> 13), 3266489909);
+
+  return 4294967296 * (2097151 & h2) + (h1 >>> 0);
+};
```
mtrezza, line 21: We cannot accept this algo (or derivations thereof) due to its problematic usage rights situation. The author has claimed to have made this "public domain", but such a claim is invalid in some jurisdictions. I have not found an OS license attributed to this code by the author. If you have found that anywhere, please provide the link.

mtrezza (changes requested):

src/CryptoUtils.js
ajmeese7: Would this solution be satisfactory?

https://github.com/bryc/code/issues/23#issuecomment-2051906976

mtrezza: Thanks for the effort, but unfortunately the underlying issue is still unaddressed. I appreciate the author's rooting for public domain contributions. However, the issue is a legal one. Unless there's _an official OS license_ attributed to the code, we cannot accept it. We also cannot accept any "self-written", non-official OS license on similar grounds.

ajmeese7: @mtrezza is the MIT license addition satisfactory?

mtrezza: Yes, but SO is not sufficient as source or link to the author. Has this been published anywhere on GitHub, and can this be added as a dependency? Because if we add code like this it won't be maintained by the author and the maintenance is on us. I'm not sure whether we should maintain this piece of code if there are libs for hash functions out there that are regularly maintained.

Why is this hash code even required in this PR? Are there no native Node hash functions we can use instead? This looks odd.

src/encode.js
```diff
@@ -9,85 +9,117 @@ import ParsePolygon from './ParsePolygon';
...
-  if (value && typeof value === 'object') {
+  } else if (Array.isArray(value)) {
+    return value.map(v => encode(v, disallowObjects, forcePointers, seen, offline, counter, initialValue));
+  } else if (value && typeof value === 'object') {
     const output = {};
     for (const k in value) {
-      output[k] = encode(value[k], disallowObjects, forcePointers, seen, offline);
+      try {
+        // Attempts to get the name of the object's constructor
+        // Ref: https://stackoverflow.com/a/332429/6456163
```
mtrezza, line 83: Same topic as earlier. If you copy/pasted someone else's code from SO then we cannot accept it.

Edit: Now that I looked at the link, this is a basic JS feature, there is really no need to reference SO here.

However, the above still holds true. I'd ask you to review the whole PR at this point and identify any code that has been copied from somewhere else, so we can investigate whether we can accept it.

mtrezza (changes requested):

mtrezza (commented):
I have more and more doubts that we can accept this PR without a performance comparison. _Encode_ is such a central code that any changes here can have a significant yet subtle impact on Parse Server performance. Perf issues are tough to analyze after the fact, so I'm somewhat hesitant.

ajmeese7:
A fix might still be worth implementing here, but I don't see myself having the cycles to work on this PR in the near future.

HOWEVER, I *did* figure out a cause of all these errors in the legacy code I inherited, if it helps anyone else. The cloud functions were returning the response of an Axios GET request, and the response object has infinite recursion in its properties with sockets. What I had to do was just parse out the `data.result` properties from the response instead of sending them across the wire with Parse Server, then I was good to go :)

Feel free to close this out if you want, so someone else can come in with a different implementation that you would prefer.

mtrezza:
It would be too bad, since you've already put so much work into it. Thinking about it, adding the recursion limit that you implemented initially may be the least intrusive one and I'd think poses the least risk to performance. It's the simplest fix. But only if we set the limit ridiculously high (10k?), we can treat it as a non-breaking change, if we assume that nested objects with 10k levels deep are rather unlikely.

On the other hand the solution you implemented would not require the limit and properly handle recursions, as we already do in other parts of the code. But with hash generation, etc we run into perf issues, so we may need a perf comparison test. In fact we have https://github.com/parse-community/parse-server/issues/7610 for that.

mtrezza:
Picking this up, I think there are 2 ways to move forward:

1) Merge the simplified version where the recursion is simply counted and limited to a fixed amount. This is unlikely to have any significant perf impact, as it's just a var counter. If we set the recursion count limit high enough we also don't have to consider it a breaking change. Probably up to https://github.com/parse-community/Parse-SDK-JS/pull/2099/commits/d33a54a9d6097863aa042dd382250e346b0acf50.

2) Properly handle recursion similar to how recursion is [currently](https://github.com/parse-community/Parse-SDK-JS/pull/2099#issuecomment-2030896912) handled in that part of the code. That would be similar to what the current state of the PR is, but requires addressing open issues such as [hash algo](https://github.com/parse-community/Parse-SDK-JS/pull/2099#discussion_r1564135116).

I would be hesitant to merge (2) without having a perf comparison in our CI, because of https://github.com/parse-community/Parse-SDK-JS/pull/2099#pullrequestreview-1999317003. So for now we could go ahead and merge (1), so we at least fix the issue of infinite recursion.

PR #2106, opened 2024-04-13
feat: Allow setting custom queue for handling offline operations via `Parse.EventuallyQueue`

mortenmo:
EventuallyQueue behavior isn't always what I need. Was hoping to have a way to replace the default to change the behavior. Using CoreManager felt like the obvious way.

## Issue

EventuallyQueue gives you no control over what happens if there are conflicts. For example,  another client has changed the object, it just looks at updatedAt time and drops the local changes. Fine as a general rule, but sometimes I need some control over that in an often offline environment.

## Approach

There were only 3 places that called EventuallyQueue directly. Just changed those to use `CoreManager.getEventuallyQueue()` following the patterns of most other things going through CoreManager.

## Tasks

- [X] Fix tests if needed
- [] Documentation??


dplewis:
Looks good! I think you may have fixed a circular dependency as well.

src/Parse.ts
```diff
@@ -380,7 +382,7 @@ const Parse: ParseType = {
     if (!this.LocalDatastore.isEnabled) {
       this.LocalDatastore.isEnabled = true;
       if (polling) {
-        EventuallyQueue.poll(ms);
+        CoreManager.getEventuallyQueue().poll(ms);
```
dplewis, line 411: @mortenmo `Parse.EventuallyQueue` will have to get retrieved from the CoreManager now instead of the `EventuallyQueue` import.

```
/**
   * @member {EventuallyQueue} Parse.EventuallyQueue
   * @static
   */
  set EventuallyQueue(queue: EventuallyQueue) {
    CoreManager.setEventuallyQueue(queue);
  },
  get EventuallyQueue() {
    return CoreManager.getEventuallyQueue();
  },

**Review by dplewis — CHANGES_REQUESTED** [review]

File: src/Parse.ts (hunk 1 above)
Comment (mortenmo) [reply src/Parse.ts:411]: oh yeah, good catch, forgot about accessing it that way.

Comment (dplewis) [reply src/Parse.ts:411]: I just did the same thing for Parse.LiveQuery if you want to check it out https://github.com/parse-community/Parse-SDK-JS/pull/2100/commits/2d838b54b4c1e13668d6c85f3e88bf3810ea1002

Comment (mortenmo) [reply src/Parse.ts:411]: committed your suggestion..

 I noticed this code in `_initialize`: 
```
    Parse.LiveQuery = new LiveQuery();
    CoreManager.setIfNeeded('LiveQuery', Parse.LiveQuery);
```

Has some of the same issue. If you did set your own `LiveQuery`, `Parse.LiveQuery` wouldn't reflect the same implementation. Never tried to do that myself or know if there is any bad sideeffects of it.

Comment (dplewis) [reply src/Parse.ts:411]: Look at my last comment https://github.com/parse-community/Parse-SDK-JS/pull/2106#discussion_r1564373238

Comment (mortenmo) [reply src/Parse.ts:411]: perfect, github didnt update fast enough for me to see your comment. thanks for the help!

**Review by dplewis — APPROVED** [review]

**Comment by dplewis** [comment]
@mtrezza This is ready for review. I'm not sure why the CI isn't running

File: src/Parse.ts [hunk 2]
```diff
@@ -151,6 +150,18 @@ const Parse: ParseType = {
   Hooks: undefined,
   Parse: undefined,
 
+  /**
+   * @member {EventuallyQueue} Parse.EventuallyQueue
+   * @static
+   */
+  set EventuallyQueue(queue: EventuallyQueue) {
```
dplewis, line 159: @mortenmo Can you add a unit test for this? `npm run test` you should see code coverage drop as this setter/getter isn't tested.

dplewis (changes requested):

mtrezza:
Does the PR title make sense?

mtrezza (changes requested):
CI fails

mtrezza (approved):
Looks good, is this ready for merge?

mortenmo:
Yeah I believe so

dplewis (approved):

parseplatformorg:
🎉 This change has been released in version [5.1.0-alpha.6](https://github.com/parse-community/Parse-SDK-JS/releases/tag/5.1.0-alpha.6)

parseplatformorg:
🎉 This change has been released in version [5.1.0-beta.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/5.1.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [5.1.0](https://github.com/parse-community/Parse-SDK-JS/releases/tag/5.1.0)

Issue #1450, opened 2022-02-15
All updates to an object field except for the last one are ignored


mstniy:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-JS/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse JS SDK](https://github.com/parse-community/Parse-SDK-JS/releases). 

### Issue Description

When updating multiple nested fields under the same object field using `ParseObject.set`, all updates except for the last one are dropped.

### Steps to reproduce

```
const obj = new Parse.Object("MyObject");
obj.set("a",{});
await obj.save();
obj.set("a.b", 5);
obj.set("a.c", 6);
```

### Actual Outcome

`obj.attributes` should be `{a: {b:5, c:6}}`

### Expected Outcome

`obj.attributes` is `{a: {c:6}}`

###  Environment

Server
- Parse Server version: `irrelevant`
- Operating system: `Ubuntu 20.04`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `local, irrelevant`

Database
- System (MongoDB or Postgres): `MongoDB`
- Database version: `5.0.6`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `local`

Client
- Parse JS SDK version: `alpha (commit d084ac9be058d9b13f8ac2bbf04615f76dc2800d)`

parseplatformorg:
🎉 This change has been released in version [5.1.0-alpha.7](https://github.com/parse-community/Parse-SDK-JS/releases/tag/5.1.0-alpha.7)

parseplatformorg:
🎉 This change has been released in version [5.1.0-beta.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/5.1.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [5.1.0](https://github.com/parse-community/Parse-SDK-JS/releases/tag/5.1.0)

PR #2116, opened 2024-04-27
fix: Local state wrong on multiple set operations on same property

### LINKED ISSUE #2115: Local state of objects not correct if setting multiple properties
### New Issue Checklist

- [X] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [X] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-JS/issues?q=is%3Aissue).
- [X] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse JS SDK](https://github.com/parse-community/Parse-SDK-JS/releases). 

### Issue Description

When looking at #1798 found that setting multiple properties on the same object locally without saving reports the wrong object back. It will only show (this is true for set and unset) the latest Op you added to that property and not all. So setting `data.b` and `data.c` without saving (or you're offline), causes `data.b` to not be set.

When the object gets saved, the properties show up on get as expected.

### Steps to reproduce

```
const myObj = new Parse.Object('MyClass');
myObj.set('data', {a: 5});
await myObj.save(null, {useMasterKey: true});

myObj.set('data.b', 2);
myObj.set('data.c', 2);

console.log(myObj.get('data'));
```

### Actual Outcome

```
{ a: 5, c: 2}
```

### Expected Outcome
```
{ a: 5, b: 2, c: 2}
```

###  Environment

Server
- Parse Server version: `7.0.0`
- Operating system: `OSX`

Client

mortenmo:
## Pull Request

## Issue

Closes: https://github.com/parse-community/Parse-SDK-JS/issues/2115
Closes: https://github.com/parse-community/Parse-SDK-JS/pull/1451

## Approach
When local state has multiple set/unset ops on the same base property, use the result of the previous instead of always going back to the server version for every Op to allow them to chain.

The test included fails until the fix is in place.

## Tasks

- [X] Add tests


dplewis:
@mtrezza @mortenmo A PR for this has already been opened. Try copying the test over from that PR. I'll review.

https://github.com/parse-community/Parse-SDK-JS/pull/1451

mortenmo:
I really need to check existing PRs better :) . The fixes looks basically identical, so feel free to commit that one and I can close this one. I added the #1450 test case here just to be sure and it does also pass.

dplewis:
@mortenmo the dot notation / nested fields was an invalid field before I removed it. There are some issues with this but an understanding of the internal state of objects is required.

This PR looks good 👍

dplewis:
@mstniy Can you review this? You know more about the internals than anyone

mstniy:
Hey. I made some changes to #1451 . The patch also handles doubly-nested objects now, it used to be still broken in that case. I also modified the regression test to also include this scenario.

mtrezza:
Is this PR made obsolete by https://github.com/parse-community/Parse-SDK-JS/pull/1451? If not, which one should be merged first, in case order matters?

mstniy:
I believe so. I copied the new regression test in this branch over to my branch, and it does pass there also.

I would suggest #1451 to be merged in favor of this one, as that one implements a more robust deep cloning logic after today's changes, evidenced by the regression test here passing there, but not vica versa.

mtrezza:
Superseded by https://github.com/parse-community/Parse-SDK-JS/pull/1451

PR #2117, opened 2024-04-29
fix: Setting and unsetting property without saving `Parse.Object` causes server error

### LINKED ISSUE #1798: Internal server error on set followed by unset
### New Issue Checklist

- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description

Setting a subdocument entirely, followed by an `unset` of a nested field causes an internal server error while the object is being saved, if the object existed beforehand.

### Steps to reproduce

```
const myObj = new Parse.Object('MyClass');
await myObj.save(null, {useMasterKey: true});
myObj.set('data', {a: 5});
myObj.unset('data.a');
await myObj.save(null, {useMasterKey: true});
console.log('a');
```

### Actual Outcome

```
error: Uncaught internal server error. Updating the path 'data.a' would create a conflict at 'data' {"code":40,"codeName":"ConflictingUpdateOperators","ok":0,"stack":"MongoServerError: Updating the path 'data.a' would create a conflict at 'data'\n    at Connection.onMessage (/home/kartal/Desktop/ocell/parse-server-example/node_modules/parse-server/node_modules/mongodb/lib/cmap/connection.js:202:30)\n    at MessageStream.<anonymous> (/home/kartal/Desktop/ocell/parse-server-example/node_modules/parse-server/node_modules/mongodb/lib/cmap/connection.js:62:60)\n    at MessageStream.emit (events.js:400:28)\n    at processIncomingData (/home/kartal/Desktop/ocell/parse-server-example/node_modules/parse-server/node_modules/mongodb/lib/cmap/message_stream.js:108:16)\n    at MessageStream._write (/home/kartal/Desktop/ocell/parse-server-example/node_modules/parse-server/node_modules/mongodb/lib/cmap/message_stream.js:28:9)\n    at writeOrBuffer (internal/streams/writable.js:358:12)\n    at MessageStream.Writable.write (internal/streams/writable.js:303:10)\n    at Socket.ondata (internal/streams/readable.js:731:22)\n    at Socket.emit (events.js:400:28)\n    at addChunk (internal/streams/readable.js:293:12)"}
MongoServerError: Updating the path 'data.a' would create a conflict at 'data'
    at Connection.onMessage (/home/kartal/Desktop/ocell/parse-server-example/node_modules/parse-server/node_modules/mongodb/lib/cmap/connection.js:202:30)
    at MessageStream.<anonymous> (/home/kartal/Desktop/ocell/parse-server-example/node_modules/parse-server/node_modules/mongodb/lib/cmap/connection.js:62:60)
    at MessageStream.emit (events.js:400:28)
    at processIncomingData (/home/kartal/Desktop/ocell/parse-server-example/node_modules/parse-server/node_modules/mongodb/lib/cmap/message_stream.js:108:16)
    at MessageStream._write (/home/kartal/Desktop/ocell/parse-server-example/node_modules/parse-server/node_modules/mongodb/lib/cmap/message_stream.js:28:9)
    at writeOrBuffer (internal/streams/writable.js:358:12)
    at MessageStream.Writable.write (internal/streams/writable.js:303:10)
    at Socket.ondata (internal/streams/readable.js:731:22)
    at Socket.emit (events.js:400:28)
    at addChunk (internal/streams/readable.js:293:12)
```
### Expected Outcome

The object should be saved to the database with data = `{}`
### Environment

Server
- Parse Server version: `5.3.0`
- Operating system: `ubuntu 20.04`

Database
- Database version: `5.0.13`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `parse dashboard`
- SDK version: `4.1.4`

mortenmo:
## Pull Request

## Issue

Closes: https://github.com/parse-community/Parse-SDK-JS/issues/1798

## Approach
On the path of making my app work well offline, this was one more blocker.

The problem is that if you set a parent object `o.set('data', {a: b, b: 3})` and then set or unset anything on `data` (example: `o.unset('data.a')` without saving between (or save failed for no connection), you get a server error on `save()`. 1798 indicates this needs to be fixed on client. What fails seems to be mongo doesn't support setting an object and then sub-properties in the same call which is what the server attempts to do in this case.

This fix merges Ops if a "parent" Op exists locally and is not changed and applies the new Op to that parent. In the end this means that when sent to server it wont create the issue above and succeed even if subproperties were set/unset/incremented as a part of local operation before save. 

Small(?) change to `ParseObject.set` with calling `validPendingParentOp` that if the current Op attribute has dots (`.`) it will look for pending parent Operations. If one is found, `applyOpToParent` is called taking the new Op and applying the change to it locally.  Put these new functions in `ParseOp`, but that might not be the best place for them (maybe ObjectStateController?).

## Tasks

- [X] Add tests
- [x] Add test to validate server can save properly now


dplewis:
Can you check code coverage? Looks like you are missing a test.

mortenmo:
Another alternative solution that might be preferred is to put a similar fix into `ParseObject._getSaveJSON()` instead? That means the Ops array stays the same (which is actually what the tests should check I'll fix that, that's in the end what the server cares about). But the data is merged in the saveJSON only.

dplewis:
@mortenmo Good Work! Can you add an integration test to make sure it's not crashing?

mortenmo:
> @mortenmo Good Work! Can you add an integration test to make sure it's not crashing?

Yes. Took a subset of related testcases and made versions that don't save between the object creation and setting/unsetting making sure the results are the same. Just had to figure out how to run them :)

mortenmo:
@dplewis type check was failing I assume from the changing CoreManager to ts. I put in a fix here but have to back that out if you fixed it somewhere else as well.

dplewis:
Yeah sorry for the confusion. Can you revert changes unrelated to the issue? This PR looks good to me. I just to test it myself. Luckily I’ve been looking into dot notation and the internals recently so it shouldn’t take too long

dplewis (approved):
LGTM!

mtrezza (dismissed):

dplewis:
@mortenmo Can you rebase from alpha? ParseObject was converted to typescript and is causing conflicts in this PR.

mortenmo:
Made some git mistakes that my git skillz couldn't get me out of, but should be good now.

dplewis (approved):
Nice Git skillz!

mtrezza (approved):

mtrezza:
Isn't there more to this issue? In https://github.com/parse-community/Parse-SDK-JS/issues/1798 it's described that it causes an internal server error, which sounds like Parse Server needs a patch as well, to that doesn't crash on malformed request?

mortenmo:
> Isn't there more to this issue? In #1798 it's described that it causes an internal server error, which sounds like Parse Server needs a patch as well, to that doesn't crash on malformed request?

@mtrezza I don't think so. The reason the server "barfed" on the Ops is that Mongo (maybe also Postgres?) doesn't support chained updates when you changed the "parent". When you change/set a property (eg `object`) to an object and then do a set `object.foo` to something in the same save, Mongo threw an error. This PR merges those on the client before sending to server (whether it is `set` or `unset`), so the server doesn't get those Op commands anymore. 

It is unclear to me if the server _should_ support an operation to `set object={ ... }` and `set object.name='foo'` in the same save or if the clients should avoid it. It is only a problem if you set a parent and then a "child" property.  If you think the server should, then yes there should be another issue opened on the server. But the JS SDK shouldn't trigger that problem anymore, but others might?

mtrezza:
I mean it should not be possible to cause the server to crash with internal server error. Even if the Parse JS SDK doesn't send this anymore, it would still be possible to send this to Parse Server via the REST API and cause the crash, right?

mortenmo:
When I tried it manually, server does respond with a 500, but to be clear doesn't completely crash. But yes, it is possible to send through and at a minimum a validation server side with a 400 response is probably better.

mtrezza:
I believe a 500 error may actually be a process crash.

PR #2141, opened 2024-05-17
style: Fix pre-commit hook and enable Prettier for TypeScript

dplewis:
## Pull Request

## Issue

After Husky updated to v8 it has been broken.

## Approach

Run the typescript parser for prettier, style and lint TS files. Prettier now runs before every commit and enforces the styling convention. No functional changes have been made.

## Tasks

- [x] Add tests

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,2 @@
+npm run prettier
+git add .
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,2 +1,1 @@
 npm run prettier
-git add
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,1 @@
-npm run prettier
+npm run pre-commit
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,2 +1,2 @@
 npm run prettier
-git add .
+git add
--- a/.prettierrc
+++ b/.prettierrc
@@ -3,3 +3,4 @@ trailingComma: "es5"
 singleQuote: true
 arrowParens: "avoid"
 printWidth: 100
+parser: "typescript"
--- a/integration/test/ParseDistTest.js
+++ b/integration/test/ParseDistTest.js
@@ -52,18 +52,21 @@ for (const fileName of ['parse.js', 'parse.min.js']) {
         request.continue();
       });
       page.on('requestfailed', request => {
-        if (request.failure().errorText  === 'net::ERR_ABORTED' && !request.url().includes('favicon.ico')) {
+        if (
+          request.failure().errorText === 'net::ERR_ABORTED' &&
+          !request.url().includes('favicon.ico')
+        ) {
           abortedCount += 1;
           promise.resolve();
         }
       });
       await page.evaluate(async () => {
         const parseLogo =
-        'https://raw.githubusercontent.com/parse-community/parse-server/master/.github/parse-server-logo.png';
+          'https://raw.githubusercontent.com/parse-community/parse-server/master/.github/parse-server-logo.png';
         const file = new Parse.File('parse-server-logo', { uri: parseLogo });
         file.save().then(() => {});
 
-        return new Promise((resolve) => {
+        return new Promise(resolve => {
           const intervalId = setInterval(() => {
             if (file._requestTask && typeof file._requestTask.abort === 'function') {
               file.cancel();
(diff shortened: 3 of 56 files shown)


.husky/pre-commit
```diff
@@ -0,0 +1,2 @@
+npm run prettier
+git add .
```
mtrezza: Really?

mtrezza (commented):
Does this pre-commit hook run prettier and then lint fix? The 2 produce different outputs.

.husky/pre-commit
dplewis: This really works

dplewis: Lint fix is required as prettier doesn't follow convention for switch statements and a few other rules. The combination of the two produces the best quality code. It's been this way for years until the pre-commit hook broke.

mtrezza: I was actually referring to the `git add .`, should this really be added? Doesn't that stage all changes for the commit, regardless of whether the developer wants to commit all changes?

mtrezza:
> Does this pre-commit hook run prettier and then lint fix? The 2 produce different outputs.

I see now this is running the prettier script which also runs lint fix; I assume it would only run prettier. All good.

.husky/pre-commit
dplewis: You are right, I've changed it

mtrezza: I don't believe that `git add` does anything. In fact, how do you distinguish between changes by prettier and other changes by the developer they don't want to commit?

dplewis: `git add` doesn't do anything, I've removed it. Changes made by prettier will be unstaged just like other changes the developer don't want to commit.

mtrezza: I believe you'd need to stash all unstaged changes, then run prettier, then stage all changes, then restore stashed changes.

Another option is to abort the commit if lint encounters errors. That's a horrible DX, hence I'm not really a fan of these precommit enforcements.

dplewis: @mtrezza That is exactly what happens minus the restore stashed part. It's working pretty good now that the changes have been made.

dplewis: > Another option is to abort the commit if lint encounters errors. That's a horrible DX, hence I'm not really a fan of these precommit enforcements.

The commit is aborted if there are linting issues. The CI is going to fail if there are lint issues anyway. Also this precommit hook will be used to autogenerate typescript declaration files.

mtrezza: This happens with what commands in the hook? Why did you add `git add .` originally, for prettier or TS?

dplewis: Basically copied the previous implementation https://github.com/parse-community/Parse-SDK-JS/blob/alpha/package.json#L132

I wanted to add the linted changes to the commit.

dplewis: >This happens with what commands in the hook?

Sorry I mean if prettier itself fails as in an `unexpected token error` would cause the commit to fail.

mtrezza: Yeah; I don't think a failing lint should prevent a commit. I believe the hooks should assist the developer, not impede them while they are experimenting with commits. Until a PR is for review, linting is secondary. That's what we have the CI check for.

dplewis: How would you get a failing lint with this setup of prettier + lint:fix?

mtrezza: Got it, it wouldn't fail, but it would restyle the code, right? I personally don't like code being restyled while I'm still working on it. That's just my personal preference. How are we doing this in our other major repos, do they all have precommit restyling, or is this repo an exception?

dplewis: Parse Server has precommit hooks for styling, the Docs repo has a hook for bundling assets.

mtrezza: Well let's align with parse server then and keep the precommit styling. Is that hook active or also inactive like in this repo due to the upgrade?

So the commit will always pass but may make style changes? How do we commit these changes, is that done automatically?

dplewis: It will happen automatically, I changed it back to be close to the original implementation and align with parse server.

dplewis:
@mtrezza I made some changes to bring the hook closer to the original implementation. This is ready for review.

mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [5.2.0-alpha.2](https://github.com/parse-community/Parse-SDK-JS/releases/tag/5.2.0-alpha.2)

parseplatformorg:
🎉 This change has been released in version [5.2.0-beta.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/5.2.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [5.2.0](https://github.com/parse-community/Parse-SDK-JS/releases/tag/5.2.0)

PR #2164, opened 2024-06-07
feat: Enable `Parse.idempotency` by default

dplewis:
## Pull Request

## Issue

There isn't a well documented way to turn idempotency on, enabling it by default only requires configuration on the server side. This is how it is handled in other SDKs

https://github.com/parse-community/Parse-Swift/pull/62
https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1790
https://github.com/parse-community/Parse-SDK-Android/pull/1190

## Approach
* Remove check for `POST` / `PUT` request method as they are handled by servers
* Add requestId to every request by default

## Tasks

- [x] Add tests

### DIFF
--- a/integration/test/IdempotencyTest.js
+++ b/integration/test/IdempotencyTest.js
@@ -22,7 +22,12 @@ function DuplicateXHR(requestId) {
 
 describe('Idempotency', () => {
   beforeEach(() => {
+    Parse.idempotency = false;
     RESTController._setXHR(XHR);
+  });
+
+  afterEach(() => {
+    Parse.idempotency = true;
   });
 
   it('handle duplicate cloud code function request', async () => {
--- a/src/CoreManager.ts
+++ b/src/CoreManager.ts
@@ -325,7 +325,7 @@ const config: Config & { [key: string]: any } = {
   PERFORM_USER_REWRITE: true,
   FORCE_REVOCABLE_SESSION: false,
   ENCRYPTED_USER: false,
-  IDEMPOTENCY: false,
+  IDEMPOTENCY: true,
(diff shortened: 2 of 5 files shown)


mtrezza:
- This would likely be a breaking change for the JS SDK, since it changes the request headers.
- What's the purpose of turning this on by default? Is that on be default in the other client SDKs?

dplewis:
This should have been enabled from the beginning. This is the default behavior of the other SDKs as I mentioned in the OP.

I wouldn't consider this a breaking change as wouldn't cause any disruptions to the developers. Adding this request header by default is pretty much useless unless the feature is enabled on the server. If it’s enabled on the server already it’s likely enabled in the SDK

A breaking change would be removing the ability to turn it off.

mtrezza:
I'm not sure it should be on by default. That's at least not what I had in mind when I designed the feature. It's costly on server side resources when it's enabled on the server and a minority of users is likely using it. Sending a UID by default from the client to the server without the server using it would not be good practice because that generates unnecessary costs.

It could be breaking considering that we change the request headers. An additional header may cause a request to be rejected on tightly configured firewalls. It's like changing a request HTTP method, that would also be a breaking change for the same reason.

dplewis:
@mtrezza Im closing this as it works for me. From a client side cost side we should purge the other SDKs headers as well from the parse.com days. Can you add documentation so developers can enable this feature since designed it?

mtrezza:
I agree that other client SDKs should also not send the idempotency header by default. For docs see [here](https://github.com/parse-community/parse-server?tab=readme-ov-file#idempotency-enforcement).

PR #2425, opened 2025-02-02
feat: Add vite

dblythy:
## Pull Request

- Report security issues [confidentially](https://github.com/parse-community/Parse-SDK-JS/security/policy).
- Any contribution is under this [license](https://github.com/parse-community/Parse-SDK-JS/blob/alpha/LICENSE).
- Link this pull request to an [issue](https://github.com/parse-community/Parse-SDK-JS/issues?q=is%3Aissue).

## Issue

Closes: #2410

## Approach

Vite is more modern than webpack, so approaching building files with vite. Testing will still be needed

## Tasks

- [ ] Add tests
- [ ] Add changes to documentation (guides, repository pages, code comments)

PR #2435, opened 2025-02-09
ci: Remove dtslint

dblythy:
## Pull Request

## Issue

Remove types check; reason see https://github.com/parse-community/parse-server/pull/9581.

## Tasks

- [ ] Add tests


.github/workflows/ci.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - name: Build Types
         run: npm run build:types
       - name: Test Types
-        run: npm run test:types
+        run: npm run build:types
```
dplewis: This is a duplicate there is a job “build:types” already

package.json
```diff
@@ -102,7 +101,6 @@
     "test": "cross-env PARSE_BUILD=node jest",
     "test:mongodb": "npm run test:mongodb:runnerstart && npm run integration",
     "test:mongodb:runnerstart": "mongodb-runner start -- --port 27017",
-    "test:types": "dtslint --expectOnly types",
```
dplewis, line 105: I think I figured out how to enable full typescript support with shipping types to npm. I was waiting for #2425 to be merged so we don’t have conflicts

.github/workflows/ci.yml
mtrezza: removed

.github/workflows/ci.yml
```diff
@@ -29,8 +29,6 @@ jobs:
       - run: npm ci
       - name: Build Types
         run: npm run build:types
-      - name: Test Types
-        run: npm run test:types
       - name: Lint Types
```
mtrezza, line 32: There is a `lint:types` step here which we don't seem to have in parse-server; what is that for if the types are auto-generated anyway? should we remove it from here, or add it to parse-server as well?

dplewis: This repo uses typescript the server uses flow type

mtrezza: Can/should that be unified?

dplewis: They could be merged however it’s much harder to debug

dplewis:
@dblythy the reason why dtslint is here is because we haven’t moved away from definitelyTyped yet. Once we move away then this can be removed

.github/workflows/ci.yml
mtrezza: Let's leave as is

mtrezza:
@dplewis I've added the link in the issue description for why we're removing this. The package doesn't seem to be intended to be used that way and we ran into issues in Parse Server. If you follow the thread you'll find people hacking stuff together to make this work somehow. Even though it seems to work in this repo for now, we're removing it preemptively to rely solely on the types build step not to fail for the CI job to pass.

dplewis:
@mtrezza go for it

mtrezza:
Do you have any concerns?

dplewis:
@dblythy @mtrezza I'd recommend keeping this until someone can get the type test suite to pass. There is a `Cannot find namespace 'Parse'` error when using at the top level that is stopping a typescript release.

Here is the test suite for reference copied from DT
https://github.com/parse-community/Parse-SDK-JS/blob/alpha/types/tests.ts

Edit: It should be easy to replace `dtslint`with [eslint-plugin-expect-type](https://github.com/JoshuaKGoldberg/eslint-plugin-expect-type)

dplewis:
@dblythy I created a PR https://github.com/parse-community/Parse-SDK-JS/pull/2452 to show case my last comment. I agree dtslint should be removed in the future.

mtrezza:
PR on hold; see https://github.com/parse-community/Parse-SDK-JS/pull/2452#issuecomment-2661633477

parseplatformorg:

PR #2441, opened 2025-02-11
fix: `Parse.Hooks` requests have double forward slash in URL

### LINKED ISSUE #2406: Parse Hooks requests have double forward slash in URL (which is an issue for Express 5.x)
### New Issue Checklist
- [X] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [X] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-JS/issues?q=is%3Aissue).
- [X] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse JS SDK](https://github.com/parse-community/Parse-SDK-JS/releases). 

### Issue Description
An Express 4.x oddity allowed requests to be made with a double forward slash in the URL if it was between the router and route ([discussed here](https://github.com/expressjs/express/issues/4427)). This is an anti-pattern and shouldn't be allowed. It looks like because of the way some routes are created in this SDK, a double forward slash is automatically being added to some internal requests. This requires us to make Parse Server continue supporting double forward-slashes even though it's really not advisable.

It's fairly easy to track the issue down in the code. You can see paths that start with forward slashes here:
https://github.com/parse-community/Parse-SDK-JS/blob/alpha/src/ParseHooks.ts#L77

And then the REST controller checks if the Server URL ends with a forward slash and, if it doesn't it adds one before tacking on the path:
https://github.com/parse-community/Parse-SDK-JS/blob/alpha/src/RESTController.ts#L235

### Steps to reproduce
Run any call that triggers an internal REST call to Hooks or (I think) Files and observe the URL used internally. You will see something like this:
```
http://localhost:1337/parse//hooks/functions
```

I discovered this running Parse Server Hooks integration tests so that's another way to reproduce this issue.

### Actual Outcome
Doubled up forward slashes.

### Expected Outcome
The SDK should not be doubling up the forward slashes.

Server
- Parse Server version: `7.4.0`
Client
- Parse JS SDK version: `5.3.0`

dplewis:
## Pull Request

## Issue

Closes: https://github.com/parse-community/Parse-SDK-JS/issues/2406

## Approach

Remove extra slash

## Tasks

- [x] Add tests

### DIFF
--- a/src/__tests__/Hooks-test.js
+++ b/src/__tests__/Hooks-test.js
@@ -98,23 +98,23 @@ describe('Hooks', () => {
   });
 
   it('should properly build POST function', async () => {
-    Hooks.createFunction('myFunction', 'https://dummy.com');
+    Hooks.createFunction('myFunction', 'https://example.com');
 
     expect(CoreManager.getHooksController().sendRequest.mock.calls[0]).toEqual([
       'POST',
       'hooks/functions',
       {
         functionName: 'myFunction',
-        url: 'https://dummy.com',
+        url: 'https://example.com',
       },
     ]);
     defaultController.sendRequest = sendRequest;
-    await Hooks.createFunction('myFunction', 'https://dummy.com');
+    await Hooks.createFunction('myFunction', 'https://example.com');
     expect(RESTController.ajax.mock.calls[0][1]).toBe('https://api.parse.com/1/hooks/functions');
   });
 
   it('should properly build POST trigger', async () => {
-    Hooks.createTrigger('MyClass', 'beforeSave', 'https://dummy.com');
+    Hooks.createTrigger('MyClass', 'beforeSave', 'https://example.com');
 
     expect(CoreManager.getHooksController().sendRequest.mock.calls[0]).toEqual([
       'POST',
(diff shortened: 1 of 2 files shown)


mtrezza (approved):

mtrezza:
Could this possibly be a breaking change, thinking about custom parsing of the URI on the server or firewall?

dplewis:
@pocketcolin

mtrezza:
We might want to merge this as a breaking change, make a stable release and then add that to the Parse Sever 8 release as a breaking change. Even though these are internal requests and therefore may be less likely to be restricted by a firewall, it's still possible.

src/__tests__/Hooks-test.js
```diff
@@ -19,118 +33,169 @@ describe('Hooks', () => {
...
-      '/hooks/triggers',
+      'hooks/triggers',
       {
         className: 'MyClass',
         triggerName: 'beforeSave',
         url: 'https://dummy.com',
       },
     ]);
+    defaultController.sendRequest = sendRequest;
+    await Hooks.createTrigger('MyClass', 'beforeSave', 'https://dummy.com');
```
mtrezza:
```suggestion
    await Hooks.createTrigger('MyClass', 'beforeSave', 'https://example.com');
```

We are using `example.com` for all example domains in tests and docs (RFC2606).

mtrezza (changes requested):

src/__tests__/Hooks-test.js
dplewis: Done

mtrezza (approved):

mtrezza:
BREAKING CHANGE: Internal REST requests for `Parse.Hooks` use a URL that contains a double forward slash, for example `http://localhost/parse//hooks/functions`. This release changes the double forward slash to a single forward slash.

@dplewis Is this correct?

dplewis:
LGTM!

parseplatformorg:
🎉 This change has been released in version [6.0.0-alpha.2](https://github.com/parse-community/Parse-SDK-JS/releases/tag/6.0.0-alpha.2)

parseplatformorg:

Issue #1386, opened 2021-07-13
Unhandled Promise rejection for ParseUser save()


tboily:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-JS/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse JS SDK](https://github.com/parse-community/Parse-SDK-JS/releases). 

### Issue Description

If an error is thrown by Parse-Server on a Parse.User.save(), the SDK does not catch it and an Unhandled Promise Rejection error is raised, making it impossible for our app to catch the error.

There may be some sort of wrapper I am not aware of, but I gave a quick look to the code and to the untrained eye, it simply looks as though there is no try-catch block in the save() method of ParseUser (`src/ParseUser.js`). Same goes for other methods like `destroy()`.

According to logs, error is raised at ParseObject.js:3137, but the absence of ParseUser in the error stack gives me the feeling it is indeed the culprit.

### Steps to reproduce

In our case, we did as follows:
Update a user with an email already used and call save().
```
// user is an instance of Parse.User()
try {
    user.email = 'alreadyInDb@test.com';
    await user.save()
} catch (err) {
    // We know there is an error but we never end up here because user's save() does not throw.
    throw err;
}
```

### Actual Outcome

Logs show an unhandled promise rejection, and nothing is caught in our catch block.

### Expected Outcome

The SDK should catch the error and throw it back to the app.

###  Environment

nodejs 12.16.1
parse 3.3.0
parse-server 4.5.0

We were working with earlier versions of parse and parse-server before and updated to the latest versions to see if this error was fixed.

Server
- Parse Server version: `4.5.0`
- Operating system: `Google App Engine (some version of Linux)`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `Google Cloud`

Database
- System (MongoDB or Postgres): `MongoDB`
- Database version: `4.4.6`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `MongoDB Atlas`

Client
- Parse JS SDK version: `3.3.0`

[...]

PR #2445, opened 2025-02-12
fix: Remove validation error handler option `error` from various methods of `Parse.Object`

### LINKED ISSUE #2104: `Parse.Object.set` sliently handles validation errors
### New Issue Checklist

### Issue Description

You can pass an option in set to handle validation errors. The problem is this option isn't well documented and isn't widely used. This method of error handling is called backbone and was removed in 2.0.0.

Removing this in favor of an thrown error is a breaking change as a lot of code uses `.set`. The main errors are for invalid key names and invalid ACL.

Ref: https://github.com/parse-community/Parse-SDK-JS/pull/2097

### Steps to reproduce

```
object.set('$$$', 'o_O', {
  error: function (thisObj, err) {
    // err is ParseError.INVALID_KEY_NAME but doesn't have an error message.
  },
});
```

### Actual Outcome

If option not passed, error is silently handled

### Expected Outcome

An Error should be thrown

###  Environment

Server
- Parse Server version: `6.1.0`
- Operating system: `MacOS`

Database
- System (MongoDB or Postgres): `Mongo`
- Database version: `6.0.29`

Client

dplewis:
## Pull Request

## Issue

You can pass an option in `set` to handle validation errors. The problem is this option isn't well documented and isn't widely used. This method of error handling is called backbone and was removed in 2.0.0. There are two errors (invalid key name and invalid acl) that are being checked. This could potentially be a breaking change if a user is setting invalid fields and never saving as saving would throw the errors.

```
object.set('$$$', 'o_O', {
  error: function (thisObj, err) {
    // err is ParseError.INVALID_KEY_NAME but doesn't have an error message.
  },
});
```
Closes: https://github.com/parse-community/Parse-SDK-JS/issues/2104

## Approach

* Remove backbone `options.error`
* Add error message for `Parse.Error.INVALID_KEY_NAME`
* Improve documentation for chaining
* `set` now throws validation errors

## Tasks

- [x] Add tests


mtrezza:
Now is a good time to merge any breaking changes. We can release with Parse Server 8.

integration/test/ParseACLTest.js
```diff
@@ -10,7 +10,9 @@ describe('Parse.ACL', () => {
 
   it('acl must be valid', () => {
     const user = new Parse.User();
-    assert.equal(user.setACL(`Ceci n'est pas un ACL.`), false);
+    expect(() => user.setACL(`Ceci n'est pas un ACL.`)).toThrow(
+      new Parse.Error(Parse.Error.OTHER_CAUSE, 'ACL must be a Parse ACL.')
+    );
```
mtrezza, line 15: How is this related to the Parse.Object.set issue, or is this something else?

dplewis: `setACL` calls `set`
```
 setACL(acl: ParseACL, options?: any): ParseObject {
    return this.set('ACL', acl, options);
  }
```

mtrezza:
BREAKING CHANGE: Removes the validation error option from `Parse.Object.set`, `Parse.Object.setACL` and instead throws on validation errors. For example, instead of using `Parse.Object.set(key, value, { error: ... })` wrap `Parse.Object.set(key, value)` into a `try/catch` block.

@dplewis Is this proposed changelog entry correct?

dplewis:
There is also `Parse.Object.unset` and `Parse.Role.setName` that throw errors now.

Edit: Basically setting / updating a field with invalid key or acl will throw error now. Before you would have to save first to get the error. If a developer fixed all errors before updating to using this change there shouldn't be any issues. If they just set fields and never saved then that would be a breaking change

mtrezza:
I guess these are 2 changes then:

1) remove `error` option
2) throw not at save but at `set` et al.

Could we do only (1) and not (2), if it's not necessary for (1)? In any case, if the primary purpose of this PR is (1), then this would be a breaking change either way, because it's possible that someone is using it, right?

dplewis:
> remove error option

If you didn't pass an error option the error would be slienced but now it's thrown

> throw not at save but at set et al.

The error would get thrown before you get a chance to save.

mtrezza:
So the original behavior is:

- if error option is set, then error is thrown at .set
- if error option is not set, then error is thrown at .save

New behavior with this PR:

- error is thrown at .set (error option doesn't exist anymore)

Is that correct?

dplewis:
This is correct. Users have reported that save didn’t throw at all like https://github.com/parse-community/Parse-SDK-JS/issues/1386 and there were a few other examples but I can't find them.

mtrezza:
Do you think it makes more sense to throw on .set or on .save? Throwing on .save seems to make more sense to me, because .set seems more like a non-op. Do we have any other similar examples in the SDK?

dplewis:
I think it should throw on .set because it's easier to debug where the error is in the stack trace. Also this PR adds a message for `INVALID_KEY_NAME` error so you can get the name of the key. The documentation and return types are improved and consistent with the rest of the SDK like `query` chaining. `.set` returning `Parse.Object | boolean` is weird.

mtrezza:
BREAKING CHANGE: Removes the error handler option `error` from `Parse.Object.set`, `Parse.Object.setACL`, `Parse.Object.unset`, `Parse.Role.setName` and instead throws on validation error. Previously, if the `error` option was set, the handler was invoked if a validation error occurred on `Parse.Object.set`, and if no handler was set, an error was thrown on `Parse.Object.save`. The new behavior is that an error is thrown at `Parse.Object.set`. For example, instead of using `Parse.Object.set(key, value, { error: ... })` wrap `Parse.Object.set(key, value)` into a `try/catch` block. 

@dplewis Is this proposed changelog entry correct?

dplewis:
LGTM!

parseplatformorg:
🎉 This change has been released in version [6.0.0-alpha.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/6.0.0-alpha.1)

parseplatformorg:

Issue #2012, opened 2023-09-01
Expand TypeScript support


mtrezza:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-JS/issues?q=is%3Aissue).

### Current Limitation

TypeScript support has been added with https://github.com/parse-community/Parse-SDK-JS/issues/1950. The next step is to expand it to more files.

#### Contributing

* Change file extension to `.ts`
* Remove @flow from file if it exists
* Add file to `./tsconfig.json`
* `npm run build:types`
* Fix types
* Add / enable tests  in `types/test.ts`
* `npm run test:types`
* Compare types to [@types/parse](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b23a36e669fa127d1035e22ca93faab85b98e49f/types/parse/index.d.ts#L11)

#### Progress (1/62)

- [x] Analytics.js
- [x] AnonymousUtils.js
- [x] Cloud.js
- [x] CloudCode.js
- [x] CoreManager.js
- [x] CryptoController.js
- [x] EventEmitter.js
- [x] EventuallyQueue.js
- [x] FacebookUtils.js
- [x] IndexedDBStorageController.js
- [x] InstallationController.js
- [x] LiveQueryClient.js
- [x] LiveQuerySubscription.js
- [x] LocalDatastore.js
- [x] LocalDatastoreController.js
- [x] LocalDatastoreController.react-native.js
- [x] LocalDatastoreUtils.js
- [x] ObjectStateMutations.js
- [x] OfflineQuery.js
- [x] Parse.js
- [x] ParseACL.js
- [x] ParseCLP.js
- [x] ParseConfig.js
- [x] ParseError.js
- [x] ParseFile.js
- [x] ParseGeoPoint.js
- [x] ParseHooks.js
- [x] ParseInstallation.js
- [x] ParseLiveQuery.js
- [x] ParseObject.js
- [x] ParseOp.js
- [x] ParsePolygon.js
- [x] ParseQuery.js
- [x] ParseRelation.js
- [x] ParseRole.js
- [x] ParseSchema.js
- [x] ParseSession.ts
- [x] ParseUser.js
- [x] Push.js
[...]

Issue #1399, opened 2021-08-27
Replace the use of 'require' by 'import' to reduce library side and improve performance. (Tree shaking)


sadortun:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-JS/issues?q=is%3Aissue).

### Current Limitation

Follow up of the discussion on #203

Currently, the SDK is imported globally, and the compilers cannot optimize the imports and remove unused code. 

Firebase have a great explanation 
https://firebase.googleblog.com/2021/08/the-new-firebase-js-sdk-now-ga.html?m=1

### Feature / Enhancement Description

This would allow to reduce the compiled client side app by eliminating the unused code.

This would also allow to fix #203 by allowing to create instances  the Parse library for those who need to interact with multiple Servers.

### 3rd Party References

https://www.codingame.com/playgrounds/7463/tree-shaking-in-javascript-with-rollup

dplewis:
@sadortun @mtrezza I tried to replace `require` with `import` but ran into an issue with babel

I used import on `EventEmitter.ts`
```
import RNEventEmitter from 'react-native/Libraries/vendor/emitter/EventEmitter';
import { EventEmitter as NodeEventEmitter } from 'events';
let EventEmitter: any;

try {
  if (process.env.PARSE_BUILD === 'react-native') {
    EventEmitter = RNEventEmitter;
    if (EventEmitter.default) {
      EventEmitter = EventEmitter.default;
    }
    EventEmitter.prototype.on = EventEmitter.prototype.addListener;
  } else {
    EventEmitter = NodeEventEmitter;
  }
} catch (_) {
  // EventEmitter unavailable
}
module.exports = EventEmitter;
export default EventEmitter;
```

Babel outputs to `lib/node.js`, as you can see the dead code has been eliminated but the unused import `_EventEmitter` didn't get removed. This is a prime example of tree shaking. Are there any babel experts that would like to have a look?
```
var _EventEmitter = _interopRequireDefault(require("react-native/Libraries/vendor/emitter/EventEmitter"));
var _events = require("events");
var EventEmitter;
try {
  EventEmitter = _events.EventEmitter;
} catch (_) {
[...]

Issue #1368, opened 2021-05-24
Check attributes spelling when using multi level includes


frouo:
### Is your feature request related to a problem? Please describe

I use Typescript and I could not find a way to check object's attributes **when using multi level includes**.

One level include works very fine (as shown below), but what if we use multi level?

### Describe the solution you'd like

Having a static pass-through function that returns the attribute:

```
attr<K extends Extract<keyof T, string>>(attr: K): K;
```
**Usage could be like**

```
  .include(`${Comment.attr("post")}.${Post.attr("author")}`)
```

or

```
  .include(`${Attributes<Comment>.get("post")}.${Attributes<Post>.get("author")}`)
```

_Note: I don't pretend to be a Typescript expert so may consider the above suggestions as pseudo-code._

### Additional context

```
export class Comment extends Parse.Object<{
  post: Post;
  text: string;
}> {
  constructor(options?: any) {
    super("Comment", options);
  }
}

export class Post extends Parse.Object<{
  author: Parse.User;
}> {
  constructor(options?: any) {
    super("Post", options);
  }
}

export async function getComments() {
  return await new Parse.Query(Comment)
    .include("post") // here Typescript works very well, it checks for Comment's attributes: include(...key: ("post" | "text" | keyof Parse.BaseAttributes)[]): Parse.Query<Comment>
    .include("post.author") // here Typescript underlines red
    .find();
}
```

### Screenshots

<img width="595" alt="Capture d’écran 2021-05-24 à 13 25 44" src="https://user-images.githubusercontent.com/2499356/119341084-a9357780-bc93-11eb-8fe2-400fb8aced35.png">

<img width="859" alt="Capture d’écran 2021-05-24 à 13 24 44" src="https://user-images.githubusercontent.com/2499356/119341056-a20e6980-bc93-11eb-827d-f0db069ddf02.png">

Thank you

parseplatformorg:
🎉 This change has been released in version [6.1.0-alpha.3](https://github.com/parse-community/Parse-SDK-JS/releases/tag/6.1.0-alpha.3)

parseplatformorg:
🎉 This change has been released in version [6.1.0](https://github.com/parse-community/Parse-SDK-JS/releases/tag/6.1.0)

PR #2452, opened 2025-02-14
refactor: Fix TypeScript error that namespace `Parse` cannot be found

dplewis:
## Pull Request

## Issue

Ideally we would like to be backwards compatible with [@types/parse](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b23a36e669fa127d1035e22ca93faab85b98e49f/types/parse/index.d.ts#L11). @types/parse uses namespaces while this SDK uses modules. This PR fixes `Cannot find namespace "Parse"` error.

## Approach

* Add Parse namespace to types only
* Remove ParseType interface
* Add types path for `parse/node` and `parse/react-native` with tests
* Add missing types
* Update `test:types` script to include test error count

Types test errors left: 495 (from 585)

TODO: (Added to https://github.com/parse-community/Parse-SDK-JS/issues/2012#issuecomment-2119602706)
- [ ] Backwards compatible with [@types/parse/index.d.ts](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b23a36e669fa127d1035e22ca93faab85b98e49f/types/parse/index.d.ts#L11)
  - [ ] Add generics to Parse.Object
  - [ ] Add generics to Parse.Query
  - [ ] Add generics to Parse.Installation
  - [ ] Add generics to Parse.Relation
  - [ ] Add generics to Parse.Schema
  - [ ] Add generics to Parse.Push
  - [ ] Add generics to Parse.Role
  - [ ] Add Parse.Cloud node / cloud code types
  - [ ] Add missing types where needed
  - [ ] Pass all tests

## Tasks

- [x] Add tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,10 +29,10 @@ jobs:
       - run: npm ci
       - name: Build Types
         run: npm run build:types
-      # - name: Test Types
-      #   run: npm run test:types
       - name: Lint Types
         run: npm run lint:types
+      - name: Test Types
+        run: npm run test:types
   check-docs:
     name: Check Docs
     timeout-minutes: 10
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,8 +29,8 @@ jobs:
       - run: npm ci
       - name: Build Types
         run: npm run build:types
-      - name: Test Types
-        run: npm run test:types
+      # - name: Test Types
+      #   run: npm run test:types
       - name: Lint Types
         run: npm run lint:types
   check-docs:
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "test": "cross-env PARSE_BUILD=node jest",
     "test:mongodb": "npm run test:mongodb:runnerstart && npm run integration",
     "test:mongodb:runnerstart": "mongodb-runner start -- --port 27017",
-    "test:types": "dtslint --expectOnly types",
+    "test:types": "dtslint --expectOnly types 2>&1 | tee temp.txt; echo \"Type tests failed: $(grep 'ERROR:' temp.txt -c)\"; rm temp.txt;",
     "posttest:mongodb": "mongodb-runner stop --all",
     "lint": "eslint --cache src/ integration/",
     "lint:fix": "eslint --fix --cache src/ integration/",
--- a/src/Parse.ts
+++ b/src/Parse.ts
@@ -37,101 +37,21 @@ import LocalDatastoreController from './LocalDatastoreController';
 import StorageController from './StorageController';
 import WebSocketController from './WebSocketController';
 
-/**
- * Contains all Parse API classes and functions.
- *
- * @static
- * @global
- * @class
- * @hideconstructor
- */
-interface ParseType {
-  ACL: typeof ACL;
-  Parse?: ParseType;
-  Analytics: typeof Analytics;
-  AnonymousUtils: typeof AnonymousUtils;
-  Cloud: typeof Cloud & {
-    /** only available in server environments */
-    useMasterKey?: () => void;
-  };
-  CLP: typeof CLP;
-  CoreManager: typeof CoreManager;
-  Config: typeof Config;
-  Error: typeof ParseError;
-  EventuallyQueue: typeof EventuallyQueue;
-  FacebookUtils: typeof FacebookUtils;
-  File: typeof File;
-  GeoPoint: typeof GeoPoint;
-  Hooks?: any;
-  Polygon: typeof Polygon;
-  Installation: typeof Installation;
-  LocalDatastore: typeof LocalDatastore;
-  Object: typeof ParseObject;
-  Op: {
-    Set: typeof ParseOp.SetOp;
-    Unset: typeof ParseOp.UnsetOp;
-    Increment: typeof ParseOp.IncrementOp;
-    Add: typeof ParseOp.AddOp;
-    Remove: typeof ParseOp.RemoveOp;
-    AddUnique: typeof ParseOp.AddUniqueOp;
-    Relation: typeof ParseOp.RelationOp;
-  };
-  Push: typeof Push;
-  Query: typeof Query;
-  Relation: typeof Relation;
-  Role: typeof Role;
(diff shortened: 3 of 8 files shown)


mtrezza:
What are we using `test:types` for? What's does that ensure?

dplewis:
it ensures that our autogenerated types matches @types/parse

dplewis:
@mtrezza Is this good to merge? I updated the CI, now we can see how many tests failed. When it reaches 0 is when we can publish types.

mtrezza:
> it ensures that our autogenerated types matches @types/parse

Could you help me understand why that is necessary? Aren't the generated types always correct, because they reflect the method and variable types as they are defined in code?

dplewis:
If a developer has an existing typescript project that uses @types/parse then switch to our types this will cause a breaking change as is.

There are 495 errors that lets us know the generated types aren't correct. 

Screenshot 2025-02-15 at 7 19 17 PM

mtrezza:
But shouldn't our generated types always be correct? In your screenshot, if the type is any, then it's any, if it's void, then it's void. We define that type in code after all. Or what does the error in your screenshot tell us - that we have set the wrong type in code?

dplewis:
I'm not a Typescript expert but I believe after this PR is merged you can publish the types if you want and incrementally add, which was always the original intention. The types test suite tells that the types in our code is wrong or missing types. We are doing this for the sake of completeness. I can foresee tons of users posting issues like https://github.com/parse-community/Parse-SDK-JS/issues/1368

mtrezza:
Originally we didn't publish the types, because during the time of gradual conversion to TS the DT types were more complete had a greater benefit for developers. Had we published the types, they'd have been in conflict with DT and a developer would have had to remove the more complete TD types for our less complete own types.

We also said that once we achieve a decent type coverage, specifically replace enough of these `any` placeholders with correct types, we'll publish our types.

Now I understand that the types check here errors because we still have many of these `any` types. But simply searching the code base for `any` types will also tell us that we still have too many, even though there may be some places where they may be actually the correct types.

1) I doubt we'll ever align 100% with DT. I don't think that's necessary for publishing our own types, as long as our types provide a decent value for developers compared to DT. DT is currently missing some types and contains outdated types due to recent code changes on our side. So even though we may use many `any` types, our additional type coverage may compensate for that. It will be our decision when we publish, and it will likely be before types test returns no errors.

2) We're aligning with DT but does that even make sense? For example, using a namespace here; is that beneficial over using modules? Are there other considerations that would be beneficial that are not in DT?

dplewis:
> We also said that once we achieve a decent type coverage, specifically replace enough of these any placeholders with correct types, we'll publish our types.

How do we check type overage? That's what the types test suite is for. It makes no sense to throw away years of work that went into @types/parse. It's more the just find `any` and replace them. For example`Parse.Cloud.beforeSave` has no typing because it doesn't exist in this SDK. How do I know this because the test fails.

> I doubt we'll ever align 100% with DT

The hard part is done with this PR. I have another branch that fixed over 200 errors. I can get this done in a week.

> We're aligning with DT but does that even make sense? For example, using a namespace here; is that beneficial over using modules? Are there other considerations that would be beneficial that are not in DT?

Let's take `Parse.Object` for example the **Value** uses `ParseObject` but the **Type** uses `Parse` namespace in `index.d.ts` because of the dot notation. Look at the PHP SDK it uses namespaces. This SDK only uses namespaces for Types.

Unfortunately we can't rename `ParseObject` to `Object` or `ParseError` to `Error` as these classes exist in CommonJS otherwise we could fix [Replace the use of 'require' by 'import' to reduce library side and improve performance. (Tree shaking)](https://github.com/parse-community/Parse-SDK-JS/issues/1399), refactor the namespace exports to use `export * from './Parse'` or `import { Query, Object } from 'parse'` but that's a different story

dplewis:
@mtrezza I edited my last comment

mtrezza:
> How do we check type overage? 

Right, how do we do that at all? We can't rely on DT types, because that may not be up to date. Our code is the source of truth.

> For exampleParse.Cloud.beforeSave has no typing because it doesn't exist in this SDK. How do I know this because the test fails.

I guess it exists in Parse Server then? So, Parse Server should provide these specific types, not this SDK, right?

If it helps with the conversion to keep DT for now, let's keep it. But I think we should set a point in time to remove it.

dplewis:
> Right, how do we do that at all? We can't rely on DT types, because that may not be up to date. I would say

We use it as a base, we can always change and update the test suite to match the current state of the SDK.

> And what is the solution if it doesn't exist in this SDK? 

Just mock them and throw an error if called client side.

mtrezza:
> We use it as a base, we can always change and update the test suite to match the current state of the SDK.

OK, but let's aim to remove the DT dependency at some point in the near future.

> Just mock them and throw an error if called client side.

Can't Parse Server provide these types? If the types a specific to the Server environment, the client SDK shouldn't provide them at all IMO.

dplewis:
> OK, but let's aim to remove the DT dependency at some point in the near future.

 We don't have a dependency on DT, we are just using the test suite thats all

>Can't Parse Server provide these types? If the types a specific to the Server environment, the client SDK shouldn't provide them at all IMO.

I don't think it can. Let's fix the types first then worry about that when we get there.

mtrezza:
> We don't have a dependency on DT, we are just using the test suite thats all

I'm referring to the dependency `dtslint` which we'll remove in https://github.com/parse-community/Parse-SDK-JS/pull/2435.

> When it reaches 0 is when we can publish types

So we'll merge https://github.com/parse-community/Parse-SDK-JS/pull/2435 once it reaches 0, is that the plan?

dplewis:
> So we'll merge https://github.com/parse-community/Parse-SDK-JS/pull/2435 once it reaches 0, is that the plan?

Now that I think about it we can use eslint to run the tests without waiting for it to reach 0. But yes that's the plan.

mtrezza:
You mean we could replace dtslint with eslint immediately? Do you want to do that, or should I merge the PR as is?

dplewis:
Let's merge this first

mtrezza (approved):

parseplatformorg:

PR #2251, opened 2024-07-26
ci: Remove husky

mtrezza:
## Pull Request

- Report security issues [confidentially](https://github.com/parse-community/Parse-SDK-JS/security/policy).
- Any contribution is under this [license](https://github.com/parse-community/Parse-SDK-JS/blob/alpha/LICENSE).
- Link this pull request to an [issue](https://github.com/parse-community/Parse-SDK-JS/issues?q=is%3Aissue).

## Issue

Closes: FILL_THIS_OUT

## Tasks

- [ ] Add tests
- [ ] Add changes to documentation (guides, repository pages, code comments)

PR #2460, opened 2025-02-19
refactor: Fix autogenerate TS types

dplewis:
## Pull Request

## Issue

Currently developers would have to generate types themselves before committing. This could lead to types being missing in the future. I know there was talks for removing husky https://github.com/parse-community/Parse-SDK-JS/pull/2251 and there maybe a better way to do this. I'm open to discussion.

## Approach

* Improve type build speed by removing prettier and lint formatting 
* Regenerate all types
* Generate types per commit.

## Tasks

- [ ] Add tests


mtrezza:
Doesn't this cause quite a delay when building the types on every commit? I mean if a developer commits every few lines, that would trigger type rebuilding every time.

dplewis:
That is correct. It's about 3-5 seconds.

mtrezza:
And that delay may depend on the machine one is working on and be even slower in some cases. Not a fan of pre-commit hooks for code generation or modification - for many reasons. I believe a failing CI test and a good description in the error message makes more sense. Let's take the Parse Server options definitions for example. Generating them automatically would be a bad idea, because if someone edits a generated file instead of the source file for definitions, that change is lost and may go unnoticed. Better just fail. As for reviewing; mostly just requires pointing to the failing CI job and the contributor should know themselves what to do - with a good error msg. We should really get rid of husky.

dplewis:
> mostly just requires pointing to the failing CI job and the contributor should know themselves what to do - with a good error msg.

Let's do this. In this case removing `build:types` from the CI should let us know that a type is missing if the contributor writes a test case. We will have to ensure a test is written everytime a type has changed.

dplewis:
Closing in favor of https://github.com/parse-community/Parse-SDK-JS/pull/2462

parseplatformorg:
🎉 This change has been released in version [6.1.0-alpha.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/6.1.0-alpha.1)

parseplatformorg:
🎉 This change has been released in version [6.1.0](https://github.com/parse-community/Parse-SDK-JS/releases/tag/6.1.0)

PR #2487, opened 2025-03-08
test: Investigate integration test timeout

dplewis:
## Pull Request

## Issue

Trying to narrow down what's causing the following issue. I can't replicate it locally so this is guess work. Everytime this issue happens it when on a test that calls `reconfigureServer`. There are currently 31 test that uses this function. I believe it's stalling in either `ParseServer.handleShutdown` or `ParseServer.server.close`. I would assume `close` would be added into `shutdown` but it isn't. There might be something stalling in shutdown too i.e LiveQueryServer. 

Another place to look is the `afterEach` in `helper.js` as there might be a stall there with deleting all records from the database or logging out user.

There is a test case on the Server that if fixed would give a clue on how to solve this issue. Unfortunately it's being skipped.

https://github.com/parse-community/parse-server/blob/4f55c3e9cfa31c6aca9c005ab21ff6193600ba89/spec/ParseServer.spec.js#L47-L49
```
Error: Timeout - Async function did not complete within 20000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL)
```
https://github.com/parse-community/Parse-SDK-JS/actions/runs/13735695802/job/38419094737?pr=2486

## Approach

* Remove unused route
* Cache distribution files for performance
* Remove potentally stalled functions
* Improve shutdown tests
* Set REQUEST_ATTEMPT_LIMIT to 1 to check for network drops
* Add try / catch to afterEach in helper.js to check for stalling
* Use jasmine default timeout (5000ms) to check if startup time is an issue.

Might have to re-run the CI a few times to see if this issue replicates

## Tasks

- [x] Add tests


dplewis:
After further investigation `parseServer.handleShutdown` calls `liveQueryServer.server.close()`. When initializing the server with `startLiveQueryServer` option `liveQueryServer.server` and `parseServer.server` are the same so we are calling `server.close()` twice. This maybe the cause of the issue. As we can see the tests still pass without getting an `EADDRINUSE` error because now we are calling close once.

dplewis:
@mtrezza This is good to go. Can you rerun it a few times before merging?

mtrezza:
This has been closed?

dplewis:
I closed it because I'm still working on it. Didn't want to keep pinging the community with dozen of commits

mtrezza:
Got it, I think you could also put a PR in draft state, so visitors will know that it doesn't require any action.

PR #2185, opened 2024-06-23
ci: Properly handle open connections after the tests finish

dplewis:
## Pull Request

- Report security issues [confidentially](https://github.com/parse-community/Parse-SDK-JS/security/policy).
- Any contribution is under this [license](https://github.com/parse-community/Parse-SDK-JS/blob/alpha/LICENSE).
- Link this pull request to an [issue](https://github.com/parse-community/Parse-SDK-JS/issues?q=is%3Aissue).

## Issue

I realized the the open connections to the server in the test suite are just the node process to run Jasmine. This causes a lot of clutter in the CI logs when trying to review PRs. Logging to console also slows down the test suite.

Screenshot 2024-06-23 at 3 25 10 PM

## Approach

* Keep all connections alive until end of the test suite
* Remove destroying of open connections
* Remove `sleep(0)` from execution
* Perf: save ~10 seconds in the test suite

## Tasks

- [x] Add tests
- [x] Add changes to documentation (guides, repository pages, code comments)

PR #2490, opened 2025-03-09
test: Prevent test timeout error by properly closing server

dplewis:
## Pull Request

## Issue

https://github.com/parse-community/Parse-SDK-JS/actions/runs/13735695802/job/38419094737?pr=2486
```
Error: Timeout - Async function did not complete within 20000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL)
```
`parseServer.handleShutdown` calls `liveQueryServer.server.close()`. When initializing the server with `startLiveQueryServer` option `liveQueryServer.server` and `parseServer.server` are the same so we are calling server.close() twice. This maybe the cause of the issue. As we can see the tests still pass without getting an EADDRINUSE error because now we are calling close once.

See: https://github.com/parse-community/Parse-SDK-JS/pull/2487

## Approach

* Remove unused route
* Cache distribution files for performance
* Remove multiple server close events.
* Improve shutdown tests
* Refactor handleShutdown
* Set REQUEST_ATTEMPT_LIMIT to 1 to check for network drops
* Remove `afterAll` since the tests end at this point we don't care about open connections as the server will shutdown anyway.

If the issue persists then we can try to close all the connections. `server.close` prevents new connections but keeps existing ones alive until they are closed. There is now a [server.closeAllConnections()](https://nodejs.org/dist/latest-v18.x/docs/api/http.html#servercloseallconnections) method introduced in Node v18.2.0. If all else fails we can look into `parseServer.handleShutdown()` internals.


jasmine.json
```diff
@@ -6,6 +6,5 @@
   "spec_files": [
     "*Test.js"
   ],
-  "random": true,
-  "timeout": 20000
```
mtrezza, line 10: Why remove timeout? What default timeout does apply?

dplewis: helper.js
```
jasmine.DEFAULT_TIMEOUT_INTERVAL = 20000;
```
It was currently set in two different places. Now it's easier to find and change.

dplewis:
@mtrezza The issue still occurred. https://github.com/parse-community/Parse-SDK-JS/actions/runs/13749596429/job/38448859505

This tells me that this issue is with `parseServer.handleShutdown()` stalling. I added logic to run handleShutdown in the background and listen to the close event. I also destroyed all the connections to ensure the event is emitted. https://github.com/parse-community/Parse-SDK-JS/pull/2490/commits/afdeb2bf0bfc89582ca8ab4198eabe7d5edfbe99

Destroy connections was removed in https://github.com/parse-community/Parse-SDK-JS/pull/2185. This didn't work because you have to destroy the connections *after* you call `server.close()` to prevent a race condition.

https://nodejs.org/api/net.html#event-close
```
Emitted when the server closes. If connections exist, this event is not emitted until all connections are ended.
```

If this fails again the offending line is in `handleShutdown` for the same reasons stated above
https://github.com/parse-community/parse-server/blob/4f55c3e9cfa31c6aca9c005ab21ff6193600ba89/src/ParseServer.js#L232

dplewis:
@mtrezza Until `parseServer.handleShutdown` is fixed on the server I moved the shutdown logic here. This gives us better control of the shutdown process. This should be good to merge. I ran it 20 times locally and got no errors timeout errors.

dplewis:
@mtrezza This is ready for review

mtrezza:
> @mtrezza Until parseServer.handleShutdown is fixed on the server I moved the shutdown logic here. This gives us better control of the shutdown process. This should be good to merge. I ran it 20 times locally and got no errors timeout errors.

Are you referring to https://github.com/parse-community/parse-server/pull/9634? Since it has been merged, are there any changes needed here, e.g. can the shutdown logic be removed here - because you wrote "Until parseServer.handleShutdown is fixed..."?

dplewis:
You can merge this now then we can change it whenever https://github.com/parse-community/parse-server/pull/9634 gets merged (I hope it works) and the server update gets released. This should work fine as is.

mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [6.1.0-alpha.3](https://github.com/parse-community/Parse-SDK-JS/releases/tag/6.1.0-alpha.3)

parseplatformorg:
🎉 This change has been released in version [6.1.0](https://github.com/parse-community/Parse-SDK-JS/releases/tag/6.1.0)

PR #2526, opened 2025-03-18
feat: Add support for MongoDB `$search` operator

### LINKED ISSUE #1989: Add support for MongoDB Atlas `$search` operator

### Current Limitation

Parse JS SDK already provides support for the MongoDB [$text](https://www.mongodb.com/docs/manual/reference/operator/query/text/) operator, but not the MongoDB Atlas (hosted MongoDB) specific [$search](https://www.mongodb.com/docs/atlas/atlas-search/searching/) operator, or [Atlas Search](https://www.mongodb.com/docs/atlas/atlas-search/).

Add support for the MongoDB Atlas `$search` operator. Note that this operator is a header operator for multiple [sub-operators](https://www.mongodb.com/docs/atlas/atlas-search/operators-and-collectors/), such as [$regex](https://www.mongodb.com/docs/atlas/atlas-search/regex/).

### Example Use Case
Search with regex using MongoDB Atlas Search.

Using the MongoDB driver directly via the Parse Server database adapter.

dblythy:
## Pull Request

## Issue

Closes: #1989

## Approach

Adds ability to use Mongodb atlas search


src/ParseQuery.ts
```diff
@@ -1587,6 +1591,48 @@ class ParseQuery<T extends ParseObject = ParseObject> {
...
+   * Triggers a MongoDb Atlas Text Search
+   *
+   * @param {string} value The string to search
+   * @param {string[]} path The fields to search
+   * @param {object} options (Optional)
+   * @param {string} options.index The index to search
+   * @returns {Promise} Returns a promise that will be resolved with the results
+   * of the search
+   * */
+
```
mtrezza, line 1603:
```suggestion
```

dplewis:
@dblythy I fixed the merge conflicts, can you improve code coverage?

src/ParseQuery.ts
```diff
@@ -1585,6 +1589,48 @@ class ParseQuery<T extends ParseObject = ParseObject> {
...
+  /*
+   * Triggers a MongoDb Atlas Text Search
+   *
+   * @param {string} value The string to search
+   * @param {string[]} path The fields to search
+   * @param {object} options (Optional)
+   * @param {string} options.index The index to search
+   * @returns {Promise} Returns a promise that will be resolved with the results
+   * of the search
+   * */
```
dplewis: */ should be moved to the next line, I don’t think the docs are generated properly

dplewis (changes requested):

src/ParseQuery.ts
```diff
@@ -1585,6 +1589,48 @@ class ParseQuery<T extends ParseObject = ParseObject> {
...
+
+  async search(value: string, path: string[], options: SearchOptions = {}): Promise<T[]> {
```
dplewis: options?: is preferred over default value here. Check the generated ParseQuery.d.ts file

src/__tests__/ParseQuery-test.js
```diff
@@ -2709,6 +2709,53 @@ describe('ParseQuery', () => {
...
+    const q = new ParseQuery('Item');
+    await expect(q.search()).rejects.toThrow('A search term is required.');
+  });
+
+  it('search term must be a string', async () => {
+    const q = new ParseQuery('Item');
+    await expect(q.search(123)).rejects.toThrow('The value being searched for must be a string.');
+  });
+
+  it('search can return an empty array if no results', async () => {
```
dplewis, line 2753: Can you add an integration test? You can use those results to finish unit test coverage

dplewis (changes requested):

src/__tests__/ParseQuery-test.js
dblythy: This feature only works with Mongodb [atlas](https://www.mongodb.com/docs/atlas/atlas-search/index-definitions/) search index, not sure it'll work with mongodb runner

dplewis: Yeah it won’t work, you can try to run it in your atlas environment then use those results for unit tests to ensure fromJSON works

src/ParseQuery.ts
```diff
@@ -1585,6 +1589,49 @@ class ParseQuery<T extends ParseObject = ParseObject> {
...
+   * @param {string} value The string to search
+   * @param {string[]} path The fields to search
+   * @param {object} options (Optional)
+   * @param {string} options.index The index to search
+   * @returns {Promise} Returns a promise that will be resolved with the results
+   * of the search
+   *
+   */
+
+  async search(value: string, path: string[], options?: SearchOptions): Promise<T[]> {
```
dplewis, line 1604: For consistency we should rename `value` to `query` and it should allow for string | string[]. This should be good enough for a simple search. We can add regex, allowAnalyzedField, and score options later or now if you prefer. For more advanced searching like compound the developer would have to use `query.aggregate`.

Edit: There is a empty line above that should be removed

dplewis (changes requested):

mtrezza (commented):
Why has this been added as a new API? I think just allowing the `$search` op in the aggregation pipeline would be enough?

dplewis:
@mtrezza I just tried it with the aggregate pipeline and it works. I think this can be closed, no need to add new api.

mtrezza:
The [issue objective](https://github.com/parse-community/Parse-SDK-JS/issues/1989#issuecomment-2398260026) was to add support for `$search` and `$searchMeta` to the aggregation pipeline. If I'm not mistaken, either the JS SDK or Parse Server are validating the aggregation pipeline stages, i.e. only the keys, not their values. If that already works, then either the key validation feature has been removed in the meantime, or the stage names are already supported. We should find out why this seems to work, to ensure that it really works.

dplewis:
The stage check was removed https://github.com/parse-community/parse-server/pull/7237

Ironically it was because somebody wanted to use `$search` lol

mtrezza:
Ha! Even in PS5 already. Are we sure there there is not further validation in the JS SDK for example? The [added test](https://github.com/parse-community/parse-server/pull/7237/files#diff-2d6ed0f1c1c470ba3976fa6bb8b1424d6ad40680a28cd1a398fa3ce2803dafbc) looks suspicious. I would rather see a test that uses `Parse.Query.aggregate` to go through the JS SDK. Even though `$search` is no supported by mongodb-runner, but we could test the error, or mock and inspect what's sent to the driver. I think this PR should be changed to add such a test if none exists yet for `$search`, otherwise I agree that it can be closed.

dplewis:
You opened https://github.com/parse-community/Parse-SDK-JS/issues/1989 did you test that it wasn't supported.

mtrezza:
That was 2 years ago, but I believe to vaguely remember that I wanted to use it and it didn't work. But maybe it was an older version of Parse Server, or maybe there has been a bug fix in the meantime, hard to tell now. I think by adding a test we can close this issue.

dplewis:
If you want to do a test then it should be done on the server since there is access to the driver. This isn't an SDK issue.

mtrezza:
We don't know whether it's an SDK issue, until we do an integration test, right? I think it make sense to add a test in either repo, that's unfortunately the [situation](https://github.com/parse-community/parse-server/issues/8787) we're in with the JS SDK being part of Parse Server.

dplewis:
I know it’s not an SDK issue because I tested it on my machine, also any aggregate pipeline gets passed to the server regardless of the stages. A good test is the error thrown if MongoDB is used without atlas search

dblythy:
Just an FYI that I use mongodb atlas search myself on my own servers:

```
const pipeline = [
        {
          $search: {
            index: "default",
            text: {
              path: ["name"],
              query: search,
            },
          },
        },
        { $match: { approved: true, mode: user.get("mode"), published: true } },
        { $limit: 100 },
      ];
      const response = await new Parse.Query(Business).aggregate(pipeline as any);

```
      
And it works great

mtrezza:
Then I'm not sure why this PR was opened in the first place. The issue was that $search and $searchMeta should be supported in the aggregation pipeline. Since it's already supported, the issue should simply have been closed.

dplewis:
> Then I'm not sure why this PR was opened in the first place.

Because somebody opened a feature request for a feature that already exist and let it stay open a year and a half. Thankfully we found 2 bugs because of this PR.

PR #2539, opened 2025-03-27
ci: Run CI with specific Parse Server versions

mtrezza:
## Summary by CodeRabbit

* **Chores**
  * Expanded CI matrix to include additional Parse Server variants (now testing multiple Parse Server versions alongside Node.js versions) with clearer matrix naming.
  * Split install/test into separate dependency-install, unit test, and integration test steps for more granular runs.
  * Added dynamic Node setup and propagation of code coverage token for uploads.


dplewis:
@mtrezza Doesn't https://github.com/parse-community/Parse-SDK-JS/pull/2557 fix this build issue? Can you update from alpha?

mtrezza:
This is a different thing. It's an attempt (wip) to change the CI to install a specific version of Parse Server, then install the current JS SDK in Parse Server and then run the full Parse Server tests. It's to ensure that a change in the JS SDK won't break a Parse Sever test.

As a first step this should only run the SDK tests with different PS versions, but that fails.

dplewis:
I'd rather keep the current way of release the SDK then open a PR on the server and see if it breaks. What happens if the Parse Server test suite does break? We can't update the Parse Server tests as the SDK hasn't been released yet. Not to mention adding 10+ minutes of ci time.

Keep the current changes in this PR and add something like `it_only_parse_server_version` to fix CI. The compatibility table also needs updating.

mtrezza:
> I'd rather keep the current way of release the SDK then open a PR on the server and see if it breaks. What happens if the Parse Server test suite does break? We can't update the Parse Server tests as the SDK hasn't been released yet.

Sure, this is just meant to detect issues in the SDK, not on the server side. The challenge is that we should detect an issue before making a stable release, because then the bug can affect developers. We don't test with alpha releases of the SDK, so we are releasing an SDK that has not been fully tested with Parse Server. 

I believe this goes back to https://github.com/parse-community/parse-server/issues/8787, which gives us 2 uses cases:

- a) SDK used as client -- the SDK tests should detect any issues; we still do a lot of mocking here, but we have integration tests, so that works for now.
- b) SDK used as Parse Server dependency -- the SDK release must work internally with Parse Server, so the SDK CI should ensure this before merging a PR

I'd like to cover (b), which is currently not covered.

Another, unrelated issue is to run the SDK CI with multiple versions of Parse Server, which is the current change in this PR.

dplewis:
> The challenge is that we should detect an issue before making a stable release, because then the bug can affect developers.

Which bugs are you talking about? It's not like installing the SDK along side Parse Server will override the SDK internally used on the server.

Besides Parse.Error, Parse._encode, Parse._decode what else does Parse Server use internally that could potentially break it?

mtrezza:
> Besides Parse.Error, Parse._encode, Parse._decode what else does Parse Server use internally that could potentially break it?

Well, for example these. I'm fine with leaving as is, but I'm afraid it will take a long time until https://github.com/parse-community/parse-server/issues/8787 get's done.

This PR instead could focus on what you [mentioned](https://github.com/parse-community/Parse-SDK-JS/pull/2539#issuecomment-2802554143) and add `it_only_parse_server_version`, so just run the SDK against different version of Parse Server. Not sure why the CI doesn't pass, need to look into that.

dplewis:
> Well, for example these.

Those haven't changed in years and probably won't change.

> but I'm afraid it will take a long time until https://github.com/parse-community/parse-server/issues/8787 get's done.

Let's start by moving all internal uses of the SDK on Parse Server to a single file so we can see what exactly is used. Then we can decide if we want to use it as a `devDependency` or `dependency`. For example `Parse._encode` is only used for Dates and Cloud Code responses. We don't need the SDK to encode dates. Edit: I see you already updated https://github.com/parse-community/parse-server/issues/8787 

> Not sure why the CI doesn't pass, need to look into that.

Default ACL was added in v8 so it's different from v7

mtrezza:
Here we have an example of https://github.com/parse-community/Parse-SDK-JS/pull/2857 breaking the Parse Server CI. Going back and forth to figure out what's causing it. We need to run the Parse Server CI here to detect these issues before release.

parseplatformorg:
### :white_check_mark: **Snyk checks have passed. No issues have been found so far.**

| Status | Scanner |  Critical |  High |  Medium |  Low | Total (0) |
|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/acinader/pr-checks/ef5c613e-83c5-4045-8aca-b21b24c7458f?refs=74a0d243de02283cb4e5bd3646ac53d50d948882&source=prChecksComment) |

:computer: Catch issues earlier using the plugins for [VS Code](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner), [JetBrains IDEs](https://plugins.jetbrains.com/plugin/10972-snyk-security), [Visual Studio](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner-vs-2022), and [Eclipse](https://marketplace.eclipse.org/content/snyk-security#external-install-button).

dplewis:
> Here we have an example of https://github.com/parse-community/Parse-SDK-JS/pull/2857 breaking the Parse Server CI. 

No way, what broke?

mtrezza:
Note sure, maybe it only broke the parse server test suite, I'm looking into it...

mtrezza:
Ok, it seems to be a npm deduping issue with the push adapter, because the Parse JS SDK is all over the place. We need to get the error codes out of there finally. So not related to https://github.com/parse-community/Parse-SDK-JS/pull/2857. The next big refactor needs to be removing the JS SDK as dependency from all other repos, this is a spaghetti dependency nightmare.

PR #2547, opened 2025-04-01
fix: Remove unused option `sessionToken` from `Parse.Query.aggregate`

### LINKED ISSUE #1393: Aggregate query does not honor `sessionToken` option if passed
### New Issue Checklist

### Issue Description
Aggregate query does not honor the `sessionToken` option.

When passing a `sessionToken` it should only return objects that are readable by the user with that session token. Instead, the ACLs are ignored and all objects are returned.

Specifically, I have a query that is retrieving all objects of a certain class. A non-aggregate query could use the session token to only return objects the user has read permissions for. In the aggregate query, all objects are returned regardless of permissions.

I know that aggregate queries are different from normal queries and it may be impossible for you to support this option, but if that is the case the `sessionToken` should not be listed as an option for the aggregate function on the docs for Parse.Query (https://parseplatform.org/Parse-SDK-JS/api/master/Parse.Query.html#aggregate).

### Steps to reproduce

- Create a user and create some Parse Objects that the user has exclusive read and write access to in the ACL
- Create a Cloud function that runs an aggregate query for these Parse Objects and returns all of them
- Log in as a second user and run the aggregate query

### Actual Outcome
The objects owned by the first user are returned as part of the results for the second user's query.

### Expected Outcome
The objects owned by the first user should be excluded from the second user's query. Only objects the second user has read access to should be returned.

###  Environment

Server
- Parse Server version: `4.5.0`
- Operating system: Windows 10, I think also linux on our deployed server
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): local (and deployed)

Database
- System (MongoDB or Postgres): MongoDB
- Database version: `4.2.6`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): local (and deployed)

Client
- Parse JS SDK version: `3.3.0`

dplewis:
## Pull Request

## Issue

Aggregate queries don't use sessionTokens and most likely never will.

Closes: https://github.com/parse-community/Parse-SDK-JS/issues/1393

## Approach

Remove options from `distinct` and `aggregate` queries

## Tasks

- [x] Add tests

### DIFF
--- a/src/ParseQuery.ts
+++ b/src/ParseQuery.ts
@@ -780,29 +780,17 @@ class ParseQuery<T extends ParseObject = ParseObject> {
    * Executes a distinct query and returns unique values
    *
    * @param {string} key A field to find distinct values
-   * @param {object} options
-   * @param {string} [options.sessionToken] A valid session token, used for making a request on behalf of a specific user.
    * @returns {Promise} A promise that is resolved with the query completes.
    */
-  distinct<K extends keyof T['attributes'], V = T['attributes'][K]>(
-    key: K,
-    options?: { sessionToken?: string }
-  ): Promise<V[]> {
-    options = options || {};
-    const distinctOptions: { sessionToken?: string; useMasterKey: boolean } = {
-      useMasterKey: true,
-    };
-    if (Object.hasOwn(options, 'sessionToken')) {
-      distinctOptions.sessionToken = options.sessionToken;
-    }
+  distinct<K extends keyof T['attributes'], V = T['attributes'][K]>(key: K): Promise<V[]> {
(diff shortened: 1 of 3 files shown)


mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [6.1.0-alpha.5](https://github.com/parse-community/Parse-SDK-JS/releases/tag/6.1.0-alpha.5)

parseplatformorg:
🎉 This change has been released in version [6.1.1](https://github.com/parse-community/Parse-SDK-JS/releases/tag/6.1.1)

PR #2919, opened 2026-02-13
fix: Type regressions in ascending/descending fields

swittk:
## Pull Request

## Issue
Typings for `ascending`, `descending`, `addAscending`, and `addDescending` in the current version are typed as `...string[]`, which misses the original object key checks which were possible in the original DefinitelyTyped typings.
This change allows for the autocompletion & key inference type checks to work once again (this includes base attributes such as `objectId`, `createdAt`, `updatedAt` too).
The comma-string keys (which are technically supported, JS-wise, but were not supported in the original DefinitelyTyped typings) are not included in this scope yet, but should be possible to add in a future PR by using modern features.

## Approach

Used key inference in the same fashion as in the prior DefinitelyTyped typings.

## Tasks

- [x] Add tests

## Summary by CodeRabbit

* **Refactor**
  * Improved type safety for query sorting and field selection to enhance compile-time validation and autocompletion. No runtime behavior changes.
* **New Features**
  * Typings now explicitly permit the special full-text ranking key ($score) and common metadata fields (creation/update timestamps and object IDs) in query sorting and selection.
* **Tests**
  * Extended type tests to cover the updated sorting and selection typings.


parseplatformorg:
### :white_check_mark: **Snyk checks have passed. No issues have been found so far.**

| Status | Scanner |  Critical |  High |  Medium |  Low | Total (0) |
|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/acinader/pr-checks/36087af6-f900-4d55-889d-54bb64be9fb5?refs=ef96ed2b411264a139d93215dfac6427df5427f0&source=prChecksComment) |

:computer: Catch issues earlier using the plugins for [VS Code](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner), [JetBrains IDEs](https://plugins.jetbrains.com/plugin/10972-snyk-security), [Visual Studio](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner-vs-2022), and [Eclipse](https://marketplace.eclipse.org/content/snyk-security#external-install-button).

src/ParseQuery.ts
```diff
@@ -1560,12 +1560,13 @@ class ParseQuery<T extends ParseObject = ParseObject> {
...
   /**
    * Method to sort the full text search by text score
+   * `$score` is a special key used only for full text search ranking.
    *
    * @returns {Parse.Query} Returns the query, so you can chain this call.
    */
   sortByTextScore() {
     this.ascending('$score');
-    this.select(['$score'] as any);
+    this.select('$score');
```
mtrezza: Could you give more context about why this changed from `Array<string>` to `string`? `Parse.Query.select` usually takes an array, but from the description I understand it also accepts a single string with comma separation. I think we should remove this type ambiguity and make select only accept an array of strings. If you agree we could pin the change for the next major release. We should then keep using the array type wherever possible because we'll have to revert to it.

swittk: Sorry for the late reply; there isn't really any need for this change (it could be just ['$score'] the same as before), but there's really no real behavioural difference underneath I think; internally it just calls Array.isArray(key) check before, but then it does the key.join() immediately after + does a split(',') afterwards, so internally it supports all 3 types I mentioned (single array of keys, keys as string args, keys as string with comma splitting), and even likely supports a mix-and-match of all 3 types as multiple args due to how it's implemented in the `addAscending` and `addDescending` call sites.
The typings I think what people are definitely using (at least in my team) are:
- `.addAscending('key1', 'key2')` (the variadic arguments version)
- or the `.addAscending(['key1', 'key2'])` (the array args version)

I don't recall seeing the comma separated string thing documented, but it's definitely working and has been there a long time. I don't think we should remove this to be honest. I do have an idea for how to make the comma-separated typed multiple keys thing work strongly typed, but I was thinking of doing a separate PR for this since it seemed like a bit of TS magic. Just making keys typed the conventional way seemed straightforward.

swittk: @mtrezza sorry for the ping; do you have any suggestions to what I should edit?

mtrezza: > I don't recall seeing the comma separated string thing documented

Then we shouldn't use it, because we may remove it. If it's a long-standing feature, then we might do it with the major version release, even though it's undocumented, just to minimize the effects. But in general, we need to get out of this multi-type hell. It complicates code, adds complexity where it's not necessary. That means more bugs, more maintenance effort. 

For code examples, I suggest to use the variant we'll keep, which is the Array version, otherwise we have more work when we remove the other types in the future.

Does this help?

swittk: I'm pretty sure both the array form (.include(['a','b','c'])) and the string form (.include('key')) are in active use. Variadic string form also sees use but likely slightly less frequent. The only one I've never seen use is the comma separated thing.

swittk: I'm pretty sure both the array form (.include(['a','b','c'])) and the string form (.include('key')) are in active use. Variadic string form also sees use but likely slightly less frequent. The only one I've never seen use is the comma separated thing.

swittk:
I think this is ready to merge

mtrezza:
@coderabbitai resume

PR #3036, opened 2026-05-07
fix: Change getSessionToken return type to string | undefined

Carlgo11:
`getSessionToken()` was incorrectly declared as returning `string | null` even though the jsdoc says it returns string or undefined. This caused issues in our parse-server implementation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Clarified how missing session tokens are represented so absence is handled consistently.
  * Updated behavior and documentation to reduce unexpected issues when a session token is not available.
  * Improved type accuracy to increase reliability for integrations and tooling that check or consume session tokens.

### DIFF
--- a/types/ParseUser.d.ts
+++ b/types/ParseUser.d.ts
@@ -172,7 +172,7 @@ declare class ParseUser<T extends Attributes = Attributes> extends ParseObject<T
      *
      * @returns {string} the session token, or undefined
      */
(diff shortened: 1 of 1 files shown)


Carlgo11:
Not really sure why the Commit Message test fails. I renamed my commit to the `<type>: <summary` format

